PR #7, opened 2020-10-29
[GH-52] baseline unit test for k8ssandra-cluster chart template

jeffbanks:
**Includes**:
- Initial unit test coverage to support k8ssandra-cluster chart template.
- Struct/model for the c* datacenter, which is utilized with unmarshall operations.
- Separation of unit and integration target tests for CI/CD targets.


jsanda (approved):
Looks good! Sorry for the delayed review.

PR #48, opened 2020-11-18
Fix Tom's name and title

idleyoungman:
Hi all! I'm a teammate of @toffer. This is a fix for correcting his name and title.


bradfordcp:
My apologies for the mistake. We really appreciate the quote.

Issue #45, opened 2020-11-18
Do not deploy Reaper schema job until Cassandra is ready


jsanda:
First install the k8ssandra-cluster chart:

```
$ helm install demo
Error: must either provide a name or specify --generate-name
johns-mbp:charts jsanda$ helm install demo ./k8ssandra-cluster/
NAME: demo
LAST DEPLOYED: Tue Nov 17 20:11:06 2020
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
```
Then check on the status of the pods:

```
$ kubectl get pods
NAME                                                 READY   STATUS             RESTARTS   AGE
cass-operator-86d4dc45cd-kt7tp                       1/1     Running            0          10h
demo-grafana-operator-k8ssandra-68d47447d4-qxh8x     1/1     Running            0          114s
demo-reaper-k8ssandra-schema-7h89x                   0/1     CrashLoopBackOff   3          88s
demo-reaper-operator-k8ssandra-6598cb7f74-vzcbv      1/1     Running            0          114s
grafana-deployment-59fb9b985c-tb772                  1/1     Running            0          57s
k8ssandra-dc1-default-sts-0                          2/2     Running            0          113s
k8ssandra-kube-prometheus-operator-755f6c444-kvgmj   1/1     Running            0          10h
prometheus-demo-prometheus-k8ssandra-0               2/2     Running            1          114s
```
Notice this line:

```
demo-reaper-k8ssandra-schema-7h89x                   0/1     CrashLoopBackOff   3          88s
```
Reaper is configured to use the same Cassandra cluster that is being deployed as its backend. While Reaper manages applying schema changes, it does create the keyspace. 

reaper-operator deploys a job to create the keyspace:

```
$ kubectl get jobs
NAME                           COMPLETIONS   DURATION   AGE
demo-reaper-k8ssandra-schema   1/1           106s       6m29s
```
The CrashLoopBackOff and restarts happen because Cassandra is not up yet, so the job fails. This is fine because the job will continue to be retried and even recreated if necessary until it succeeds. This can be confusing though those for users. If there are deployment problems, and a user sees that error, he might mistakenly think that is the issue.

[...]

PR #53, opened 2020-11-19
Modify reaper configuration to use CassandraDatacenter

burmanm:
This is for issue #45 , requires updated reaper-operator to function correctly.

### DIFF
--- a/charts/k8ssandra-cluster/templates/reaper.yaml
+++ b/charts/k8ssandra-cluster/templates/reaper.yaml
@@ -11,12 +11,9 @@ spec:
     storageType: cassandra
     jmxUserSecretName: {{ .Release.Name }}-reaper-secret-k8ssandra
     cassandraBackend:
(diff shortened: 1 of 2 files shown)


jsanda:
These changes depend on https://github.com/thelastpickle/reaper-operator/pull/6.

jsanda (changes requested):
I pushed an updated image of reaper-operator to Docker Hub.

@burmanm can you please update the image in `charts/k8ssandra-cluster/templates/reaper-operator/deployment.yaml` to the following:

```
image: docker.io/k8ssandra/reaper-operator:73199f440649
```

I guess it could be configurable, but I am fine not having it configurable for now. It is really more of a development setting.

Can you also bump the version numbers in the each `Chart.yaml`? For now all of the charts need to be on the same version.

jsanda (approved):

PR #56, opened 2020-11-20
Add a line to remind to change directory to charts

mabiyou:
For some users who might be new to helm charts, would be better to add a line to remind to change directory to the k8ssandra charts folder before running the installations.

### DIFF
--- a/charts/README.md
+++ b/charts/README.md
@@ -4,6 +4,7 @@ K8ssandra is installed and configured through Helm charts.
 [Helm 3](https://helm.sh/) must be installed to use the charts.
 
 The charts are not yet deployed to a Helm repo. For now you need use the charts directly from this Git repo.
+Or if you cloned a copy of this repo, cd into the charts folder: k8ssandra/charts
 
 # Getting Started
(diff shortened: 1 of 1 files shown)


jsanda:
Hi @mabiyou 

Thanks for PR! 

It looks like the README needs to be updated because the charts are now in fact getting published. See this [section](https://k8ssandra.io/docs/getting-started/#configure-helm-repository) in the docs.

Would you update the README to be consistent with the docs?

mabiyou:
@jsanda that's great news, the [docs](https://k8ssandra.io/docs/getting-started/#configure-helm-repository) look much better now, I think we should update both: https://github.com/k8ssandra/k8ssandra/blob/main/README.md and https://github.com/k8ssandra/k8ssandra/blob/main/charts/README.md to include the installation steps, will do it shortly.

jsanda:
Hi @mabiyou 

Is this PR still needed since you also opened #63?

jdonenine:
This older PR has been overtaken by others so I'll go ahead and close this one.

PR #57, opened 2020-11-20
Update README.md

mabiyou:
Updating the README.md as per the new instructions in https://k8ssandra.io/docs/getting-started/#configure-helm-repository

### DIFF
--- a/charts/README.md
+++ b/charts/README.md
@@ -1,19 +1,54 @@
 # K8ssandra Helm Charts
 K8ssandra is installed and configured through Helm charts.
 
-[Helm 3](https://helm.sh/) must be installed to use the charts.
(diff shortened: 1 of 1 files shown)


jsanda (changes requested):
I prefer to avoid duplicating a lot of the info that is already in the docs at https://k8ssandra.io/docs/getting-started/. The duplication creates extra maintenance. 

How about just a pointer to the docs?

Since we have integrated docs, it might be better to keep the READMEs more developer-focused.

mabiyou:
That's a good ideal, I didn't realize the https://k8ssandra.io/docs/getting-started/ was also under the same repo, for which I added some steps for uninstall and cleanup CRDs.

PR #66, opened 2020-11-23
Added storageClass and PV size to Helm values for external customisation

vmarchese:
The original chart didn't allow for the persistentVolume to be configurable. 
It is particularly useful to allow the configuration for 
- storageClass
- storage request

since the `standard` storageClass is not always present and must be created  (k3d for instance has only support for a local-path provisioner, Openshift is usually installed with glusterfs-storage,...) .

I added the following section in the default values file:

```yaml
name: k8ssandra
clusterName: k8ssandra
datacenterName: dc1
size: 1

k8ssandra:
  # The k8ssandra chart installs operators with cluster scope. It can be installed in a
  # different namespace. The namespace and name of the k8ssandra chart release are needed
  # to configure some resources like the Grafana data source.
  namespace: default
  name: k8ssandra
  persistentVolume:
    storageClass: standard
    size: 5Gi
...
```

allowing the configuration with  standard helm values override.

I didn't touch the `accessModes`  since I assume that `ReadWriteOnce` must not be changed for K8ssandra use case.

### DIFF
--- a/charts/k8ssandra-cluster/templates/cassdc.yaml
+++ b/charts/k8ssandra-cluster/templates/cassdc.yaml
@@ -22,12 +22,16 @@ spec:
   size: {{ .Values.size }}
   storageConfig:
     cassandraDataVolumeClaimSpec:
+{{- if (eq "-" .Values.k8ssandra.persistentVolume.storageClass) }}
       storageClassName: standard
+{{- else }}
+      storageClassName: "{{ .Values.k8ssandra.persistentVolume.storageClass }}"
+{{- end }}        
       accessModes:
         - ReadWriteOnce
       resources:
         requests:
-          storage: 5Gi
+          storage: "{{ .Values.k8ssandra.persistentVolume.size }}"
   config:    
     jvm-options:
       initial_heap_size: "800M"
--- a/charts/k8ssandra-cluster/templates/cassdc.yaml
+++ b/charts/k8ssandra-cluster/templates/cassdc.yaml
@@ -22,16 +22,12 @@ spec:
   size: {{ .Values.size }}
   storageConfig:
     cassandraDataVolumeClaimSpec:
-{{- if (eq "-" .Values.k8ssandra.cassandraLibDirVolume.storageClass) }}
-      storageClassName: standard
-{{- else }}
-      storageClassName: "{{ .Values.k8ssandra.cassandraLibDirVolume.storageClass }}"
-{{- end }}        
+      storageClassName: {{ .Values.k8ssandra.cassandraLibDirVolume.storageClass | default "standard" }}
       accessModes:
         - ReadWriteOnce
       resources:
         requests:
-          storage: "{{ .Values.k8ssandra.cassandraLibDirVolume.size }}"
+          storage: {{ .Values.k8ssandra.cassandraLibDirVolume.size | default "5Gi" }}
   config:    
     jvm-options:
       initial_heap_size: "800M"
--- a/charts/k8ssandra-cluster/templates/cassdc.yaml
+++ b/charts/k8ssandra-cluster/templates/cassdc.yaml
@@ -22,16 +22,16 @@ spec:
   size: {{ .Values.size }}
   storageConfig:
     cassandraDataVolumeClaimSpec:
-{{- if (eq "-" .Values.k8ssandra.persistentVolume.storageClass) }}
+{{- if (eq "-" .Values.k8ssandra.cassandraLibDirVolume.storageClass) }}
       storageClassName: standard
 {{- else }}
-      storageClassName: "{{ .Values.k8ssandra.persistentVolume.storageClass }}"
+      storageClassName: "{{ .Values.k8ssandra.cassandraLibDirVolume.storageClass }}"
 {{- end }}        
       accessModes:
         - ReadWriteOnce
       resources:
         requests:
-          storage: "{{ .Values.k8ssandra.persistentVolume.size }}"
+          storage: "{{ .Values.k8ssandra.cassandraLibDirVolume.size }}"
   config:    
     jvm-options:
       initial_heap_size: "800M"
--- a/charts/backup/Chart.yaml
+++ b/charts/backup/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.14.0
+version: 0.15.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/charts/k8ssandra-cluster/Chart.yaml
+++ b/charts/k8ssandra-cluster/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.14.0
+version: 0.15.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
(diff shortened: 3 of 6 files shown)


johankritzinger (approved):
Closely matches my local branch, as it was required.

jsanda:
Hi @vmarchese 

My apologies for the delay in looking at this. Thanks for the PR! 

I created a design doc to look at how we expose different settings. It is #109. It would be great if you could take a look and see if you have any feedback. How (if at all) might you change this PR in light of that design doc? There is no right or wrong answer. I am just curious.

The `persistentVolume` property name doesn't give any indication as what it is used for. Maybe a comment or a slight name change to indicate it is for `/var/lib/cassandra`.

vmarchese:
> Hi @vmarchese
> 
> My apologies for the delay in looking at this. Thanks for the PR!
> 
> I created a design doc to look at how we expose different settings. It is #109. It would be great if you could take a look and see if you have any feedback. How (if at all) might you change this PR in light of that design doc? There is no right or wrong answer. I am just curious.
> 
> The `persistentVolume` property name doesn't give any indication as what it is used for. Maybe a comment or a slight name change to indicate it is for `/var/lib/cassandra`.

Sure, will do! 
Allow me some time to review the doc (quite busy at the moment with everyday stuff) and I'll change the PR :) 

V.

vmarchese:
> Hi @vmarchese
> 
> My apologies for the delay in looking at this. Thanks for the PR!
> 
> I created a design doc to look at how we expose different settings. It is #109. It would be great if you could take a look and see if you have any feedback. How (if at all) might you change this PR in light of that design doc? There is no right or wrong answer. I am just curious.
> 
> The `persistentVolume` property name doesn't give any indication as what it is used for. Maybe a comment or a slight name change to indicate it is for `/var/lib/cassandra`.

That's a very interesting approach and I think it is useful to users. 
I don't know if the exposed storage configuration could be strictly related to the type of workload since you could have a "read heavy" configuration where you don't need a lot of space (e.g. a geo-distributed lookup map that changes really slowly and you can use a quorum consistency level for writes and local quorum for reads) 

On the other hand, the chart could be smart, as you mentioned. Something like (in pseudo)
```
if values.k8ssandra.workload.type is defined
  if values.k8ssandra.persistentVolume is defined
     persistentVolume  = values.k8ssandra.persistentVolume
  else
     persistentVolume = values.workload.type.persistentVolume
  end
else
    persistentVolume  = default
end
```

This would allow to overriding of the configuration with two levels of priority (i.e. persistentVolume then workload type)

The relative section in the values file would be something like

```yaml
k8ssandra:
  namespace: default
  name: k8ssandra
  workload:
     type: heavy
  # this overrides the default persistentVolume for workload type
  persistentVolume:
      storageClass: standard
      size: 10Gi

# this is injected if the persistentVolume section in the k8assandra configuration is not present
workload:
   heavy: 
      persistentVolume: 
         storageClass: standard
         size: 10Gi
```
I don't know if it makes sense  :)

As for the naming, I agree with you. Have you given a thought about using a tool like [helm-docs](https://github.com/norwoodj/helm-docs) for documenting the charts?

jsanda:
The workload idea is definitely future thinking. For this PR, I would be happy with a name more descriptive than `persistentVolume`.

[helm-docs](https://github.com/norwoodj/helm-docs) looks interesting. I will look into it. Thanks :)

vmarchese:
> The workload idea is definitely future thinking. For this PR, I would be happy with a name more descriptive than `persistentVolume`.
> 
> [helm-docs](https://github.com/norwoodj/helm-docs) looks interesting. I will look into it. Thanks :)

Ok, my doubt is related to the fact that you usually have many  directories in  `/var/lib/cassandra` and, as a matter of fact, they should be separated for a high performance workload (e.g. for sure `data` and `commitlog`). 
Btw, naming things  is not my strong suit... 

What about `cassandraLibDir` ?

jsanda:
> Ok, my doubt is related to the fact that you usually have many directories in `/var/lib/cassandra` and, as a matter of fact, they should be separated for a high performance workload (e.g. for sure `data` and `commitlog`).

Agreed!

There is some work in cass-operator right now for adding support for specifying multiple PVCS. @cscetbon can you share any details on this?

`cassandraLibDirVolume` sounds good to me :)

cscetbon:
> There is some work in cass-operator right now for adding support for specifying multiple PVCS. @cscetbon can you share any details on this?

Yes, along with the cassandra data volume it'll be allowed to add more PVCs in order to separate cassandra volumes or just provide volumes for new containers being added (for things like probes, log analysis, etc...)
An example could be:
```
 storageConfig:
    cassandraDataVolumeClaimSpec: &rw_1g_vol
      storageClassName: server-storage
      accessModes:
        - ReadWriteOnce
      resources:
        requests:
          storage: 1Gi
    additionalVolumes:
      - name: server-logs
        mountPath: /var/log/cassandra
        pvcSpec: *rw_1g_vol
      - name: cassandra-commitlogs
        mountPath: /var/lib/cassandra/commitlog
        pvcSpec: *rw_1g_vol
```

see https://github.com/datastax/cass-operator/pull/330 for more.

vmarchese:
> > There is some work in cass-operator right now for adding support for specifying multiple PVCS. @cscetbon can you share any details on this?
> 
> Yes, along with the cassandra data volume it'll be allowed to add more PVCs in order to separate cassandra volumes or just provide volumes for new containers being added (for things like probes, log analysis, etc...)
> An example could be:
> 
> ```
>  storageConfig:
>     cassandraDataVolumeClaimSpec: &rw_1g_vol
>       storageClassName: server-storage
>       accessModes:
>         - ReadWriteOnce
>       resources:
>         requests:
>           storage: 1Gi
>     additionalVolumes:
>       - name: server-logs
>         mountPath: /var/log/cassandra
>         pvcSpec: *rw_1g_vol
>       - name: cassandra-commitlogs
>         mountPath: /var/lib/cassandra/commitlog
>         pvcSpec: *rw_1g_vol
> ```
> 
> Those volumes would need to be mounted wherever they're used like in the cassandra container. See [datastax/cass-operator#330](https://github.com/datastax/cass-operator/pull/330) for more.

Cool, in the meantime I changed the chart value name to `cassandraLibDirVolume``

jsanda:
@cscetbon thanks for replying and for providing the example. I have a follow up question, but I will ask it in [datastax/cass-operator#330](https://github.com/datastax/cass-operator/pull/330).

jsanda:
@vmarchese can you rebase your branch please?

charts/k8ssandra-cluster/templates/cassdc.yaml
```diff
@@ -22,12 +22,16 @@ spec:
   size: {{ .Values.size }}
   storageConfig:
     cassandraDataVolumeClaimSpec:
+{{- if (eq "-" .Values.k8ssandra.cassandraLibDirVolume.storageClass) }}
```
jsanda, line 25: @vmarchese Can you change to instead check if `storageClass` is undefined/unset?

jsanda (changes requested):

charts/k8ssandra-cluster/templates/cassdc.yaml
vmarchese: It already checks the existence of the value `storageClass`. 
If it is not defined, it defaults to "standard" as before.

So, if the values file is:

```yaml
k8ssandra:
  namespace: default
  name: k8ssandra
...
```

the CR is:
```yaml
apiVersion: cassandra.datastax.com/v1beta1
kind: CassandraDatacenter
metadata:
  name: dc1



Did I get your point ?

Comment (jsanda) [reply charts/k8ssandra-cluster/templates/cassdc.yaml:25]: I interpreted `{{- if (eq "-" .Values.k8ssandra.cassandraLibDirVolume.storageClass) }}` as a check to see if `storageClass` equals `-`. I thought it could something like:

```yaml
{{ .Values.k8ssandra.cassandraLibDirVolume.storageClass | default "standard" }}
```

jsanda (approved):
@vmarchese Thanks for the PR!

PR #91, opened 2020-11-24
[GH 90] make prometheus external url configurable

### LINKED ISSUE #90: Please allow setting the prometheus URL command line option --web.external-url
In order to access prometheus through a proxy, I need to be able to set a path to prometheus instead of using a virtual host. I don't know if this option affects where grafana looks to find the data. If it does, we would need to be able to make grafana work with this modification as well.

jsanda:
Fixes #90 

To test,

```
$ helm install k8ssandra-tools ./k8ssandra
$ helm install prom-config ./k8ssandra-cluster -f prom-values.yaml
```

where `prom-values.yaml` looks like this:

```yaml
monitoring:
  prometheus:
    externalUrl: localhost
```

### DIFF
--- a/charts/backup/Chart.yaml
+++ b/charts/backup/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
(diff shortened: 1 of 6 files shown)


jakerobb (approved):
LGTM.

shallada (approved):

PR #99, opened 2020-11-28
Implement unit tests for templates of CassandraDatacenter and Reaper

burmanm:

PR #100, opened 2020-11-30
Integration test for helm+kind setup

burmanm:
@jsanda Lets discuss more in the Slack. I think we would want more unified testing tools between repositories (reaper-repo, cass-operator, terratest etc .. all doing the same thing differently). Also to verify Grafana we'll need some other approach.

This PR is built as a branch of #99 .


tests/integration/simple_integration_test.go
```diff
@@ -0,0 +1,164 @@
...
+)
+
+var (
+	namespace = "k8ssandra"
+	releaseName string
+	releaseNameCluster string
+)
+
+var _ = Describe("Install the cluster", func() {
+	// TODO This assumes the kind was started with correctly setup cluster. Could we automate it here to provider proper e2e?
```
jsanda, line 27: I just want to point out that while kind is often used during development, we want to maintain the freedom to run against other types of clusters.

tests/integration/simple_integration_test.go
```diff
@@ -0,0 +1,164 @@
...
+			kubectlOptions := k8s.NewKubectlOptions("", "", namespace)
+			options := &helm.Options{KubectlOptions: kubectlOptions}
+
+			releaseName = fmt.Sprintf(
+				"demo-%s", strings.ToLower(random.UniqueId()))
+			helm.Install(GinkgoT(), options, crdChartPath, releaseName)
+
+			// k8s module has no select by label
+			clientset, err := k8s.GetKubernetesClientFromOptionsE(GinkgoT(), kubectlOptions)
+			pods, _ := clientset.CoreV1().Pods(namespace).List(context.Background(), metav1.ListOptions{LabelSelector: "name=cass-operator"})
```
jsanda, line 77: @jeffbanks This is one of the areas that precipitated some of the recent discussions. Here we are using a generated, static client. In medusa-operator and in reaper-operator we use the controller-runtime dynamic client. 

On the one hand it is convenient that the `k8s` package from terratest already initializes the client for us. On the other hand, I prefer to stick with the controller-runtime client throughout. We use it extensively in both production and test code for operators.

tests/integration/simple_integration_test.go
```diff
@@ -0,0 +1,164 @@
...
+
+	BeforeSuite(func() {
+		kubectlOptions := k8s.NewKubectlOptions("", "", "default")
+		options := &helm.Options{KubectlOptions: kubectlOptions}
+
+		// Add traefik repo and update repos
+		helm.RunHelmCommandAndGetOutputE(GinkgoT(), options, "repo", "add", "traefik", "https://helm.traefik.io/traefik")
+		helm.RunHelmCommandAndGetOutputE(GinkgoT(), options, "repo", "update")
+
+		// Deploy traefik
```
jsanda, line 37: I am fine with leaving traefik out for an initial set of tests. In fact, I prefer to leave it out and tackle that separately.

tests/integration/simple_integration_test.go
```diff
@@ -0,0 +1,164 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"net/http"
+	"path/filepath"
+	"strings"
+	"time"
+)
+
+var (
+	namespace = "k8ssandra"
+	releaseName string
```
jsanda, line 22: This is a minor nit, but I find the `releaseNameCluster` variable name a bit confusing. Initially I thought it was the name of the C* cluster and/or CassandraDatacenter. And `releaseName` does not indicate which release/chart it corresponds to. I would prefer something more explicit, albeit a bit verbose, e.g., `k8ssandraReleaseName` and `k8ssandraClusterReleaseName`.

tests/integration/simple_integration_test.go
```diff
@@ -0,0 +1,164 @@
...
+					"ingress.traefik.repair.host": "repair.localhost",
+				},
+				KubectlOptions: k8s.NewKubectlOptions("", "", namespace),
+			}
+			helm.Install(GinkgoT(), options, clusterChartPath, releaseNameCluster)
+
+			// Wait for CassandraDatacenter to be ready..
+			k8s.RunKubectl(GinkgoT(), kubectlOptions, "wait", "--for=condition=Ready", "cassandradatacenter/dc1", "--timeout=300s")
+
+			// Verify all the correct services are there
```
jsanda, line 109: I am not entirely convinced that we need to verify services for the various custom resources. I think the verification as much as possible should focus on the custom resources, i.e., CassandraDatacenter, Reaper, Prometheus. We may need to use the services in some places, and that is fine. I tend though to think about the different resources that an operator creates as an implementation detail detail of the custom resource.

tests/integration/simple_integration_test.go
```diff
@@ -0,0 +1,164 @@
...
+			// Medusa (if enabled)
+
+			// Verify traefik is ready
+			kubectlOptions = k8s.NewKubectlOptions("", "", "traefik")
+			k8s.RunKubectl(GinkgoT(), kubectlOptions, "wait", "--for=condition=Ready", "pod", "-l", "app.kubernetes.io/name=traefik", "--timeout=300s")
+			kubectlOptions = k8s.NewKubectlOptions("", "", namespace)
+
+			// TODO Verify with cqlsh that Cassandra is working properly
+
+			// Verify that prometheus is polling the Cassandra instance correctly
```
jsanda, line 136: I really like the approach you are taking with the Prometheus verification! This would catch things like screwing up the ServiceMonitor or if the metric collector is not deployed with Cassandra.

tests/integration/simple_integration_test.go
```diff
@@ -0,0 +1,164 @@
...
+	BeforeEach(func() {
+		kubectlOptions := k8s.NewKubectlOptions("", "", "default")
+		k8s.CreateNamespace(GinkgoT(), kubectlOptions, "k8ssandra")
+	})
+
+	AfterEach(func() {
+		kubectlOptions := k8s.NewKubectlOptions("", "", namespace)
+		options := &helm.Options{KubectlOptions: kubectlOptions}
+
+		// Verify everything was actually deleted.. I don't think Service is
```
jsanda, line 52: Is this actually verifying that things are deleted?

jsanda (commented):
Based on what I see here, I am not convinced on the need for terratest. We could easily enough write our own thin wrappers for calling Helm. With that said, I prefer to get some automated tests in place first and then address terratest.

tests/integration/simple_integration_test.go
burmanm: It's not, since we know it's not deleted (as there's no deletion job). Some of these comments are meant for "TODO".

tests/integration/simple_integration_test.go
burmanm: Then we need some other method of accessing the cluster resources (such as prometheus). A port-forwarding is another option, but that doesn't really test if we actually have in-cluster working solution or just in-pod.

tests/integration/simple_integration_test.go
burmanm: There's nothing really expecting kind, other than "a cluster must be up" as this test does not spin one.

tests/integration/simple_integration_test.go
jeffbanks: @burmanm 

[This module](https://github.com/k8ssandra/k8ssandra/blob/f90a29de0d19eab0838663a223e54ab6769b240e/tests/integration/util/lib.go) may help where we now have `Ginkgo` & `Omega` into the `Terratest` foundations for k8ssandra operator & cluster integration tests.

One approach may be that we iterate over the resource(s) scoped by namespace and use of test driven resource names.  

For now, I've been targeting the pods and releases.  If we need to go to `crd` and `service` levels is necessary, those should be doable as we can use similar patterns for verification of their existence or non-existence.

An example of using the integration test module:

```
By("Expecting to have the operator and cluster in deployed status")
Ω(IsReleaseDeployed(options.Operator, "k8ssandra")).Should(BeTrue())
Ω(IsReleaseDeployed(options.Cluster, "cassdc")).Should(BeTrue())

By("Expecting to have running cass-operator pod")
Ω(IsPodRunning(options.Operator, "cass-operator")).Should(BeTrue())
```

jeffbanks (approved):
Good use of our standardized test APIs `Gomega`, `Ginkgo`, and `Terratest`.

I made some mention of reuse of our util/module functions included in our [integration test foundations](https://github.com/k8ssandra/k8ssandra/pull/105) PR (pending final review).  

This is something we can do in another iteration as needed.

tests/integration/simple_integration_test.go
```diff
@@ -0,0 +1,164 @@
...
+					"ingress.traefik.monitoring.grafana.host": "grafana.localhost",
+					"ingress.traefik.monitoring.prometheus.host": "prometheus.localhost",
+					"ingress.traefik.repair.host": "repair.localhost",
+				},
+				KubectlOptions: k8s.NewKubectlOptions("", "", namespace),
+			}
+			helm.Install(GinkgoT(), options, clusterChartPath, releaseNameCluster)
+
+			// Wait for CassandraDatacenter to be ready..
+			k8s.RunKubectl(GinkgoT(), kubectlOptions, "wait", "--for=condition=Ready", "cassandradatacenter/dc1", "--timeout=300s")
```
jsanda, line 107: I wasn't really familiar with `kubectl wait` before I saw this. That is really nice :)

tests/template_cassdc_test.go
```diff
@@ -0,0 +1,117 @@
+package tests
```
jsanda, line 1: Is this test the same as the one from #99? If so does it make sense to remove it from this PR?

tests/template_reaper_test.go
```diff
@@ -0,0 +1,72 @@
+package tests
+
+import (
```
jsanda, line 3: Is this test the same as the one from #99? If so does it make sense to remove it from this PR?

jsanda (changes requested):

tests/integration/simple_integration_test.go
```diff
@@ -0,0 +1,164 @@
...
+			kubectlOptions = k8s.NewKubectlOptions("", "", namespace)
+
+			// TODO Verify with cqlsh that Cassandra is working properly
+
+			// Verify that prometheus is polling the Cassandra instance correctly
+			// Wait for the Prometheus to be ready
+			k8s.RunKubectl(GinkgoT(), kubectlOptions, "wait", "--for=condition=Ready", "pod", "-l", "app=prometheus", "--timeout=300s")
+
+			// Poll Prometheus targets and check that it contains our cluster
+			res, err := http.Get("http://prometheus.localhost:8080/api/v1/targets")
```
jsanda, line 141: Should this part be commented out for now?

tests/template_cassdc_test.go
burmanm: Yes, since it's a branch of that PR. A rebase will remove these commits.

tests/template_reaper_test.go
burmanm: Same as above.

tests/integration/simple_integration_test.go
burmanm: ..why?

jsanda: The test failed for me the first time with the Prometheus checks. I assumed there was an issue with ingress. I didn't realize that I was running against a GKE cluster. I just ran again against a kind cluster configured with `./tests/bin/cluster.sh` and everything passed 👌

tests/integration/simple_integration_test.go
jsanda: >At this point I don't have a strong opinion on the method we choose.

Since we already invested in traefik for use with kind and since you already have it working in the test, I changed my mind and now have a stronger opinion. Traefik works for me 👍

tests/integration/simple_integration_test.go
jsanda: @jeffbanks We definitely will have a need for a common test library. I do not want to get too caught focusing on it right now though. I fear doing so may be a bit of a premature optimization considering we currently have limited test automation right across the various repos. I would rather focus more on what scenarios we want to test and getting coverage in place. Then we can iterate and refactor common code as we go.

>For now, I've been targeting the pods and releases.

Take a top down approach. Start with thinking about the various custom resources we are creating.

* How do we verify that a `CassandraDatacenter` object is ready?
* How do we verify that a `Reaper` object is ready?
* How do we verify that a `Grafana` object is ready?
* How do we verify that a `Prometheus` object is ready?

Then let's think about the various resources that are charts are installing or not installing depending on what is enabled. We should able to get unit tests in place fairly easily to cover this stuff. 

We should also have unit tests to make sure labels are being set correctly.

jsanda:
@burmanm can you rebase please?

jsanda (changes requested):
Can you separate the integration and unit tests into separate dirs and then update the `test` target in Makefile to only run unit tests. Then we are good to go!

jsanda (approved):

PR #105, opened 2020-12-01
Integration Testing Foundations

jeffbanks:
Associated w/ efforts started with PR#8, re-based integration test branch with 200+ changes in main.

**Work in-progress:**
- Migration of numerous test changes along with current master functionality.
- Separation and cleanup of test util. module.
- Modification of setup and teardown functions.

**Remaining:**
- Enablement of `cass-cluster` verification post `cass-operator` setup.
- Prometheus cleanup issue found during existing test run.
- Logging reduction where necessary.


tests/integration/cass_integration_test.go
```diff
@@ -0,0 +1,168 @@
...
+	deleteCRDTimeoutSeconds = 15
+
+	k8ssandraNamespace  = "k8ssandra-ns"
+	operatorReleaseName = "k8ssandra-1"
+	clusterReleaseName  = "cluster-1"
+	clusterName         = "dc1"
+)
+
+// TestCassOperator performs basic installation of k8ssandra operator and c* cluster using setup & teardown.
+func TestCassOperatorAndClusterInstall(t *testing.T) {
```
jsanda, line 31: Does this test verify that the CassandraDatacenter is fully initialized, i.e., ready? More than anything else, that is what I want to see in terms of verification.

tests/integration/cass_integration_test.go
```diff
@@ -0,0 +1,168 @@
...
+	var installErr error
+	installResult, installErr := helm.RunHelmCommandAndGetOutputE(t, helmOptions, "install", "-n", k8ssandraNamespace,
+		releaseName, chartName, "--create-namespace", "--insecure-skip-tls-verify")
+
+	util.Log(t, "Helm", "Install", installResult, installErr)
+	return (installErr == nil)
+}
+
+// cleanupOperator cleanup of operator specific resources.
+func cleanupOperator(t *testing.T, helmOptions *helm.Options, releaseName string, identity util.CRDIdentity) {
```
jsanda, line 135: I am not sure I understand the need for this function. Can we just use`helm uninstall`?

I know that `helm uninstall` does not completely remove everything, namely CRDs. That's ok though because that is the actual behavior currently.

tests/integration/cass_integration_test.go
```diff
@@ -0,0 +1,168 @@
...
+		util.DeleteOperatorCRD(t, kubeOptions, identity, deleteCRDTimeoutSeconds)
+		util.DeleteRelease(t, helmOptions, releaseName)
+		util.DeletePodsByNamespace(t, helmOptions)
+		util.DeleteNamespace(t, helmOptions)
+	}
+	uninstall(t, helmOptions, releaseName)
+}
+
+// cleanupCluster cleanup of cluster specific resources.
+func cleanupCluster(t *testing.T, helmOptions *helm.Options, releaseName string, identity util.ClusterIdentity) {
```
jsanda, line 153: Similar to above, can't we just use `helm uninstall?`

jsanda (changes requested):
I have more questions, butI need to make another, longer pass through the code first.

tests/integration/util/lib.go
```diff
@@ -0,0 +1,418 @@
...
+	require.NoError(t, json.Unmarshal(jsonRendered, &renderedMap))
+
+	return ClusterIdentity{
+		ClusterName:            clusterName,
+		RenderedCassdcTemplate: renderedOutput,
+	}
+}
+
+// Annotate of release name, and release namespace, scoped to namespace
+func Annotate(t *testing.T, helmOptions *helm.Options, releaseName string) *AnnotateResult {
```
jsanda, line 79: I am not sure I understand why we need this function. Or to ask another way, why are we overwriting annotations that Helm adds?

tests/integration/cass_integration_test.go
```diff
@@ -0,0 +1,168 @@
...
+func TestCassOperatorAndClusterInstall(t *testing.T) {
+
+	kubeOptions := k8s.NewKubectlOptions("", "", k8ssandraNamespace)
+	helmOptions := &helm.Options{KubectlOptions: kubeOptions}
+
+	fmt.Println("helmOptions namespace is", helmOptions.KubectlOptions.Namespace)
+	require.NotNil(t, helmOptions)
+
+	// operator installation
+	installOperator(t, helmOptions, operatorReleaseName)
```
jsanda, line 40: If I follow the code correctly, it looks like the test does the following and in this order:

* Install k8ssandra chart
* Do some verification of k8ssandra release
* Delete objects installed as part of release
* Install k8ssandra-cluster chart
* Do some verification
* Delete objects installed as part of release

While we currently have the two charts, this is definitely a valid scenario. I don't think it is worth spending time on this scenario though since we are going to combine the charts.

I am not sure I see the need for all of the deletion code. We should be able to do cleanup simply by calling `helm uninstall`, albeit we need to uninstall in the correct order.

tests/integration/cass_integration_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+// cleanupOperator cleanup of operator specific resources.
+func cleanupOperator(t *testing.T, helmOptions *helm.Options, releaseName string, identity util.CRDIdentity) {
+
+	if util.IsNamespaceExisting(t, helmOptions.KubectlOptions, k8ssandraNamespace) {
+
+		kubeOptions := k8s.NewKubectlOptions("", "", k8ssandraNamespace)
+		helmOptions := &helm.Options{KubectlOptions: kubeOptions}
+
+		util.Annotate(t, helmOptions, releaseName)
```
jsanda, line 142: As mentioned elsewhere, we should in general be able to handle deletion/cleanup with `helm uninstall`. We should not have to manually delete stuff.

tests/integration/cass_integration_test.go
```diff
@@ -0,0 +1,168 @@
...
+		util.DeleteDeployment(t, helmOptions, chartName)
+		util.DeleteOperatorCRD(t, kubeOptions, identity, deleteCRDTimeoutSeconds)
+		util.DeleteRelease(t, helmOptions, releaseName)
+		util.DeletePodsByNamespace(t, helmOptions)
```
jsanda, line 146: Pods are almost always owned by some other object, e.g., Deployment, StatefulSet, etc. Deletes are typically cascaded by the owning object.

tests/integration/cass_integration_test.go
```diff
@@ -0,0 +1,168 @@
...
+		util.DeleteNamespace(t, helmOptions)
```
jsanda, line 147: We are calling `DeleteNamespace` in two different functions. I think we want to do this once either at the beginning or at the end of the test, provided the charts are installed in the same namespace.

Even if the charts are installed different namespaces, I would still do that sort of global clean up at the beginning or end of the test.

jsanda (changes requested):
Here is what I would like to see for an initial integration test:

* Install the k8ssandra chart
* Install the k8ssandra-cluster chart
    * Use default setting to keep it simple
* Verify that the CassandraDatacenter reaches the ready state
* Uninstall the k8ssandra-cluster chart
* Verify that the CassandraDatacenter has been deleted
* Uninstall the k8ssandra chart
* Verify that the cass-operator Deployment has been deleted.

tests/integration/cass_integration_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+	assert.NotNil(t, helmOptions.KubectlOptions)
+	assert.NotNil(t, helmOptions)
+
+	util.CreateNamespace(t, helmOptions)
+	clusterIdentity := util.CreateClusterIdentity(t, helmOptions.KubectlOptions, clusterName, releaseName)
+
+	cleanupCluster(t, helmOptions, releaseName, clusterIdentity)
+
+	repoAdd(t, helmOptions, "k8ssandra", "https://helm.k8ssandra.io/")
```
burmanm, line 82: (comment not specific to this line only) This forces the installation to target the repository version? Including the repoUpdates and such. I would assume the integration / e2e-test is something we would want to test with the local versions - without any remote component, so as to verify the changes we've made in a PR for example.

tests/integration/cass_integration_test.go
jeffbanks: I'm including an overall comment that addresses this thought process for why this was used.

tests/integration/cass_integration_test.go
jeffbanks: Updating with a summary comment to address why this was being used.

tests/integration/cass_integration_test.go
jeffbanks: Updating with a summary comment to address why this was created.

tests/integration/util/lib.go
jeffbanks: Updating with a summary comment to address why this was created.

tests/integration/cass_integration_test.go
jeffbanks: The initial set of logic is for the setup and verification is primarily focused on the cass-operator.  There is an issue getting past that step related to Prometheus error being reported upon install.  

Adding to a summary comment for this PR (coming soon), with a deconstructed view of what is being verified.  That is, separating into two units of work for verification of:
- k8ssandra (cass-operator) referenced
- k8ssandra (k8ssandra-cluster)

tests/integration/cass_integration_test.go
jeffbanks: @burmanm this is a good question.
@jsanda this may be better answered by you.  I'm thinking this is a test run by our CI/CD pipeline with access to the repository version.  How do we want to align the versions with the local test executions?

tests/integration/cass_integration_test.go
jeffbanks: @jsanda and @burmanm 

I've scaled-down the issue into a deconstructed example w/ an umbrella cleanup. 

This will help us to better brainstorm an overall approach.  

There's quite a bit going on in the this PR.  My bad.

**Basic Example**
```
kubeOptions := k8s.NewKubectlOptions("", "", "k8ssandra")
helmOptions := &helm.Options{KubectlOptions: kubeOptions}

// Cleanup
uninstallResult, uninstallErr := helm.RunHelmCommandAndGetOutputE(t, helmOptions, "uninstall", "-n", "k8ssandra", "k8ssandra", "k8ssandra/k8ssandra")
util.Log(t, "Helm", "Uninstall", uninstallResult, uninstallErr)

// Install the operator with all the defaults defined by the chart
installResult, installErr := helm.RunHelmCommandAndGetOutputE(t, helmOptions, "install", "-n", "k8ssandra",
	"k8ssandra", "k8ssandra/k8ssandra", "--create-namespace")
util.Log(t, "Helm", "Install", installResult, installErr)

// Basic verification of CRD for operator
lookupResult := util.LookupCRDByName(t, helmOptions, "cassandradatacenters.cassandra.datastax.com")
require.Equal(t, "customresourcedefinition.apiextensions.k8s.io/cassandradatacenters.cassandra.datastax.com",
	lookupResult)
```

**Context**
My understanding is that to perform verification of the installation (either operator or cluster), the previous associated resources (if existing) would need to be cleaned as part of the setup.  For example, if verifying that a CRD is being created properly, the older one would need to be removed. Otherwise, the validation check for its existence would always pass.  

The attempt was to cover the case where a developer has an existing `cassandradatacenters.cassandra.datastax.com` and a clean slate is required.  If that approach is out of scope due to combining of charts, that becomes an unneeded verification, and I agree, we could remove that type of cleanup and verification activities.  However, this may lead to some ugly state conditions for the developer if an operator's CRD exists and they can't remove it. 

The developer may try to wipe it out directly, which seems reasonable.
`kubectl delete crd/cassandradatacenters.cassandra.datastax.com`

In my experience, the response never returns, as a deadlock condition presents itself when finalizers are associated.
- [deadlock](https://github.com/kubernetes/kubernetes/issues/60538#issuecomment-369099998)

**Umbrella uninstall**
Below is a referenced issues I found related to installing without proper cleaning.  This led me down the path of the cleanup operations specific to k8ssandra `operator` and `cluster`.

- [resource exists](https://github.com/helm/helm/issues/7418#issuecomment-590840159)

**Specific Example**
> helm install -n k8ssandra k8ssandra k8ssandra/k8ssandra --create-namespace

```
Error: rendered manifests contain a resource that already exists. Unable to continue with install: ClusterRole "cass-operator" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation 

validation error: key "meta.helm.sh/release-name" must equal "k8ssandra": current value is "k8ssandra-1"; annotation validation error: key "meta.helm.sh/release-namespace" must equal "k8ssandra": current value is "k8ssandra-ns"

```
**Question**
For our integration testing, are we wanting to allow customization in the test for any of the names (namespace and/or release name)?  

If not, then cleanup doesn't need to track for those differences by doing lookups of existing k8ssandra resources.  That would negate the need for value replacements used in the patching steps.  

Regardless, we may need to document how a developer can manually "clean slate" when the need arises.  Otherwise, the umbrella uninstall and install activities will fail due to lingering objects.

tests/integration/cass_integration_test.go
jsanda: I think that the tests should be running against the local chart versions.

Suppose I submit a PR with some chart changes. We want tests running against my changes to verify that there are no regressions.

tests/integration/cass_integration_test.go
jsanda: >For our integration testing, are we wanting to allow customization in the test for any of the names (namespace and/or release name)?

I am not sure I understand. What do you mean by customization?

jeffbanks: > I am not sure I understand. What do you mean by customization?

Allowing the developer to customize their local release names or namespace names.

```
releaseName := "k8ssandra"
namespace := "k8ssandra"
```

tests/integration/util/lib.go
```diff
@@ -0,0 +1,418 @@
...
+
+// LogResults helper for output of map-based result details
+func LogResults(t *testing.T, message string, results map[string]string) {
+	for k, v := range results {
+		logger.Log(t, fmt.Sprintf("%s Id: %s Result: %s", message, k, v))
+	}
+}
+
+// Pause helper responsible for pausing for a specified time duration and interval.
+func Pause(t *testing.T, message string, interval, timeout time.Duration) {
```
burmanm, line 248: Out of interest.. what's wrong with ``time.Sleep(time.Duration)`` ? It blocks the currently executing goroutine as well.

tests/integration/cass_integration_test.go
jsanda: Yes, I would expect that the test should still pass if those values change.

tests/integration/util/lib.go
jeffbanks: Right, nothing wrong with that usage as well.  This was a **util/helper function** in our test util module that assists with using a one-liner in our test setup/teardown where we want to identify why the pause, with a bit more flexibility for interval as needed.  I found that this logging context really helped when sifting through the log details when something isn't working as expected.

An example:
`Pause(t, "Pause for Cleanup Resource Completion", 1, 8*time.Second)`

tests/integration/operator_integration_test.go
```diff
@@ -0,0 +1,69 @@
...
+	require.Len(t, releases, 2, "Expected two k8ssandra namespaced release entries.")
+
+	for _, rel := range releases {
+		if rel.Name == clusterReleaseName {
+			require.True(t, rel.Status == deployedStatus && rel.Namespace == namespace)
+		}
+	}
+}
+
+func TestK8ssandraSetup(t *testing.T) {
```
jsanda, line 60: What I would like to see in this function is verification that the CassandraDatacenter get initialized and is ready.

jsanda (changes requested):
I appreciate the changes and patience. I know that there has been a lot of churn with this. In addition to the CassandraDatacenter verification, I would like to see the tests written using [ginkgo](https://github.com/onsi/ginkgo).

jeffbanks:
> I would like to see the tests written using [ginkgo](https://github.com/onsi/ginkgo).

To clarify this.  Are you asking for removal of Terratest usage completely, or are you requesting the structure of the test follows a Gingko BDD model?

tests/integration/operator_integration_test.go
jeffbanks: Added to latest w/ release and pod verification.

tests/integration/kubectl/kubectl_integration_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/gruntwork-io/terratest/modules/k8s"
+	"github.com/gruntwork-io/terratest/modules/random"
+	"github.com/stretchr/testify/require"
+	authv1 "k8s.io/api/authorization/v1"
+)
+
+// An example of verifying the kubectl is operational and that
+// authorization for basic operations.
+func TestKubectlAuthorization(t *testing.T) {
+
```
jsanda, line 17: I am not sure that this test is necessary. If kubectl isn't configured properly, I think we can pretty much assume stuff will break.

tests/integration/operator/operator_integration_test.go
```diff
@@ -0,0 +1,97 @@
...
+	UninstallRelease(ctx.Cluster, clusterReleaseName)
+	UninstallRelease(ctx.Operator, operatorReleaseName)
+	return ctx
+}
+
+var _ = Describe(suiteName, func() {
+
+	var options K8ssandraOptionsContext
+
+	Describe("baseline required preconditions", func() {
```
jsanda, line 44: Is this scenario redundant since we already have to install the k8ssandra chart in the `k8ssandra cluster integration` scenario below?

tests/integration/operator/operator_integration_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+			InstallChart(options.Operator, operatorChart, operatorReleaseName, "")
+			InstallChart(options.Cluster, clusterChart, clusterReleaseName, "")
+
+			By("Expecting to have the operator and cluster in deployed status")
+			Ω(IsReleaseDeployed(options.Operator, "k8ssandra")).Should(BeTrue())
+			Ω(IsReleaseDeployed(options.Cluster, "cassdc")).Should(BeTrue())
+
+			By("Expecting to have running cass-operator pod")
+			Ω(IsPodRunning(options.Operator, "cass-operator")).Should(BeTrue())
```
jsanda, line 78: Not sure I understand the purpose of the `IsPodRunning` function. It looks like it just checks to see if there are _any_ running pods in the specified namespace. The first and most important thing we want to check is that the CassandraDatacenter has been created and is ready.

jsanda (changes requested):

tests/integration/operator/operator_integration_test.go
jeffbanks: @jsanda 
Yes, really happy to get this one behind us, so feel free to expand a bit if the below doesn't assist with the intention of this test operation.

---

Working backwards a bit, there is a `LookupRunningPods` function in the `test util module` that performs lookup of only running pods scoped by namespaces.  Many of the util functions are there to introduce "building-blocks" for reuse.

https://github.com/k8ssandra/k8ssandra/blob/725090f0fa107586a105ca279876e7114704314f/tests/integration/util/lib.go#L154

```go
result, err := k8s.RunKubectlAndGetOutputE(GinkgoT(), ctx.kubeOptions, "get", "pods", "-n",
		ctx.namespace, "--field-selector=status.phase=Running", "-o", "name", "--no-headers=true")
```
The `IsPodRunning` function invokes the `LookupRunningPods` with a prefix name and in this case, identifies our 'cass-operator'.   It returns a true status when the prefix name is located in that set.  False otherwise.

Example:
`cass-operator-cd9b57568-mw596                        1/1     Running   2          14h`

https://github.com/k8ssandra/k8ssandra/blob/725090f0fa107586a105ca279876e7114704314f/tests/integration/util/lib.go#L121

There is also a verification of running pods at the label level for **cass-operator** as part of the util module.
```go
Ω(IsLabeledPodExisting(options.Operator, "cass-operator")).Should(BeTrue())Ω(IsLabeledPodExisting(options.Operator, "cass-operator")).Should(BeTrue())
```

This test is checking the running pods and the deployed releases specific to the deployed name `cass-operator` which is the CRD kind of `CassandraDataCenter`.

tests/integration/operator/operator_integration_test.go
jeffbanks: Not the entire portion as it verifies that the test setup conditions are valid.  If something changed to the setup function logic, this would catch that.  As we add more test function (edge cases, etc.) I suspect we don't want to perform this to ensure our options are setup correctly.

```go
By("Expecting operator and cluster options to be setup")
Ω(options).ShouldNot(BeNil())
Ω(options.Cluster).ShouldNot(BeNil())
Ω(options.Operator).ShouldNot(BeNil())
```

I agree, we can remove the following portion as this is now covered in the integration with the cass-cluster.

```go
InstallChart(options.Operator, operatorChart, operatorReleaseName, "")
By("Expecting the operator to be in a deployed status")
Ω(IsReleaseDeployed(options.Operator, "k8ssandra")).Should(BeTrue())
```

tests/integration/kubectl/kubectl_integration_test.go
jeffbanks: This is checking for authorization w/ `kubectl` for the specific commands being used.  I'm good with removing if you don't see value.  

The idea was to get the permission checks out of the way in some lightweight integration function as opposed to finding out later in one of the more complicated verifications.

jsanda: I am fine with making the assumption that kubectl is properly configured as a prerequisite step.

tests/integration/operator/operator_integration_test.go
jsanda: >Many of the util functions are there to introduce "building-blocks" for reuse.

I can appreciate the bottom up approach of trying to put some infrastructure in place before we get too far ahead with tests; however, I think we would be better served right now with more of a top down approach with getting more test coverage in place first and then follow up with refactoring common code into shared packages/libraries. 

We have a spent a decent amount of time on this PR, but we still have not gotten to the critical verification step - check that the CassandraDatacenter is ready. 

With respect to the `IsPodRunning` and `LookupRunningPods` functions, they could be combined, e.g., 

```
$ kubectl get pods --field-selector=status.phase=Running -l name=cass-operator 
```

With a lot of these checks, you need to introduce some polling since things are done asynchronously. 

Lastly, we already have code in reaper-operator (that was also used in an old version of medusa-operator) for verifying that a Deployment is ready. Take a look [here](http://github.com/k8ssandra/reaper-operator/blob/d4834fcfac4d901f2269c66f57ed5a0d0b3d17c0/test/framework/framework.go#L128-L128). I am not saying that the code in reaper-operator has to be _the_ approach, but it certainly good to be aware of it and understand it.

jeffbanks: > With a lot of these checks, you need to introduce some polling since things are done asynchronously.

Polling and waits are built-in to the module functions: [wait](https://github.com/k8ssandra/k8ssandra/blob/a2e592fed685583a3849222da640f93cfc46a3a3/tests/integration/util/lib.go#L83) , [pause](https://github.com/k8ssandra/k8ssandra/blob/a2e592fed685583a3849222da640f93cfc46a3a3/tests/integration/util/lib.go#L93) , [pause-impl](https://github.com/k8ssandra/k8ssandra/blob/a2e592fed685583a3849222da640f93cfc46a3a3/tests/integration/util/lib.go#L190)

tests/integration/operator/operator_integration_test.go
jeffbanks: The  InstallChart for operator has been removed from this setup check as part of the latest update.

The setup will continue to verify that the options are created as expected to support the tests scoped to this package.

jeffbanks:
Moved this to WIP GA as I will need to refactor integration foundations with effort to consolidate merge of k8ssandra-operator and k8ssandra-cluster.

jeffbanks:
Additional refactoring will be required for this PR once the [single chart changes](https://github.com/k8ssandra/k8ssandra/pull/116) are fully in-place.

The majority of changes will reside in the existing separation of the cass-operator and cass-cluster models, and expected to be fairly straightforward.

jeffbanks:
@jsanda for this review changes include:

- Single chart refactoring.
- Kind cluster support setup/teardown.
- Merged with current main.

jeffbanks:
This effort is being addressed as part of #204 w/ release 1.0 components and new approach to integration tests.

PR #109, opened 2020-12-03
Cassandra configuration design doc

jsanda:
This is a first pass at a design on how we may want to expose various configuration setting.

### DIFF
--- a/enhancements/cassandra-configuration.md
+++ b/enhancements/cassandra-configuration.md
@@ -1,0 +1,265 @@
+# Overview
+Cassandra has a lot of configuration settings that are exposed in different places like,
+
+* cassandra.yaml
+* jvm.options
+* cassandra-env.sh
+* logback.xml
+
+The purpose of this document is determine what configuration settings to expose in k8ssandra 
+and how. Doing so requires some understanding of how things are done in cass-operator.
+
+**Note:** This doc currently does not take into consideration any differences between 3.11 
+and 4.0.
+
+# Cass Operator
+
+## Topology
+Cass Operator utilizes Cassandra racks to establish anti-affinity for pods. Let's look at an 
+example CassandraDatacenter:
+
+```yaml
+
+apiVersion: cassandra.datastax.com/v1beta1
+kind: CassandraDatacenter
+metadata:
+  name: dc1
+spec:
+  clusterName: multi-rack
+  serverType: cassandra
+  serverVersion: 3.11.7
+  managementApiAuth:
+    insecure: {}
+  size: 9
+  racks:
+  - name: rack-1
+    zone: us-east1-b
+  - name: rack-2
+    zone: us-east1-c
+  - name: rack-3
+    zone: us-east1-d
+  ...
+```
+
+**Note:** Zones will soon be deprecated in favor of free form labels.
+
+We are specifying a 9 node cluster with 3 racks. cass-operator will evenly distribute C* 
+nodes across the racks.
+
+Each rack includes a zone property which corresponds to an availability zone. cass-operator 
+will use affinity and anti-affinity to ensure that pods from each rack are scheduled into 
+the corresponding availability zone and also to ensure that C* pods are not co-located on 
+k8s worker nodes.
+
+Cass Operator always configures the cluster to use `GossipingPropertyFileSnitch`. I am not 
+sure what happens if I try to specify a different value for `endpoint_snitch`, but 
+regardless, this is an example of a setting that we should not expose in k8ssandra.
+
+Now let's look at how to create a multi-DC cluster (multi-region is outside the scope of 
+this doc).
+
+Here is the first DC:
+
+```yaml
+apiVersion: cassandra.datastax.com/v1beta1
+kind: CassandraDatacenter
+metadata:
+  name: dc1
+  namespace: dev
+spec:
+  clusterName: multi-dc
+  ...
+```
+Here is the second DC:
+
+```yaml
+apiVersion: cassandra.datastax.com/v1beta1
+kind: CassandraDatacenter
+metadata:
+  name: dc2
+  namespace: dev
+spec:
+  clusterName: multi-dc
+  ...
+```
+
+We create two CassandraDatacenters that have the same clusterName and same namespace.
+
+If we want common settings across the two DCs, e.g., auth, heap settings, etc., we need
+to specify those settings for each CassandraDatacenter.
+
+## Configuration
+Now let's look at how we specify some settings from cassandra.yaml and jvm.options:
+
+```yaml
+apiVersion: cassandra.datastax.com/v1beta1
+kind: CassandraDatacenter
+metadata:
+  name: dc1
+# ...
+spec:
+# ...
+  # Everything under the config key is passed to the 
+  # cass-config-builder init container that runs on pod creation, and 
+  # then marshalled into config files.
+  config:
+    cassandra-yaml:
+      num_tokens: 8
+      memtable_allocation_type: heap_buffers
+      file_cache_size_in_mb: 1000
+      
+      authenticator: org.apache.cassandra.auth.PasswordAuthenticator
+      authorizer: org.apache.cassandra.auth.CassandraAuthorizer
+      role_manager: org.apache.cassandra.auth.CassandraRoleManager
+      
+      roles_validity_in_ms: 2000
+      permissions_validity_in_ms: 2000
+      credentials_validity_in_ms: 2000
+      
+      # How long before a node logs slow queries. Select queries that 
+      # take longer than this timeout to execute, will generate an
+      # aggregated log message, so that slow queries can be identified. 
+      # Set this value to zero to disable slow query logging. 
+      slow_query_log_timeout_in_ms: 500
+
+    jvm-options:
+      # Set the database to use 14 GB of Java heap
+      initial_heap_size: "14G"
+      max_heap_size: "14G"
+
+      additional-jvm-opts:
+        # As the database comes up for the first time, set system keyspaces to RF=3
+        - "-Dcassandra.system_distributed_replication_dc_names=dc1"
+        - "-Dcassandra.system_distributed_replication_per_dc=3"
+```
+
+I tried to include a number of properties for illustration. There are some however that 
+deserve further discussion.
+
+First, I want to point out that there are number of properties that we do not want to expose 
+including the following:
+
+* storage_port
+* ssl_storage_port
+* native_transport_port
+* seed_provider
+* listen_address
+* native_transport_port
+* rpc_address
+* data_files_directories
+* commitlog_directories
+
+There are a number of properties around authentication/authorization. we may not want to 
+directly expose properties like `authenticator` and `authorizer`. The auth subsystem is 
+pluggable which is great. I can use a custom `IAuthenticator` by specifying the class name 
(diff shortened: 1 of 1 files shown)


enhancements/cassandra-configuration.md
```diff
@@ -0,0 +1,265 @@
...
+      
+      # How long before a node logs slow queries. Select queries that 
+      # take longer than this timeout to execute, will generate an
+      # aggregated log message, so that slow queries can be identified. 
+      # Set this value to zero to disable slow query logging. 
+      slow_query_log_timeout_in_ms: 500
+
+    jvm-options:
+      # Set the database to use 14 GB of Java heap
+      initial_heap_size: "14G"
```
burmanm, line 127: Note to this.. does the customer still set their own properties for the max-memory the container uses? The non-heap part. Should we add some sort of "best effort guess" to this configuration in the future?

jsanda: >does the customer still set their own properties for the max-memory the container uses?

Yes, this should still be configurable. CassandraDatacenter has a `resources` property for specifying CPU and memory resources for the pod. Here is an example:

```yaml
apiVersion: cassandra.datastax.com/v1beta1
kind: CassandraDatacenter
metadata:
  name: dc1
spec:
  clusterName: resources-example
  serverType: cassandra
  serverVersion: 3.11.7
  managementApiAuth:
    insecure: {}
  size: 3
  config:
    jvm-options:
      initial_heap_size: "800m"
      max_heap_size: "800m"
  resources:
    requests:
      cpu: 2
      memory: 2Gi
    limits:
      cpu: 2
      memory: 2G
  storageConfig:
    cassandraDataVolumeClaimSpec:
      storageClassName: standard
      accessModes:
      - ReadWriteOnce
      resources:
        requests:
          storage: 5Gi
```
We definitely want to expose this.

>Should we add some sort of "best effort guess" to this configuration in the future?

That is a great question, and something definitely worth considering. Suppose we add a property to specify the type of workload - ready heavy, write heavy, time series, etc. We could provide more tailored defaults based on what the user specifies. If the user does not specify a workload type, we fall back to the baseline or generic defaults.

enhancements/cassandra-configuration.md
```diff
@@ -0,0 +1,265 @@
...
+still have the ability to set specific GC properties.
+
+# K8ssandra
+## Topology
+```yaml
+k8ssandra:
+  clusterName: multi-rack-multi-dc
+  datacenters:
+  - name: dc1
+    racks:
```
tomereli, line 189: What about the number of nodes in each datacenter? How will `k8ssandra` know how many nodes to assign to each datacenter?
In `CassandraDatacenter.yaml` we have the `size` parameter which translates to the number of Cassandra nodes in the datacenter. AFAIK the cass-operator will distribute the C* node evenly between the racks.
Is the plan of `k8ssandra` to do the same with distributing C* nodes between DCs?

tomereli: Also, please consider the following use case in which I am working on (comments appreciated). In this case it would be nice if we could specify the number of C* nodes per rack.

# k8ssandra with disaggregated storage

## Requirements

- The number of virtual Cassandra racks (failure domains) must be **less than or equal** to the number of disaggregated storage failure domains
- The number of k8s compute physical failure domains can be much higher

## Working Example

**Compute (k8s)**
- 10 racks (k8s failure domains)
- 30 nodes per rack - total 300 nodes

**Disaggregated Storage**
- 3 failure domains

**Storage (Cassandra)**
- 3 virtual racks
- 100 nodes per rack
  - 98 active C* nodes per rack
  - 2 backup C* nodes per rack

### k8s cluster configuration

#### failure domains mapping

Each physical node belongs to a physical failure domain.
In order to answer the first requirement, we map each physical node to a virtual Cassandra rack by grouping physical racks to virtual racks and assigning each k8s node the label `failure-domain.beta.kubernetes.io/zone` accordingly:

| physical rack | physical nodes | virtual rack  | failure-domains.beta.kubernetes.io/zone |
|---|---|---|---|
| rack1 | 1-30 | rack1 | "rack1" |
| rack2 | 31-60 | rack1 | "rack1" |
| rack3 | 61-90 | rack1 | "rack1" |
| rack4 | 91-120 | rack2 | "rack2" |
| rack5 | 121-150 | rack2 | "rack2" |
| rack6 | 151-180 | rack2 | "rack2" |
| rack7 | 181-210 | rack3 | "rack3" |
| rack8 | 211-240 | rack3 | "rack3" |
| rack9 | 241-270 | rack3 | "rack3" |
| rack10 | 271-280 | rack1 | "rack1" |
| rack10 | 281-290 | rack2 | "rack2" |
| rack10 | 291-300 | rack3 | "rack3" |

### k8ssandra cluster configuration

k8ssandra configuration will include the virtual racks we defined earlier (according to the [k8ssandra topology configuration design doc](https://github.com/k8ssandra/k8ssandra/blob/c5410f47413b69e2b57777b75cf68f9d6fe15a7a/enhancements/cassandra-configuration.md#topology-1))

```yaml
k8ssandra:
  clusterName: disaggregated-storage
  size: 294
  datacenters:
  - name: dc1
    racks:
    - name: rack1
      zone: rack1
    - name: rack2
      zone: rack2
    - name: rack3
      zone: rack3
```

This will translate to the following cass-operator CassandraDatacenter.yaml:

```yaml
apiVersion: cassandra.datastax.com/v1beta1
kind: CassandraDatacenter
metadata:
  name: dc1
spec:
  clusterName: disaggregated-storage
  serverType: cassandra
  serverVersion: 3.11.7
  managementApiAuth:
    insecure: {}
  size: 294
  racks:
  - name: rack1
    zone: rack1
  - name: rack2
    zone: rack2
  - name: rack3
    zone: rack3
  ...
```

Which will eventually translate to 3 stateful sets, one per Cassandra virtual failure domain - rack1,rack2 and rack3. The following is the rack1 statefulset example:

disaggregated-storage-dc1-sts-rack1.yaml
```yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: cassandra
spec:
  serviceName: "disaggregated-storage-dc1-service"
  replicas: 98
  template:
    metadata:
      labels:
        app: cassandra
    spec:
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
              - matchExpressions:
                - key: failure-domain.beta.kubernetes.io/zone
                  operator: In
                  valures:
                  - rack1
        podAntiAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            - labelSelector:
              matchExpressions:
                - key: app
                  operator: In
                  values:
                    - cassandra
              topologyKey: failure-domains.beta.kubernetes.io/zone
```

### k8s scheduling

The above StatefulSet NodeAffinity manifest tells Kubernetes to schedule each replica to one of the k8s nodes with label `failure-domain.beta.kubernetes.io/zone=rack1`. The podAntiAffinity rule tells Kubernetes not to place more than one instance of the application in the same k8s node.
The result - each C* node is running on a dedicated k8s node with the rack1 label.

Due to the 10 physical racks to 3 virtual racks mapping, Kubernetes is free to schedule each of the 98 C* nodes of each rack to any one of 100 k8s nodes with that rack's label.

Since each node in virtual rack1 belongs to physical rack1, rack2, or rack3 - it is also guaranteed that each C* node will run on one of the physical failure domains.

### k8s volume provisioning

Each C* node (pod) in the statefulset defines a volume which is used for the C* storage using the predefined `server-storage` storage class.
Kubernetes will provision the volume with the topology information, and the CSI plugin will provision the volume in the corresponding disaggregated storage failure doma, according to the rack number.


### Failure and recovery

Recall that we pre-allocated 100 k8s nodes per Cassandra rack, but set the number of nodes in each rack to 98.
This leaves us with 2 backup nodes for each rack which are used for failure recovery.

#### C* node (pod) failure

Since every C* node runs on a dedicated k8s node, when a single pod fails it will be restarted on the same node.

#### k8s node failure

On node failure, k8s will restart the failed C* node (pod) on one of the 2 backup nodes of that virtual rack.
The PVC of the failed node will be attached to the backup node.

#### Physical compute rack failure

We can't cover up for a whole rack failure in this Cassandra deployment example, so this is left up for Cassandra to handle.
In case of RF=3 murmor3 configuration,  Cassandra will still have 2 replicas in rack2 and rack3 so it will still have quorum.

**Edge case - physical rack10 failure**

Physical rack10 contains 10 nodes from Cassandra failure domain. This means that 10 nodes in each virtual Cassandra rack are down, so in case that Cassandra distributed the 3 replicas between those specific nodes there will be no availability.

To prevent this edge case, instead of splitting all nodes from rack10 between the virtual racks - we can assign them to one of the racks.
This, However, is not possible currently with cass-operator / k8ssandra.

tomereli: **Finer-graned scheduling support to k8ssandra and cass-operator**
Another way to solve this is with the `allowMultipleNodesPerWorker` configuration option from cass-operator - instead of using "backup k8s nodes", we can allow 2 C* nodes to run on the same k8s worker node in some occasions.
Currently, `allowMultipleNodesPerWorker` simply removes the podAntiAffinity rule so that Kubernetes can theoretically schedule much more than 2 C* nodes in the same k8s node, but we can use the new [pod-topology-spread-constraints](https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/) to fine-tune the behavior.

jsanda: > What about the number of nodes in each datacenter? How will `k8ssandra` know how many nodes to assign to each datacenter?

We use a default value of 1. The user will have to set it more generally though.

> In `CassandraDatacenter.yaml` we have the `size` parameter which translates to the number of Cassandra nodes in the datacenter. AFAIK the cass-operator will distribute the C* node evenly between the racks.
> Is the plan of `k8ssandra` to do the same with distributing C* nodes between DCs?

cass-operator does a really good job of trying to ensure there are balanced racks, but it cannot be guaranteed based on what the user specifies. Rather than using the `size` property, I prefer to use something like `nodesPerRack` and `numRacks` which will allow us to ensure balanced racks.

jsanda:
>failure domains mapping
Each physical node belongs to a physical failure domain.
In order to answer the first requirement, we map each physical node to a virtual Cassandra rack by grouping physical racks to virtual racks and assigning each k8s node the label failure-domain.beta.kubernetes.io/zone accordingly:

Assuming I understand correctly, this is what cass-operator already does.

I also want to point out that this [PR](https://github.com/datastax/cass-operator/pull/320) was recently merged in cass-operator to use labels instead of zones in the racks configuration. This will give more flexibility.

enhancements/cassandra-configuration.md
tomereli: Actually you don't need `numRacks` if the user anyway has to define all racks names and labels (and he should). The `nodesPerRack` is also not really needed now that I think about it again - I originally suggested this cause I thought that in some occasions users may prefer to have more nodes in some racks, but after reading more about the caveats of unbalanced racks I take it back. So you can safely ignore my long comment (I can delete it or you can `resolve` it).

steve-todorov:
Hey @jsanda and everybody else,

I'm currently trying to figure out how to pass some custom configuration options down to the cass-operator.
Would you mind letting me know what's the state of this PR? Is there another way to achieve this now?

Issue #68, opened 2020-11-23
Stargate Integration


jdonenine:
**Is your feature request related to a problem? Please describe.**
N/A

**Describe the solution you'd like**
Deploy Stargate (https://github.com/stargate/stargate) as part of K8ssandra to enable easy API integration with the underlying Cassandra cluster.

**Describe alternatives you've considered**
N/A

**Additional context**
Legacy reference:  [Jira](https://k8ssandra.atlassian.net/browse/K8C-76)

PR #115, opened 2020-12-07
#68: added initial Stargate integration

jakerobb:
This is functional, but incomplete. I still need to:
1. Make a few things configurable
2. Add ingress
3. Expose metrics in Grafana

I intend to add those things to this PR before it gets merged.

UPDATE:
* Ingress and metrics will be addressed in separate PRs. Configurability is done, and I've also added startup, liveness, and readiness probes.


charts/k8ssandra-cluster/templates/stargate/stargate.yaml
```diff
@@ -0,0 +1,42 @@
...
+              value: "{{ .Values.clusterName }}"
+            - name: CLUSTER_VERSION
+              value: "3.11"
+            - name: SEED
+              value: "{{ .Values.clusterName }}-seed-service"
+            - name: DATACENTER_NAME
+              value: "{{ .Values.datacenterName }}"
+            - name: RACK_NAME
+              value: "default"
+            - name: ENABLE_AUTH
```
jsanda, line 41: Is this Cassandra authentication? If so I would leave this disabled for now. We are not yet exposing C* auth.

charts/k8ssandra-cluster/templates/stargate/service.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: v1
```
jsanda, line 1: Wrap the template in an if block like this:

```
{{- if .Values.stargate.enabled }}
```

charts/k8ssandra-cluster/templates/stargate/service_monitor.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: monitoring.coreos.com/v1
```
jsanda, line 1: Wrap the template in an if block like this:

```
{{- if .Values.stargate.enabled }}
```

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
```diff
@@ -0,0 +1,42 @@
+---
```
jsanda, line 1: Wrap the template in an if block like this:

```
{{- if .Values.stargate.enabled }}
```

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
```diff
@@ -0,0 +1,42 @@
...
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ .Release.Name }}-k8ssandra-stargate
+  labels:
```
jsanda, line 6: Add `{{ include "k8ssandra-cluster.labels" . | indent 4 }}`

jsanda (changes requested):

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
jakerobb: Yep, it's C* auth. I've changed the default value and left it not overridable in values.yaml. We'll have to remember and come back and expose it here when we add support for auth.

charts/k8ssandra-cluster/templates/stargate/service.yaml
```diff
@@ -0,0 +1,26 @@
...
+kind: Service
+metadata:
+  name: {{ .Release.Name }}-k8ssandra-stargate-service
+  labels:
+{{ include "k8ssandra-cluster.labels" . | indent 4 }}
+spec:
+  type: ClusterIP
+  ports:
+    - name: crud-graphql
```
jsanda, line 10: Can we rename this to `graphql`?

charts/k8ssandra-cluster/templates/stargate/service.yaml
```diff
@@ -0,0 +1,26 @@
...
+      port: 8080
+      protocol: TCP
+    - name: authz-rest
+      port: 8081
+      protocol: TCP
+    - name: crud-rest
```
jsanda, line 16: Can we rename this to `rest`?

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
```diff
@@ -0,0 +1,42 @@
...
+    app: {{ .Release.Name }}-k8ssandra-stargate
+spec:
+  replicas: 2
```
jsanda, line 9: Let's make `replicas` configurable with a default value of 1.

jsanda (changes requested):
Can you add liveness and readiness probes in the deployment? In terms of priorities, I put the health checks ahead of ingress and metrics.

We can add ingress and metrics in follow up PRs.

charts/k8ssandra-cluster/templates/stargate/service.yaml
```diff
@@ -1,4 +1,3 @@
----
```
jakerobb: the linter doesn't like these.

charts/k8ssandra-cluster/templates/stargate/service.yaml
```diff
@@ -9,13 +9,13 @@ metadata:
 spec:
   type: ClusterIP
   ports:
-    - name: crud-graphql
+    - name: graphql
       port: 8080
       protocol: TCP
-    - name: authz-rest
+    - name: authorization
```
jakerobb, line 14: renamed this one too :)

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
```diff
@@ -46,12 +46,11 @@ spec:
             - name: CLUSTER_VERSION
               value: {{ .Values.stargate.clusterVersion | quote }}
             - name: SEED
-              value: {{ .Values.clusterName + "-seed-service" | quote }}
```
jakerobb: Helm didn't like this syntax. I'm not sure why.

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
```diff
@@ -0,0 +1,75 @@
...
+              value: {{ .Values.stargate.clusterVersion | quote }}
+            - name: SEED
+              value: "{{ .Values.clusterName }}-seed-service"
+            - name: DATACENTER_NAME
+              value: {{ .Values.datacenterName | quote }}
+            - name: RACK_NAME
+              value: {{ .Values.rackName | quote }}
+            - name: ENABLE_AUTH
+              value: "false"
+          startupProbe:
```
jsanda, line 61: I think we may want adjust the probes a bit. Here is some output from a 3-node cluster that I just created:

```
$ kubectl get pods
NAME                                                        READY   STATUS             RESTARTS   AGE
cass-operator-86d4dc45cd-95ght                              1/1     Running            0          2m12s
grafana-deployment-69688df955-8bk6f                         1/1     Running            0          31s
k8ssandra-kube-prometheus-operator-755f6c444-vw2fx          1/1     Running            0          2m12s
prometheus-stargate-test-prometheus-k8ssandra-0             2/2     Running            1          84s
stargate-test-dc1-default-sts-0                             2/2     Running            0          84s
stargate-test-dc1-default-sts-1                             1/2     Running            0          84s
stargate-test-dc1-default-sts-2                             1/2     Running            0          84s
stargate-test-grafana-operator-k8ssandra-57cd775b6f-h686r   1/1     Running            0          85s
stargate-test-k8ssandra-stargate-5c84fc5d5f-gd66s           0/1     CrashLoopBackOff   3          85s
stargate-test-reaper-operator-k8ssandra-cdb4c54b9-n8l56     1/1     Running            0          85s
```
In my experience thus far it seems like it can take upwards of 1 minute for a Cassandra pod to become ready (assuming backup/restore disabled). cass-operator starts Cassandra instances serially. For my 3 node cluster, I would expect it to take around 3 minutes before the C* cluster (i.e., CassandraDatacenter) is ready. And this is with reasonable hardware.

With these settings the, we will only see more restarts of Stargate as the C* cluster size increases. I suggest adding an `initialDelaySeconds` for `startupProbe` with maybe a default value of 30 seconds. See how that does. It might make sense to increase the value based on the size of the C* cluster up to some limit.

It might be worthwhile to experiment with increasing the `periodSeconds` property of `startupProbe`. It defaults to 10 seconds. A higher value might help as well.

jsanda (changes requested):

jdonenine (commented):
Higher level question... with these changes should be bumping the version of the overall k8ssandra chart?

https://github.com/k8ssandra/k8ssandra/blob/37ec2067e7aa1cfae8f928458f8e1022ce9bcea0/charts/k8ssandra/Chart.yaml#L18

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
```diff
@@ -0,0 +1,75 @@
...
+    spec:
+      containers:
+        - name: {{ .Release.Name }}-k8ssandra-stargate
+  {{- if eq .Values.stargate.containerImage "auto" }}
+    {{- if eq .Values.stargate.clusterVersion "3.11" }}
+          image: "stargateio/stargate-3_11:v0.0.9"
+    {{- else if eq .Values.stargate.clusterVersion "4.0" }}
+          image: "stargateio/stargate-4_0:v0.0.9"
+    {{- else }}
+          image: "containerImage-auto-did-not-recognize-clusterVersion-{{ .Values.stargate.clusterVersion }}"
```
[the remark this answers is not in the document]
jakerobb: good thinking -- didn't know that existed. :)

jakerobb:
> Higher level question... with these changes should be bumping the version of the overall k8ssandra chart?

Good question. @jsanda do you have thoughts here?

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
jsanda: The templating engine does not have a string concatenation operator. You don't need it though. It can be written as,

```yaml
{{ .Values.clusterName }}-seed-service"
```

The only env vars that need to use the `quote` function are `CLUSTER_VERSION` and `ENABLE_AUTH` assuming it is templated as well.

charts/k8ssandra-cluster/templates/stargate/service.yaml
jsanda: What is the line that causes the problem, and what is the error?

jsanda:
> > Higher level question... with these changes should be bumping the version of the overall k8ssandra chart?
> 
> Good question. @jsanda do you have thoughts here?

Yes, we will bump the chart version prior to merge.

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
jsanda: I discovered while testing on a GKE cluster earlier today that the startup problem may not be available on all clusters. Given this I think we need to only rely on the liveness and readiness probes for now.

We need a couple other changes for the `livenessProbe`:

* Set a default value for `initialDelaySeconds`
* Make `initialDelaySeconds` configurable

It probably makes sense to make the other probe properties configurable as well for consistency.

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
```diff
@@ -0,0 +1,75 @@
...
+              value: {{ .Values.clusterName | quote }}
+            - name: CLUSTER_VERSION
+              value: {{ .Values.stargate.clusterVersion | quote }}
+            - name: SEED
+              value: "{{ .Values.clusterName }}-seed-service"
+            - name: DATACENTER_NAME
+              value: {{ .Values.datacenterName | quote }}
+            - name: RACK_NAME
+              value: {{ .Values.rackName | quote }}
+            - name: ENABLE_AUTH
```
jsanda, line 59: It looks like we need auth enabled by default. I thought that enabling it would require Cassandra auth to be enabled. That does not appear to be the case. 

@dougwettlaufer even with auth disabled, Stargate is still telling me I need an auth token when I try using the REST API. Can you explain that?

The `AuthnTableBasedService` in Stargate queries the `system_auth.roles` table, but based on some initial testing, it looks like we do not have to have auth enabled in Cassandra. cass-operator adds a user in the `roles` table, so that may be sufficient. 

@dougwettlaufer can you confirm whether we need to have auth enabled in Cassandra? We have some work to do before we are ready to fully support auth in k8ssandra so hopefully the answer is no :)

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
```diff
@@ -0,0 +1,75 @@
...
+  template:
+    metadata:
+      labels:
+        app: {{ .Release.Name }}-k8ssandra-stargate
+    spec:
+      containers:
+        - name: {{ .Release.Name }}-k8ssandra-stargate
+  {{- if eq .Values.stargate.containerImage "auto" }}
+    {{- if eq .Values.stargate.clusterVersion "3.11" }}
+          image: "stargateio/stargate-3_11:v0.0.9"
```
jsanda, line 23: These images should be bumped to

* stargateio/stargate-3_11:v1.0.0
* stargateio/stargate-3_11:v1.0.0

jsanda (changes requested):

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
```diff
@@ -0,0 +1,75 @@
...
+              memory: {{ .Values.stargate.resourceLimits.memory | quote }}
+          env:
+            - name: JAVA_OPTS
+              value: {{ .Values.stargate.javaOpts | quote }}
+            - name: CLUSTER_NAME
+              value: {{ .Values.clusterName | quote }}
+            - name: CLUSTER_VERSION
+              value: {{ .Values.stargate.clusterVersion | quote }}
+            - name: SEED
+              value: "{{ .Values.clusterName }}-seed-service"
```
jsanda, line 54: We can use the service name, but it has to be its DNS name which looks like:

```yaml
{{ .Values.clusterName }}-seed-service.{{ .Release.Namespace }}.svc.cluster.local
```

jsanda (changes requested):
After more testing and debugging, I confirmed the following:

* We can specify the seed service for the seed env var.

jsanda (changes requested):
After further testing/debugging I confirmed the following:

* We can specify the cluster seed service for the seed env var
* The Stargate nodes is participating in gossip
* The Stargate node is not reported in `nodetool status` (I think because it does not join the ring)

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
dougwettlaufer: @jsanda as of now you need to have auth enabled for stargate to use the HTTP based services, even if it's just using the cassandra/cassandra superuser. It doesn't necessarily need to be enabled on the C* side though.

charts/k8ssandra-cluster/templates/stargate/service.yaml
jakerobb: I don't think it gave a line number; it said that the object didn't have a name. I believe that the divider causes the linter to see an implicit "object" before it, with no properties at all.

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
jakerobb: I thought it was Go templating, and that as such I could use any valid Go expression. Is that not the case?

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
```diff
@@ -0,0 +1,75 @@
...
+      labels:
+        app: {{ .Release.Name }}-k8ssandra-stargate
+    spec:
+      containers:
+        - name: {{ .Release.Name }}-k8ssandra-stargate
+  {{- if eq .Values.stargate.containerImage "auto" }}
+    {{- if eq .Values.stargate.clusterVersion "3.11" }}
+          image: "stargateio/stargate-3_11:v0.0.9"
+    {{- else if eq .Values.stargate.clusterVersion "4.0" }}
+          image: "stargateio/stargate-4_0:v0.0.9"
```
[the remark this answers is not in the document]
jakerobb: Unsure at this time. That might be a good question for Doug.

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
jakerobb: The `cluster.local` portion of that is configurable in Kubernetes, so we should make that configurable in the template as well. I'm on it.

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
```diff
@@ -0,0 +1,109 @@
...
+                echo "Cassandra is ready. Starting Stargate..."
+      containers:
+        - name: {{ .Release.Name }}-k8ssandra-stargate
+  {{- if eq .Values.stargate.containerImage "auto" }}
+    {{- if eq .Values.stargate.clusterVersion "3.11" }}
+          image: "stargateio/stargate-3_11:v1.0.0"
+    {{- else if eq .Values.stargate.clusterVersion "4.0" }}
+          image: "stargateio/stargate-4_0:v1.0.0"
+    {{- else }}
+      {{- fail tpl "clusterVersion {{ .Values.stargate.clusterVersion }} is not supported." }}
```
jsanda, line 41: When I use a bad `clusterVersion`, `fail` fails :)

```
$ helm template bad-stargate ./k8ssandra-cluster/ -f stargate-values.yaml
Error: template: k8ssandra-cluster/templates/stargate/stargate.yaml:41:10: executing "k8ssandra-cluster/templates/stargate/stargate.yaml" at <fail>: wrong number of args for fail: want 1 got 2
```
It looks like the right syntax is,

```
{{- fail "clusterVersion .Values.stargate.clusterVersion is not supported." }}
```

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
```diff
@@ -0,0 +1,109 @@
...
+    {{- end }}
+  {{- else }}
+          image: {{ required .Values.stargate.containerImage | quote }}
```
jsanda, line 44: The `quote` function is not needed here.

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
```diff
@@ -0,0 +1,109 @@
...
+              cpu: {{ .Values.stargate.cpuReqMillicores | default 200 }}m
+              memory: {{ mul 1.5 (.Values.stargate.heapMB | default 1024) }}Mi
+          env:
+            - name: JAVA_OPTS
+              value: >-
+                -XX:+CrashOnOutOfMemoryError
+                -Xms{{ .Values.stargate.heapMB | default 1024 }}M
+                -Xmx{{ .Values.stargate.heapMB | default 1024 }}M
+            - name: CLUSTER_NAME
+              value: {{ .Values.clusterName | quote }}
```
jsanda, line 86: The `quote` function is not needed here.

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
```diff
@@ -0,0 +1,109 @@
...
+            - name: CLUSTER_VERSION
+              value: {{ .Values.stargate.clusterVersion | quote }}
+            - name: SEED
+              value: "{{ .Values.clusterName }}-seed-service.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain | default "cluster.local" }}"
+            - name: DATACENTER_NAME
+              value: {{ .Values.datacenterName | quote }}
```
jsanda, line 92: The `quote` function is not needed here.

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
```diff
@@ -0,0 +1,109 @@
...
+            - name: RACK_NAME
+              value: {{ .Values.rackName | quote }}
```
jsanda, line 94: The `quote` function is not needed here.

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
```diff
@@ -0,0 +1,109 @@
...
+              - -c
+              - |
+                echo "Waiting for at least one cassandra node to finish bootstrapping..."
+                while ! nslookup {{ .Values.clusterName }}-seed-service.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain | default "cluster.local" }}; do
+                    sleep 5
+                done
+                echo "Cassandra is ready. Starting Stargate..."
+      containers:
+        - name: {{ .Release.Name }}-k8ssandra-stargate
+  {{- if eq .Values.stargate.containerImage "auto" }}
```
jsanda, line 35: Do you think that `containerImage` and `clusterVersion` should be determined by what version of Cassandra is used? I tend to think so. We can address that in a follow up PR.

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
jsanda, line 92: The `quote` function is not needed here.

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
jsanda, line 94: The `quote` function is not needed here.

jsanda (changes requested):

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
jakerobb: Yes -- but since we're hardcoding that in our cass-operator config, I think there's no way to do it?

charts/k8ssandra-cluster/templates/stargate/stargate.yaml
jakerobb: ah, thanks. I forgot to test that. 😬

jsanda (approved):

Issue #60, opened 2020-11-22
Implement a deletion job


jsanda:
**Is your feature request related to a problem? Please describe.**
As mentioned in #59 deletions in Helm are unordered. If we consolidate the k8ssandra and k8ssandra-cluster charts, we need to ensure that the CassandraDatacenter is deleted before cass-operator. 

**Describe the solution you'd like**
We can declare a `pre-delete` hook on the chart. The hook will be configured to run a `Job` that performs ordered deletions of resources. When the user executes `helm uninstall`, Helm will wait to run until the deletion job completes successfully. Really the `Job` only needs to worry about deleting the CassandraDatacenter and any other objects with finalizers for that matter. We can let Helm take care of everything else.

In summary we will need the following:

* A k8s client application with the required deletion logic
    * We will write this in Go
    * We will use the controller-runtime dynamic client
* A Docker image of the application
* A k8s `Job` that runs the image

@jeffbanks I am going to assign this to you since you have been doing related work in #8. I can work with you to get things set up and configured.

jeffbanks:
Appears that this will become resolved with the modifications for merging to a single chart where the deletion job for uninstall is being added as well.

Referencing the PR [here](https://github.com/k8ssandra/k8ssandra/pull/116)

jeffbanks:
@burmanm assigned this to you so you get credit for this one.  Looks like you are addressing with your other ticket.

PR #116, opened 2020-12-07
Merge to single chart directory, add deletion job for uninstall

### LINKED ISSUE #59: Combine the k8ssandra and k8ssandra-cluster charts
The separate, two chart installation is awkward and error-prone.

Initially the k8ssandra chart was supposed to install the stack - Prometheus, Grafana, Reaper, cass-operator, prometheus-operator, etc. The k8ssandra-cluster chart was supposed to be focused on creating and configuring the CassandraDatacenter. We intended for everything or as much as possible in the k8ssandra chart to be cluster-scoped by default. This means you would only run a single instances of cass-operator, Prometheus, Grafana, etc.

We ran into various issues with trying to make things cluster-scoped; consequently, more and more things have been moved from the k8ssandra chart to the k8ssandra-cluster chart. To be clear, we intend to support cluster-scoped deployments of the various resources. We simply did not have time to address those issues prior to KubeCon.

As it currently stands, k8ssandra-cluster only installs cass-operator and prometheus-operator.

There was another reason for the separate charts. cass-operator adds a [finalizer](https://book.kubebuilder.io/reference/using-finalizers.html) to every CassandraDatacenter. Kubernetes will not delete an object until all finalizers are cleared. If cass-operator is deleted before the CassandraDatacenter, deletion is essentially blocked. You have to manually removed the finalizer in order for deletion of the CassandraDatacenter to complete. I mention all of this because there is no ordered deletion with `helm uninstall`. If everything is in the same chart, we could potentially wind up in a band state where deletion of the CassandraDatacenter is blocked.

We can work around the unordered deletion with a `pre-delete` hook in the chart. (See this [section](https://helm.sh/docs/topics/charts_hooks/#hooks-and-the-release-lifecycle) in the Helm docs for details). I will write up a separate ticket for implementing the `pre-delete` hook.

Provided we have a way to cleanly handle deletions, I propose the following for this issue:

* Move all components/templates into the k8ssandra chart
* Do a namespace-scoped install of cass-operator by default
* Do a namespace-scoped install of prometheus-operator by default

With these changes everything will be namespace-scoped by default. In future tickets we can look to add support multi-namespace or cluster-wide scope for the various components.

burmanm:
Merge k8ssandra-cluster and k8ssandra charts, add pre-delete deletion job to provide uninstall.

Fixes #59 and #60


burmanm:
This should be added to the tests in the integration test PRs eventually.

pkg/cleaner/cleaner.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (a *Agent) RemoveResources() error {
+	// Remove CassandraDatacenter (cass-operator should delete all the finalizers and associated resources)
+	if err := a.removeCassandraDatacenter(); err != nil {
+		log.Fatalf("Failed to remove Cassandra cluster(s): %v", err)
+	}
+	return nil
+}
+
+func (a *Agent) removeCassandraDatacenter() error {
```
jsanda, line 43: This is going to delete all CassandraDatacenters in the namespace which could include ones that were not install by the Helm chart. We need to be careful and only delete ones installed by the chart. We probably need to filter release label.

pkg/go.mod
```diff
@@ -0,0 +1,36 @@
+module github.com/k8ssandra
```
jsanda, line 1: I think I prefer to have a single `go.mod` for the whole repo.

charts/k8ssandra/templates/cleaner/role.yaml
```diff
@@ -0,0 +1,73 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
```
jsanda, line 2: I would like to refine the role to only include what is actually needed.

charts/k8ssandra/templates/cleaner/role.yaml
```diff
@@ -0,0 +1,73 @@
...
+      - apps
+    resources:
+      - deployments
+      - daemonsets
+      - replicasets
+      - statefulsets
+    verbs:
+      - '*'
+  - apiGroups:
+      - monitoring.coreos.com
```
jsanda, line 36: I am pretty sure we do not need anything from `monitoring.coreos.com`

charts/k8ssandra/templates/cleaner/role.yaml
```diff
@@ -0,0 +1,73 @@
...
+      - deployments/finalizers
+    verbs:
+      - update
+  - apiGroups:
+      - datastax.com
+    resources:
+      - '*'
+    verbs:
+      - '*'
+  - apiGroups:
```
jsanda, line 56: Is this needed?

charts/k8ssandra/templates/cleaner/role.yaml
```diff
@@ -0,0 +1,73 @@
...
+    resources:
+      - poddisruptionbudgets
+    verbs:
+      - '*'
+  - apiGroups:
+      - cassandra.datastax.com
+    resources:
+      - '*'
+    verbs:
+      - '*'
```
jsanda, line 67: Can the verbs here be limited to get, list, and delete? Might need watch as well.

charts/k8ssandra/templates/cleaner/role.yaml
```diff
@@ -0,0 +1,73 @@
...
+      - persistentvolumeclaims
+      - events
+      - configmaps
+      - secrets
+    verbs:
+      - '*'
+  - apiGroups:
+      - ""
+    resources:
+      - namespaces
```
jsanda, line 23: Is this needed?

charts/k8ssandra/templates/cleaner/role.yaml
```diff
@@ -0,0 +1,73 @@
...
+    verbs:
+      - get
+  - apiGroups:
+      - apps
+    resources:
+      - deployments
```
jsanda, line 29: Are any of these needed?

jsanda (changes requested):

pkg/go.mod
burmanm: This will be easier to add after the tests has been merged.

burmanm:
@jsanda I cherry-picked those two commits.

pkg/go.mod
jsanda: 👍

pkg/cleaner/cleaner.go
```diff
@@ -0,0 +1,90 @@
...
+
+	cassdcapi "github.com/datastax/cass-operator/operator/pkg/apis/cassandra/v1beta1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/tools/clientcmd/api"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
```
jsanda, line 16: Should these be constants?

charts/k8ssandra/templates/cleaner/role.yaml
```diff
@@ -0,0 +1,17 @@
...
+{{ include "k8ssandra-cluster.labels" . | indent 4 }}
+rules:
+  - apiGroups:
+      - cassandra.datastax.com
+    resources:
+      - cassandradatacenters
+    verbs:
+      - get
+      - list
+      - update
```
jsanda, line 15: Is `update` needed?

jsanda (changes requested):
I still need to spend some time testing the changes,  but this looks good! 

We are going to have to do some doc updates as well. I can help with that if you want.

pkg/cleaner/cleaner_test.go
```diff
@@ -0,0 +1,118 @@
+package cleaner
```
jeffbanks, line 1: @burmanm @jsanda 
Providing a [link](https://github.com/k8ssandra/k8ssandra/tree/operator-integration-tests) over to the integration testing foundations underway.   

As part of those iterations, there are quite a few lessons-learned and functions available that we could reference for much of this cleanup.

I was hopeful we could get that PR merged before we got too far in the other areas.  As such, I would be willing to assist with a ticket to get that effort merged with this cleanup test as we move forward.

pkg/main.go
```diff
@@ -0,0 +1,36 @@
...
+	"os"
+
+	"github.com/k8ssandra/cleaner"
+)
+
+var (
+	podNameSpaceEnvVar = "POD_NAMESPACE"
+)
+
+func main() {
```
jeffbanks, line 15: Is this something that the developers will run to perform a local cleanup?

pkg/cleaner/suite_test.go
```diff
@@ -0,0 +1,75 @@
...
+const (
+	timeout  = time.Second * 10
+	interval = time.Millisecond * 250
+)
+
+var cfg *rest.Config
+var k8sClient client.Client
+var testEnv *envtest.Environment
+
+func TestAPIs(t *testing.T) {
```
jeffbanks, line 33: Same comment as above about integration test foundations PR reuse.

jeffbanks (approved):
Overall, LGTM.  I left 3 comments w/ some suggestions and one question about the main.  Other than those, I agree w/ John's feedback for the additional changes.

pkg/main.go
burmanm: This is ran as part of the ``helm uninstall`` 's pre-delete hook

pkg/cleaner/cleaner_test.go
burmanm: I'm not following really, what would this benefit from? It only creates two CassandraDatacenters to the testenv and then runs the cleaner basically.

jeffbanks: The thought is we would have more sameness/consistency in our k8ssandra tests with build-up and reuse of integration test functions (e.g. util module). Maybe something similar for our unit testing.

[Here](https://github.com/k8ssandra/k8ssandra/blob/operator-integration-tests/tests/integration/util/lib.go) is the current version of the `util module` where I've been building some common operations.

I included some setup & helper models as well to manage the various options (operator, cluster, network, etc.) that  many of the integration tests require.

burmanm: Sure, but none of those are really useful for this test. This test isn't running against real cluster.

jsanda: @jeffbanks The test use [envtest](https://github.com/kubernetes-sigs/controller-runtime/tree/master/pkg/envtest). As @burmanm, it is not a real cluster. envtest does deploy etcd and the apiserver, but that is it.

burmanm:
.. don't merge, the rebase broke something.

burmanm:
Closing this in favor of newer PR (safer way to do it).

PR #119, opened 2020-12-09
run unit tests against PRs

jsanda:


.github/workflows/main.yaml
```diff
@@ -108,3 +110,20 @@ jobs:
...
+    needs:
+      - setup
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.15
+        id: go
+      - name: Check out code into the Go module directory
```
burmanm, line 124: We should probably use a cache for the downloaded go modules to improve performance. For example (this is from the github action's page):

```yaml
- uses: actions/cache@v2
  with:
    path: ~/go/pkg/mod
    key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
    restore-keys: |
      ${{ runner.os }}-go-
```

jsanda: >We should probably use a cache

Yes. Thanks for letting me be lazy and showing me what to add :)

.github/workflows/main.yaml
```diff
@@ -108,3 +110,26 @@ jobs:
...
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.15
+        id: go
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
```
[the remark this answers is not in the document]
jsanda: Yup, good catch

burmanm (approved):

PR #124, opened 2020-12-14
Adding routePrefix to url template

jeffbanks:
Inclusion of routePrefix to the url for Grafana datasources.

### DIFF
--- a/charts/k8ssandra-cluster/templates/grafana/datasource.yaml
+++ b/charts/k8ssandra-cluster/templates/grafana/datasource.yaml
@@ -12,6 +12,6 @@ spec:
         timeInterval: 5s
       name: Prometheus
       type: prometheus
-      url: http://{{ .Release.Name }}-prometheus-k8ssandra.{{ .Release.Namespace }}:9090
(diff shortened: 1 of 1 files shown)


charts/k8ssandra-cluster/templates/grafana/datasource.yaml
```diff
@@ -12,6 +12,6 @@ spec:
         timeInterval: 5s
       name: Prometheus
       type: prometheus
-      url: http://{{ .Release.Name }}-prometheus-k8ssandra.{{ .Release.Namespace }}:9090
+      url: http://{{ .Release.Name }}-prometheus-k8ssandra.{{ .Release.Namespace }}:9090/{{ .Values.monitoring.prometheus.routePrefix }}
```
jsanda, line 15: I think we want a check here to see if `routePrefix` is defined.

jeffbanks: Updated with test coverage.

jsanda (changes requested):
The changes look good and the unit test is 👍 

I have one small request. Can you change the Ω functions to use `Expect`? I know it is a bit more verbose but its usage seems more common, and it makes the code more readable.

jsanda (changes requested):
There is one other change needed. You need to bump the chart versions across all the charts.

jsanda (approved):

Issue #89, opened 2020-11-24
ingressroutes are lost during scale up / scale down operations


tomereli:
## Bug Report

**Describe the bug**
k8ssandra-cluster scale up/down reverts changes to ingress

**To Reproduce**
1. Provision a Cassandra cluster according to https://k8ssandra.io/docs/getting-started/
2. Install traefik according to https://k8ssandra.io/docs/topics/ingress/traefik/kind-deployment/#4-install-traefik-via-helm
3. Upgrade the cluster by enabling ingress. Example command:
```
helm upgrade k8ssandra-cluster  ./charts/k8ssandra-cluster --set ingress.traefik.enabled=true --set ingress.traefik.monitoring.grafana.host=grafana.tomer-cass --set ingress.traefik.monitoring.prometheus.host=prometheus.tomer-cass --set ingress.traefik.repair.host=repair.tomer-cass
```
4. Verify access to services UIs (should work) and ingressroutes are in place:
![image](https://user-images.githubusercontent.com/13386849/100151145-70197780-2ea9-11eb-92a6-ade100af68b0.png)

5. Scale up the cluster to 3 nodes and wait for it to be fully operational
6. Scale down the cluster back to 1 node
7. See error - cannot access services. The reason is that the ingressroutes dissapeared.

**Expected behavior**
ingressroutes should not get affected by scale up / scale down operations.

**Screenshots**

**Environment (please complete the following information):**

* Helm charts version info

![image](https://user-images.githubusercontent.com/13386849/100151394-bff83e80-2ea9-11eb-97a0-453811a80947.png)

* Helm charts user-supplied values

![image](https://user-images.githubusercontent.com/13386849/100151485-e027fd80-2ea9-11eb-812f-012cc95bd49e.png)

* Kubernetes version information:

![image](https://user-images.githubusercontent.com/13386849/100151573-fe8df900-2ea9-11eb-9479-89736942e7c8.png)

* Kubernetes cluster kind:
On premises (multiple VM nodes) provisioned using kubeadm

**Additional context**

jsanda:
@jeffbanks Can you take a look at this?

jeffbanks:
Yes, researching and reaching out to get more information regarding commands in steps 5 & 6.

jeffbanks:
Hi @tomereli can you provide some more details for us to further investigate?

What commands did you use for the following steps:
- Scale up the cluster to 3 nodes and wait for it to be fully operational
[...]

PR #129, opened 2020-12-18
Include use of --reuse-values flag for helm update

jeffbanks:
Related to an [issue](https://github.com/k8ssandra/k8ssandra/issues/89) where previously set values are not preserved for `helm update`.

### DIFF
--- a/docs/content/en/docs/topics/provision-a-cluster/_index.md
+++ b/docs/content/en/docs/topics/provision-a-cluster/_index.md
@@ -374,7 +374,10 @@ Notice the value of `size: 1` in cassdc.yaml. This is the Cassandra DataCenter d
 
 To scale up, you could change the `size` to 3. Example with helm:
 
-`helm upgrade k8ssandra-cluster-a k8ssandra/k8ssandra-cluster --set size=3`
+`helm upgrade k8ssandra-cluster-a k8ssandra/k8ssandra-cluster --set size=3 --reuse-values`
+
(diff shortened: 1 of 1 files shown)


docs/content/en/docs/topics/provision-a-cluster/_index.md
```diff
@@ -374,7 +374,10 @@ Notice the value of `size: 1` in cassdc.yaml. This is the Cassandra DataCenter d
 
 To scale up, you could change the `size` to 3. Example with helm:
 
-`helm upgrade k8ssandra-cluster-a k8ssandra/k8ssandra-cluster --set size=3`
+`helm upgrade k8ssandra-cluster-a k8ssandra/k8ssandra-cluster --set size=3 --reuse-values`
+
+Note: using `--reuse-values` to ensure keeping settings from previous `helm upgrade`.
```
burmanm, line 379: I still feel odd that we have to tell people how to use helm, yet I completely think we should. But I wonder, if someone executes the commands in different order, will this bite them later? Say they scaled first and then modified the traefik rules.

jsanda: >I still feel odd that we have to tell people how to use helm

Why is that? I did not have much prior experience with Helm prior to this project. I have been bitten by this very thing :)

>I wonder, if someone executes the commands in different order, will this bite them later? Say they scaled first and then modified the traefik rules.

At least in theory it should not bite them. The various controller involved should making the specified state match the actual state.

jsanda (approved):

PR #141, opened 2020-12-25
Update Reaper to 2.1.3

### LINKED ISSUE #106: Reaper cannot manage cluster after a repair run if Cassandra cluster restarted
## Bug Report

**Describe the bug**
When adding a cluster to Reaper, you have to specify one or more seed hosts, i.e., contact points. I am using the term _seed host_ since that is the terminology that Reaper uses. We specify the name of the CassandraDatacenter headless service which will resolve to the Cassandra pods. 

We want Reaper to store the name of the service rather than the pod IPs since the pod IPs are not stable. We set the `REAPER_ENABLE_DYNAMIC_SEED_LIST` environment variable to `false` in the Reaper deployment. This causes Reaper to persist the service name instead of the endpoints. The relevant code can be found [here](http://github.com/thelastpickle/cassandra-reaper/blob/fd1427f879cea15567e3915ed95a332ee5f78d1e/src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java#L254-L254).

In the Cassandra backend, we will then have a row like this:

```
cqlsh> select * from reaper_db.cluster ;

 name        | last_contact                    | partitioner                                 | properties       | seed_hosts                          | state
-------------+---------------------------------+---------------------------------------------+------------------+-------------------------------------+--------
 reaper-test | 2020-12-02 00:00:00.000000+0000 | org.apache.cassandra.dht.Murmur3Partitioner | {"jmxPort":7199} | {'reaper-test-reaper-test-service'} | ACTIVE
```
In a test environment that I had up and running for a while, I found that Reaper could no longer make JMX connections to any of the Cassandra nodes after I had done a rolling restart of the Cassandra cluster. A rolling restart means Cassandra pods will get new IPs. That should not be a problem provided Reaper is using the k8s service to resolve endpoint addresses.

I took a look at my test cluster and here is what I found:

```
cqlsh:reaper_db> select * from reaper_db.cluster;

 name | last_contact                    | partitioner                                 | properties       | seed_hosts                                    | state
------+---------------------------------+---------------------------------------------+------------------+-----------------------------------------------+--------
 demo | 2020-12-01 00:00:00.000000+0000 | org.apache.cassandra.dht.Murmur3Partitioner | {"jmxPort":7199} | {'10.244.1.82', '10.244.2.54', '10.244.3.89'} | ACTIVE
```
IP addresses are stored instead of the CassandraDatacenter service. I think that if a subset of the Cassandra nodes are restarted, Reaper can recover. It can connect to a node that has not restarted and will get the changed addresses. I am not 100% certain that this is the case, but I believe I observed this behavior some time ago. 

If however the whole cluster is restarted, Reaper cannot make any JMX connections, which means it cannot manage the cluster.

How did I wind up with IP addresses in the `seed_hosts` column? When Repair does a repair run, it updates the row if the nodes being repaired do not match what is stored in `seed_hosts`. The relevant code can be found [here](http://github.com/thelastpickle/cassandra-reaper/blob/fd1427f879cea15567e3915ed95a332ee5f78d1e/src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java#L298-L298)

**To Reproduce**
1. Install a k8ssandra-cluster with repair auto scheduling enabled.

```
$ helm install k8ssandra k8ssandra/k8ssandra

$ helm install repair-test k8ssandra/k8ssandra-cluster -f repair-values.yaml
```

where `repair-values.yaml` looks like this:

```yaml
name: repair-test
clusterName: repair-test
size: 3
repair:
  reaper:
    autoschedule: true
```

2. Verify that the cluster is added in Reaper with the k8s service name. Use a command like this:

```
$ kubectl -it <cassandra-pod> -c cassandra -- cqlsh -e "select seed_hosts from reaper_db.cluster where name = 'repair-test'"
```

3. Do a rolling restart of the Cassandra cluster. This can be done in multiple ways. One way is to update the CassandraDatacenter spec with `kubectl edit cassdc repair-test` and add `rollingRestartRequested: true` to the spec.
 
4. After the cluster restart is complete, wait for a repair to run or manually schedule one.

5. When the repair finishes, rerun the query from step 2. The `seed_hosts` column should not contain IP addresses.  

**Expected behavior**
Reaper should be able to manage a cluster after the cluster is restarted.

* Helm charts version info
```
$ helm ls -A
NAME        	NAMESPACE	REVISION	UPDATED                                	STATUS  	CHART                   	APP VERSION
demo        	default  	1       	2020-12-01 17:25:22.459807945 +0000 UTC	deployed	k8ssandra-cluster-0.13.0	3.11.7
demo-backup 	default  	1       	2020-12-01 16:45:00.818197619 +0000 UTC	deployed	backup-0.13.0           	0.1.0
demo-restore	default  	1       	2020-12-01 16:46:56.367255906 +0000 UTC	deployed	restore-0.13.0          	0.1.0
k8ssandra   	default  	1       	2020-12-01 17:25:07.72605783 +0000 UTC 	deployed	k8ssandra-0.13.0        	3.11.7
traefik     	traefik  	1       	2020-12-01 16:26:57.231930337 +0000 UTC	deployed	traefik-9.11.0          	2.3.3
```

```
$ helm get values demo
USER-SUPPLIED VALUES:
clusterName: demo
repair:
  reaper:
    autoschedule: true
size: 3
```
```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.0", GitCommit:"e19964183377d0ec2052d1f1fa930c4d7575bd50", GitTreeState:"clean", BuildDate:"2020-08-26T14:30:33Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-06-03T04:00:21Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```

* Kubernetes cluster kind:
kind

**Additional context**
This is going to require changes in Reaper. K8ssandra is using Reaper 2.0.5. I believe the latest release is 2.1.2. We will want to upgrade Reaper at some point, so I do not think it makes sense to try and back port any changes.

burmanm:
Fixes #106

### DIFF
--- a/charts/backup/Chart.yaml
+++ b/charts/backup/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/).
-version: 0.17.1
+version: 0.18.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/charts/k8ssandra-cluster/Chart.yaml
+++ b/charts/k8ssandra-cluster/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.17.1
+version: 0.18.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/charts/k8ssandra-cluster/templates/reaper.yaml
+++ b/charts/k8ssandra-cluster/templates/reaper.yaml
@@ -6,7 +6,7 @@ metadata:
   labels:
 {{ include "k8ssandra-cluster.labels" . | indent 4 }}
 spec:
-  image: thelastpickle/cassandra-reaper:2.0.5
+  image: thelastpickle/cassandra-reaper:2.1.3
   serverConfig:
     {{- if .Values.repair.reaper.autoschedule }}
     autoScheduling:
--- a/charts/k8ssandra/Chart.yaml
+++ b/charts/k8ssandra/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.17.1
+version: 0.18.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/charts/restore/Chart.yaml
+++ b/charts/restore/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.17.1
+version: 0.18.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/scripts/helm-version.py
+++ b/scripts/helm-version.py
@@ -13,9 +13,12 @@ class Semver:
 
     def incr_major(self):
         self.major = self.major + 1
+        self.patch = 0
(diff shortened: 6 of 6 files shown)


jsanda (changes requested):
@burmanm can you go ahead and bump the chart versions?

PR #157, opened 2021-01-07
Fix layout issues of home screen cards

jdonenine:
* Small layout issue in the docs that Chris mentioned to me earlier
* Core of the problem is FF's sizing of images within flex containers, it behaves differently than Chrome and Safari (I didn't try Edge)
* The template that we use includes these cards, which get a flex display applied to them, in this particular case there is no necessity for that, so I've forced the display in those cased only back to the default.
* This same change also fixed a wrapping issue with the cards that I noticed - that happens on all browsers, not just FF

FF Icon Issue Before and After:

image

Card Wrapping Issue Before and After:

image

### DIFF
--- a/docs/assets/scss/_variables_project.scss
+++ b/docs/assets/scss/_variables_project.scss
@@ -98,6 +98,7 @@ Add styles or override variables from the theme here.
       border-radius: 0;
       border: none;
       box-shadow: 0px 14px 32px rgba(0, 0, 0, 0.08);
+      display: block;
 
       img {
         margin-top: 31px;
--- a/docs/content/en/_index.md
+++ b/docs/content/en/_index.md
@@ -53,7 +53,7 @@ linkTitle: K8ssandra
 	<main role="main" class="td-main">
 		<div class="row">
 			<section class="col">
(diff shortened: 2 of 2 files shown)


bradfordcp:
Can we get this change in `docs/assets/scss/_variables_project.scss` instead of a `style` attribute?

bradfordcp:
@jdonenine I pushed the requested change. Can you give this a once over?

jdonenine:
> @jdonenine I pushed the requested change. Can you give this a once over?

I went the original route because I wasn't sure that those classes weren't used anywhere else, but I just did a quick search and while the td-main parent is used a bunch of places, the cards aren't, so shouldn't both anything elsewhere. 👍

jdonenine (commented):
👍

jdonenine:
In an irony of branch protections... I reviewed the change you submitted, but I can't approve and merge, because I originally authored the PR, so you'll have to do it @bradfordcp 😄

bradfordcp:
lol, I wanted somebody else to look at it. I'll move it forward now.

bradfordcp (approved):

Issue #156, opened 2021-01-07
Make resource requests and limits configurable for Cassandra pods


jsanda:
**Is your feature request related to a problem? Please describe.**
It is currently not possible to set resource requests and limits for Cassandra pods.

**Describe the solution you'd like**
To start I would like the resource requests and limits to be configurable at the cluster level as follows:

```yaml
# example values.yaml
k8ssandra:
  resources:
    requests:
      cpu: 4
      memory: 16Gi
    limits:
      cpu: 8
      memory: 16Gi
```

When we add multi-DC support we may want to make this configurable at the DC level as well.

**Describe alternatives you've considered**

**Additional context**

PR #160, opened 2021-01-07
Allow setting resources and allowMultipleNodesPerWorker in the values

florissmit10:
Setting allowMultipleNodesPerWorker to true is only allowed when resources are set.


jsanda:
Hi @florissmit10 

Thank you for submitting the PR! Before we can accept any contributions, we need you to sign the CLA at https://cla.datastax.com/.

Thanks

John

jsanda:
Fixes #156

charts/k8ssandra-cluster/templates/cassdc.yaml
```diff
@@ -28,6 +28,11 @@ spec:
       resources:
         requests:
           storage: {{ .Values.k8ssandra.cassandraLibDirVolume.size | default "5Gi" }}
+  allowMultipleNodesPerWorker: {{ .Values.allowMultipleNodesPerWorker | default false}}
```
jsanda, line 31: Can you move these properties under the top-level `k8ssandra` property?

charts/k8ssandra-cluster/values.yaml
```diff
@@ -6,6 +6,8 @@ name: k8ssandra
 clusterName: k8ssandra
 datacenterName: dc1
 size: 1
+allowMultipleNodesPerWorker: false
```
jsanda, line 9: If we have `allowMultipleNodesPerWorker: true`, should there be a validation check for resources being defined?

jsanda (changes requested):
Can you also please add/update unit tests?

charts/k8ssandra-cluster/values.yaml
florissmit10: At this time the validation of the CassandraDatacenter crd gives a pretty clear error message when you fail to set the resources:
`Error: admission webhook "cassandradatacenter-webhook.cassandra.datastax.com" denied the request: CassandraDatacenter write rejected, attempted to use multiple nodes per worker without cpu and memory requests and limits`

I could add some conditional statements but it would make the template bigger, and the error message less clear I think. Maybe a comment in the default values.yaml?

jsanda: The error message from the webhook is great. The problem is that the chart install still happens and other objects still get created. The user would have to `helm uninstall` before trying the install again. We could use the `required` template function to make sure `resources` is defined. It would fail fast before anything gets deployed.

florissmit10: Ah yeah good point!

charts/k8ssandra-cluster/templates/cassdc.yaml
```diff
@@ -28,6 +28,17 @@ spec:
       resources:
         requests:
           storage: {{ .Values.k8ssandra.cassandraLibDirVolume.size | default "5Gi" }}
+  allowMultipleNodesPerWorker: {{ .Values.k8ssandra.allowMultipleNodesPerWorker | default false}}
+{{- if .Values.k8ssandra.allowMultipleNodesPerWorker}}
+  resources:
+    limits: 
+{{ toYaml (required "set resource limits/requets when enabling allowMultipleNodesPerWorker" .Values.k8ssandra.resources.limits) | indent 6}}
```
jsanda, line 35: Mispelling at "requets"

tests/unit/template_cassdc_test.go
```diff
@@ -130,5 +133,42 @@ var _ = Describe("Verify CassandraDatacenter template", func() {
 			// Two containers, medusa and cassandra
 			Expect(len(cassdc.Spec.PodTemplateSpec.Spec.Containers)).To(Equal(2))
 		})
+
+		It("setting allowMultipleNodesPerWorker to true", func() {
```
jsanda, line 137: Can you add two more unit tests - one where `allowMultipleNodesPerWorker` is `true` and `resources` are not specified and one where `allowMultipleNodesPerWorker` is `false` and `resources` is set?

tests/unit/template_cassdc_test.go
```diff
@@ -130,5 +133,42 @@ var _ = Describe("Verify CassandraDatacenter template", func() {
...
+					"k8ssandra.resources.limits.cpu":      "1",
+					"k8ssandra.resources.requests.memory": "2Gi",
+					"k8ssandra.resources.requests.cpu":    "1",
+				},
+				KubectlOptions: defaultKubeCtlOptions,
+			}
+
+			renderTemplate(options)
+
+			Expect(*cassdc.Spec.Resources.Limits.Memory()).To(Equal(resource.MustParse("2Gi")))
```
jsanda, line 167: 👍

jsanda (changes requested):

tests/unit/template_cassdc_test.go
```diff
@@ -124,11 +135,101 @@ var _ = Describe("Verify CassandraDatacenter template", func() {
...
 			// Initcontainer should only have jmx and jolokia
 			Expect(len(cassdc.Spec.PodTemplateSpec.Spec.InitContainers)).To(Equal(3))
 			// Two containers, medusa and cassandra
 			Expect(len(cassdc.Spec.PodTemplateSpec.Spec.Containers)).To(Equal(2))
 		})
+
+		It("setting allowMultipleNodesPerWorker to true", func() {
+			options := &helm.Options{
+				SetValues: map[string]string{
+					"k8ssandra.allowMultipleNodesPerWorker": "true",
```
jsanda, line 149: Is this is a duplicate test of the one at line 184, minus the resource verification? If so, can you remove it?

jsanda (changes requested):
There is one other thing needed for the PR. Can you please bump the chart versions to 0.20.0 in each of the 4 charts?

tests/unit/template_cassdc_test.go
```diff
@@ -33,13 +34,23 @@ var _ = Describe("Verify CassandraDatacenter template", func() {
 		err = nil
 	})
 
-	renderTemplate := func(options *helm.Options) {
-		renderedOutput := helm.RenderTemplate(
+	renderTemplateE := func(options *helm.Options) error {
```
jsanda, line 37: I have a similar change in #178 :)

tests/unit/template_cassdc_test.go
```diff
@@ -33,13 +34,23 @@ var _ = Describe("Verify CassandraDatacenter template", func() {
...
+		renderedOutput, renderError := helm.RenderTemplateE(
```
jsanda, line 38: This could be simplified a bit as,

```go
		renderedOutput, err := helm.RenderTemplateE(
			GinkgoT(), options, helmChartPath, helmReleaseName,
			[]string{"templates/cassdc.yaml"},
		)

		if err == nil {
			err = helm.UnmarshalK8SYamlE(GinkgoT(), renderedOutput, cassdc)
		}

		return err
```

Because we using functions that return an error instead of panic, we need to update all the `renderTemplate` that should succeed to calls like this:

```go
Expect(renderTemplate(options)).To(Succeed())
```

jsanda (changes requested):

tests/unit/template_cassdc_test.go
florissmit10: Ah yeah that looks good!
First time programming Go for me, so these patterns are a bit new :P

jsanda (changes requested):
Some other PRs have been merged before yours, so you need to bump the chart versions to 0.22.0. After that I think we are ready to merge :)

jsanda (approved):
Thanks for the PR!

PR #162, opened 2021-01-07
Property cleanup

jsanda:
This PR removes unused properties and moves properties under the top-level `k8ssandra` property.


jeffbanks (approved):
Looks good.

PR #179, opened 2021-01-09
initial support for enabling Cassandra authentication and authorization

jsanda:


charts/k8ssandra-cluster/templates/cassdc.yaml
```diff
@@ -33,10 +34,31 @@ spec:
...
+    cassandra-yaml:
+{{- if .Values.k8ssandra.configuration.auth }}
+      authenticator: PasswordAuthenticator
+      authorizer: CassandraAuthorizer
+{{- else }}
+      authenticator: AllowAuthenticator
+      authorizer: AllowAuthorizer
+{{- end }}
+      role_manager: CassandraRoleManager
+      roles_validity_in_ms: {{ .Values.k8ssandra.configuration.auth.caches.rolesValidityMillis }}
```
burmanm, line 47: Is this intended that even if auth is set to false we still push these values?

charts/k8ssandra-cluster/templates/cassdc.yaml
```diff
@@ -33,10 +34,31 @@ spec:
       resources:
         requests:
           storage: {{ .Values.k8ssandra.cassandraLibDirVolume.size | default "5Gi" }}
-  config:    
+  config:
+    cassandra-yaml:
+{{- if .Values.k8ssandra.configuration.auth }}
```
burmanm, line 39: .auth.enabled?

tests/unit/template_cassdc_test.go
```diff
@@ -171,6 +173,66 @@ var _ = Describe("Verify CassandraDatacenter template", func() {
...
+			clusterSize := 3
+			authCachePeriod := int64(7200000)
+			options := &helm.Options{
+				KubectlOptions: defaultKubeCtlOptions,
+				SetValues: map[string]string{
+					"k8ssandra.datacenterName":                                      dcName,
+					"k8ssandra.size":                                                strconv.Itoa(clusterSize),
+					"k8ssandra.configuration.auth.enabled":                          "true",
+					"k8ssandra.configuration.auth.caches.rolesValidityMillis":       strconv.FormatInt(authCachePeriod, 10),
+					"k8ssandra.configuration.auth.caches.rolesUpdateMillis":         strconv.FormatInt(authCachePeriod, 10),
```
burmanm, line 187: Should these be `+1`, `+2` etc to verify that we actually set the correct value in the helm chart? At the moment this wouldn't catch it if there's a copy-paste error from another line.

charts/k8ssandra-cluster/templates/cassdc.yaml
jsanda: Yes, it is intended so as to avoid more conditional logic. The corresponding properties are always set in cassandra.yaml.

jsanda:
@burmanm This is ready for review. The organization of the auth related properties in values.yaml is likely going to change as we have been discussing outside of this PR. I am fine with addressing those changes in a follow up PR once we settle on how we want the configs to look.

burmanm (approved):

PR #188, opened 2021-01-12
cass-operator upgrade

### LINKED ISSUE #112: Upgrade cass-operator to > 1.5.0
k8ssandra is currently using a patched version of cass-operator that is between 1.4.1 and 1.5.0. The patches did not land in 1.5.0.

I want to rebase our patched version on top of 1.5.0 so that we can take advantage of scale down. The version we are using does not support scale down.

I will try to get the patches into the next release of cass-operator.

https://github.com/datastax/cass-operator/packages/174279 has the changes that we need.

jsanda:
Fixes #112

Making this a draft PR until we are able to switch over to an official datastax/cass-operator image. We are just waiting on images to be pushed.

### DIFF
--- a/charts/k8ssandra/charts/cass-operator/templates/webhook-service.yaml
+++ b/charts/k8ssandra/charts/cass-operator/templates/webhook-service.yaml
@@ -6,7 +6,7 @@ metadata:
     name: cass-operator-webhook
 spec:
   ports:
-  - port: 443
-    targetPort: 8443
+    - port: 443
+      targetPort: 8443
   selector:
     name: cass-operator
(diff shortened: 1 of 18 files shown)


charts/k8ssandra/charts/cass-operator/templates/webhook-service.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
     name: cass-operator-webhook
 spec:
   ports:
-  - port: 443
-    targetPort: 8443
+    - port: 443
```
burmanm, line 9: These type of changes make me want the lint test to check for correct (and defined) yaml indentation. Would make it a lot simpler to read these PRs if each one didn't redo some sort of indenting.

burmanm (approved):

PR #197, opened 2021-01-15
cass-operator chart

### LINKED ISSUE #113: Make cass-operator namespace-scoped by default
I want to make this change in conjunction with #59. cass-operator and prometheus-operator are configured with cluster-wide scope. Everything else is namespace-scoped. We want everything (if possible) to be namespace-scoped.

jsanda:
Fixes #113. 

#188 needs to be merged first.

This PR does not include support for:

* cluster-scoped deployments
* validating webhook


charts/cass-operator/templates/deployment.yaml
```diff
@@ -1,60 +1,45 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: {{ .Values.deploymentName }}
+  name: {{ include "cass-operator.fullname" . }}
+  labels:
+    {{- include "cass-operator.labels" . | nindent 4 }}
```
burmanm, line 6: Now that we would manage the cass-operator, should we include the k8ssandra-labels also?

charts/cass-operator/values.yaml
```diff
@@ -0,0 +1,66 @@
...
+  # TODO It better to keep this commented out or set it to empty string?
+  # registryOverride:
+
+imagePullSecrets: []
+nameOverride: ""
+fullnameOverride: ""
+
+
+serviceAccount:
+  # TODO Should we have a create property? Don't we always want to create a service account?
```
burmanm, line 24: It should be always generated.

charts/cass-operator/values.yaml
```diff
@@ -0,0 +1,66 @@
...
+  # Specifies whether a service account should be created
+  create: true
+  # Annotations to add to the service account
+  annotations: {}
+  # The name of the service account to use.
+  # If not set and create is true, a name is generated using the fullname template
+  name: ""
+
+# TODO Do we need this for an operator deployment?
```
burmanm, line 33: Are there some other tools that might want it? I think at least datadog uses podAnnotations for some health scraping. I don't know why the choice between podAnnotation vs. annotations in general. Also, should that sort of parameter be at top level (add additional annotations to every pod / every container / etc) instead of having to define it for every part separately?

charts/cass-operator/values.yaml
```diff
@@ -0,0 +1,66 @@
...
+podAnnotations: {}
+
+# TODO Do we need this for an operator deployment?
+podSecurityContext: {}
```
burmanm, line 37: I'll go with the same opinion as above, I think one securityContext per deployment of k8ssandra should be enough. I'm struggling to think why some parts would have different securityContexts.

charts/cass-operator/values.yaml
```diff
@@ -0,0 +1,66 @@
...
+  # resources, such as Minikube. If you do want to specify resources, uncomment the following
+  # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
+  # limits:
+  #   cpu: 100m
+  #   memory: 128Mi
+  # requests:
+  #   cpu: 100m
+  #   memory: 128Mi
+
+vmwarePSPEnabled: false
```
burmanm, line 64: Just a note in general, I don't like this structure of configuration - even if we can't modify it right now.

charts/k8ssandra/Chart.yaml
```diff
@@ -25,11 +25,9 @@ version: 0.25.0
 appVersion: 3.11.7
 dependencies:
   - name: cass-operator
-    version: 1.5.0
-    # The cass-operator chart lives in https://github.com/datastax/cass-operator. There are
-    # changes in master that are needed for K8C-19. For now we will sync (i.e., copy) the
-    # cass-operator chart into the charts subdirectory.
-    #repository: https://github.com/datastax/cass-operator/tree/master/charts
+    version: 0.25.0
```
burmanm, line 28: Separate discussion, but related to yesterday's meeting. If we split to sub charts, do they have to always follow the same version as parent? I'd like to separate so it becomes easier to track subchart version history. We can discuss in Slack.

charts/cass-operator/templates/deployment.yaml
jsanda: Yeah, we definitely want to include the labels. I think the approach I want to take is to define a common chart that each of our chart pulls in as a dependency. The common chart can define various helper templates, like for labels. See https://github.com/technosophos/common-chart/tree/master/common for an example. Bitnami takes this approach as well https://github.com/bitnami/charts/tree/master/bitnami.

charts/cass-operator/values.yaml
jsanda: 👍

charts/cass-operator/values.yaml
jsanda: Good question. This is part of the auto generated stuff. I think we need to consider what makes the most sense in terms of our operator deployments and not deployments in general.

charts/cass-operator/values.yaml
jsanda: We can modify it. Care to elaborate?

charts/k8ssandra/Chart.yaml
jsanda: At the moment, yes the charts need to have the same version. We obviously need to change that. And as we start to introduce more charts, I think changing it becomes a bigger priority.

charts/cass-operator/values.yaml
burmanm: It feels like a mess. Some at the this level are deployment related, some like this one feels out of place. Also they're not ordered in any way, so user would have to check all the parameters to find something that might be relevant.

jsanda: I will see if I can clean it up and organize it a bit better.

charts/cass-operator/values.yaml
jsanda: >I don't know why the choice between podAnnotation vs. annotations in general

It makes more sense if you look at the `deployment.yaml` template. The annotations are only added to the pod template, not the deployment itself.

charts/cass-operator/values.yaml
jsanda: >I'm struggling to think why some parts would have different securityContexts.

In terms of security best practices I think it make a lot more sense for each application to run owner its own distinct user. I don't think it is a good idea for example for the Cassandra user to be used for every container created and run in k8ssandra.

jsanda:
@burmanm can you do do another review on this? 

I don't feel like I fully addressed the concerns about `values.yaml`, but I feel like that is something we can iterate on in subsequent PRs as the chart refactoring continues. Once we get this merged, then I would like to focus on #176 and test out namespace-scope for cass-operator and the Prometheus bits.

charts/cass-operator/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,13 @@
+kind: ClusterRoleBinding
```
burmanm, line 1: Should this have the same: {{- if .Values.clusterScoped }} ? Or does it not matter if there's no ClusterRole for it?

burmanm: Is this extra file? There's also "rolebinding.yaml" that includes the same ClusterRole?

jsanda: This ClusterRole is needed all the time. I know it is needed for decommission support to  work with persistent volumes which are cluster scoped. See https://github.com/datastax/cass-operator/pull/242. 

I it also mentions nodes. I will have to see what kind of calls are made for nodes, and then try to rename this to something a bit more intuitive.

jsanda: I added a comment in `clusterrole.yaml` explaining why it is needed.

burmanm (approved):

PR #200, opened 2021-01-18
Minikube install instructions

Miles-Garnsey:
This PR to add Minikube installation instructions. Notes - 

* I have removed the service.type: NodePort setting. Ideally we'd probably like this to be a LoadBalancer rather than a NodePort (since we recommend `port-forward` to access the service anyway. But changing this to LoadBalancer (as described in the [chart doco](https://github.com/traefik/traefik-helm-chart/blob/master/traefik/values.yaml)) did not have the intended effect and the service was still created as NodePort.
* This is probably something to look at in both the Minikube and KinD deployment charts, because this setting is insecure (per the Traefik chart [docs](https://github.com/traefik/traefik-helm-chart/blob/ba98e364cf82d3c0d179b321aa668d53dc10e030/traefik/values.yaml#L222)). While this is fine for local development, it would be a risk in any cloud environment. I have included a warning about this in the Minikube deployment instructions.


jdonenine:
Hi @Miles-Garnsey thanks for the PR!  While we review, would you mind going through our CLA signing process and confirm that you've completed it? We'll need that done before we can marge.

https://cla.datastax.com/

docs/content/en/docs/topics/ingress/traefik/minikube-deployment/_index.md
```diff
@@ -0,0 +1,51 @@
...
+
+### 1. Install Minikube and start the cluster
+
+Installation instructions for Minikube are [here](https://minikube.sigs.k8s.io/docs/start/).
+
+Once installed, running `minikube start` will bring up a local cluster.
+
+### 2. Install Traefik via Helm
+
+Traefik can be installed via in conjunction with the below Traefik values file -
```
jdonenine, line 19: Small typo here, I think you're just missing the word "Helm" in `installed via Helm`

jdonenine (changes requested):
@Miles-Garnsey thanks for the updates and PR - a good addition, I think.

Your notes about production use of our example Traefik configurations is a good one, I've created #207 to update other relevant docs to reflect what you've included here.

If you can update the small typo I commented on, I will take care of the changes to address #207 in other places and we can get this merged.

Thanks!

jdonenine:
@Miles-Garnsey thanks for the updates and PR - a good addition, I think.

Your notes about production use of our example Traefik configurations is a good one, I've created #207 to update other relevant docs to reflect what you've included here.

jdonenine:
I should have mentioned that I locally tested deployment using minikube myself per your instructions, went as planned.

docs/content/en/docs/topics/ingress/traefik/minikube-deployment/_index.md
Miles-Garnsey: Thanks, you're right. I have updated.

Miles-Garnsey:
Thanks @jdonenine, I have softened the comments on cloud deployments slightly, to acknowledge that you'd also need to create security groups/firewall rules (for AWS, GCP respectively) in order to get yourself into real trouble. 

I did some more research to confirm my assumptions around that. I would still be cautious about deploying nodeport services to public clouds though, not something I'd be game to try.

jdonenine (approved):
Thanks for the updates, 👍

PR #208, opened 2021-01-20
Configure Cassandra default superuser

### LINKED ISSUE #159: Make default super user configurable
cass-operator generates a super user that is configurable. If authentication is enabled (see #151), we should make the super user configurable.

I should have the option to bring my own credentials which should be specified via a secret name. Alternatively, I should be able to just specify a username and have k8ssandra generate a password and the secret for me.

We might have properties something like this:

```yaml
# example values.yaml
k8ssandra:
  auth:
    enabled: true
    superuser:      
      secret: ""
      username: admin
```
If the `secret` property is set, we ignore `username`. If `username` is set we generate a random password and create the secret that will get specified in the cassdc.yaml template.

jsanda:
Fixes #159 

#179 needs to be merged first.


jsanda:
The following info is relevant for testing purposes. 

See the comments [here](https://github.com/jsanda/k8ssandra/blob/cassandra-default-superuser/charts/k8ssandra/values.yaml#L13) for an explanation of the default behavior as well as for the new superuser properties.

To obtain the credentials you can use the following:

```
$ kubectl get secret <secret name> -o json | jq -r '.data.username' | base64 --decode

$ kubectl get secret <secret name> -o json | jq -r '.data.password' | base64 --decode
```

I suggest then verifying those credentials can be used to access cqlsh one on the of C* pods as follows:

```
$ kubectl exec -it <cassandra-pod> -c cassandra -u <username> -p <password>
```

burmanm (approved):

PR #214, opened 2021-01-20
Add FAQ outlining the versions of the high-order components deployed …

jdonenine:
…in our system.

This should represent the current set of versions included - We'll need to update this prior to 1.0.0 when we bring in new versions of some items, like the management-api.

image


johnsmartco (approved):
Jeff, these edits look good to me. If the version numbers are valid for K8ssandra today, go ahead and merge. But if these edits look ahead to 1.0 (late Feb), wait. (?) Thanks.

jdonenine:
Thanks for the review @johnsmartco the versions are current, so I've gone ahead and merged.  They'll get updated as we add newer versions moving towards 1.0.0.

Issue #83, opened 2020-11-24
Add multi-DC support


jdonenine:
**Is your feature request related to a problem? Please describe.**
As a user, I want to be able to create a multi-DC cluster with the k8ssandra-cluster chart. 

**Describe the solution you'd like**
This would likely come in the form of a helm upgrade. As a user, I may create a cluster with one DC. Later I decided that I want to add a second DC.

**Additional context**
Legacy reference: [Jira](https://k8ssandra.atlassian.net/browse/K8C-71)

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-138) by [Unito](https://www.unito.io)

jsanda:
related to #109

jdonenine:
Closing as an old dup of newer ticket https://github.com/k8ssandra/k8ssandra/issues/524

Issue #54, opened 2020-11-19
Make racks configurable


jsanda:
**Is your feature request related to a problem? Please describe.**
Users should be able to configure the racks of a `CassandraDatacenter`. Here is an example manifest:

```yaml
apiVersion: cassandra.datastax.com/v1beta1
kind: CassandraDatacenter
metadata:
  name: k8ssandra-test
spec:
  clusterName: cluster1
  serverType: cassandra
  serverVersion: "3.11.7"
  managementApiAuth:
    insecure: {}
  size: 3
  storageConfig:
    cassandraDataVolumeClaimSpec:
      storageClassName: server-storage
      accessModes:
        - ReadWriteOnce
      resources:
        requests:
          storage: 5Gi
  racks:
    - name: rack1
      zone: us-central1-a
    - name: rack2
      zone: us-central1-b
    - name: rack3
      zone: us-central1-c
```

**Describe the solution you'd like**
The `racks` property of `CassandraDatacenter` should support templated values in the [cassdc.yaml](https://github.com/k8ssandra/k8ssandra/blob/main/charts/k8ssandra-cluster/templates/cassdc.yaml) template.

demoontz:
also storage: should be templated iguess

jsanda:
Hi @demoontz 

Yes storage should be templated as well. I would prefer to address that in a separate ticket though. I just have not gotten around to writing it up. 

Would you be interested in working on this?

jsanda:
related to #109

jsanda:
Note that the `zone` property has been deprecated in favor of `nodeAffinityLabels`.

jsanda:
Racks need to be set at the DC level. We are targeting multi-DC support post-1.0 (see #83). For this ticket we can assume a single DC.

PR #221, opened 2021-01-21
Refactoring to introduce cluster level configs

jsanda:
This PR moves configs out from under the `k8ssandra` property and puts them under `cassandra`. 

The `datacenterName` and `size` properties have been removed. There is a now a `datacenters` array. Each element should have `name` and `size` properties.

Note that this PR does not intend to introduce multi-DC support. We will address that in https://github.com/k8ssandra/k8ssandra/issues/83. In fact, in this PR you will see that I hard code access to `datacenters[0]`, so even if a user defines additional DCs, they are ignored. 

We want this refactoring in place though before we tackle https://github.com/k8ssandra/k8ssandra/issues/54 which we want to do for 1.0.


jeffbanks (approved):
LGTM

Issue #163, opened 2021-01-07
Traefik configuration support for Stargate services


jdonenine:
**Describe the solution you'd like**
Ingress configuration via Traefik should include what's necessary to allow external access to all of the stargate services (CQL, REST, Graph, Auth) when enabled.

**Describe alternatives you've considered**
Manually exposing via NodePort configuration on the services.

PR #236, opened 2021-01-23
#163: add ingress config for Stargate

jakerobb:
I haven't tested with k3d yet, but I've copied over the traefik config and updated the docs. In terms of kind, I think this is good to go. I'll address k3d over the weekend.


charts/k8ssandra/templates/traefik.ingressroutes.yaml
```diff
@@ -43,3 +45,92 @@ spec:
         - name: {{ $clusterName }}-{{ $datacenterName }}-service
           port: 9042
 {{- end }}
+{{- if $traefik.stargate.enabled }}
```
bradfordcp, line 48: Since Stargate supports native K8s ingress it doesn't necessarily have to live under `$traefik.stargate`. Traefik is only used here for TCP ingress routing. I'm open to nesting HTTP-based ingress under a separate key (IE `.Values.ingress.stargate.enabled`).

Separately we should consider what this looks like when Stargate ends up within a sub-chart.

bradfordcp (commented):
I don't see any blockers here, just some comments to consider.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
```diff
@@ -43,3 +45,92 @@ spec:
...
+{{- $stargateHost := default "localhost" $traefik.stargate.host }}
+apiVersion: networking.k8s.io/v1beta1
+kind: Ingress
+metadata:
+  name: k8ssandra-stargate
+  labels:
+{{ include "k8ssandra.labels" . | indent 4 }}
+spec:
+  rules:
+    - host: {{ $stargateHost }}
```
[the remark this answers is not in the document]
jakerobb: No, `$stargateHost` is assigned with a default value of `localhost` on line 53. The main `values.yaml` also specifies a default.

I could be convinced that each route should have its own overridable host value, falling back to the default if unspecified.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
```diff
@@ -43,3 +45,92 @@ spec:
...
+      http:
+        paths:
+          - path: /v1/auth
```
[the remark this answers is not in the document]
jakerobb: I looked and did not find any such support.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
jakerobb: Fair point. Maybe we should avoid implementation-specific keys in our config entirely? `ingress.traefik` would then only be for traefik-specific settings.

jakerobb:
I'd appreciate any insight people might have about the linter failure. I'm pretty sure it's because traefik.ingressroutes.yaml renders out with `---` on its first line, thus making an empty object with no name -- but that's not new with my changes.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
```diff
@@ -43,3 +45,92 @@ spec:
         - name: {{ $clusterName }}-{{ $datacenterName }}-service
           port: 9042
 {{- end }}
+{{- if $traefik.stargate.enabled }}
+---
+{{- if and $traefik.stargate.cassandra.enabled $traefik.cassandra.enabled }}
+  {{- fail "ingress.traefik.stargate.cassandra.enabled and ingress.traefik.cassandra.enabled cannot both be enabled." }}
+{{- end }}
+{{- $stargateHost := default "localhost" $traefik.stargate.host }}
```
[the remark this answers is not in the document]
jakerobb: It's in both; this one makes it not blow up if you explicitly set it to an empty string in your value overrides.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
```diff
@@ -43,3 +45,92 @@ spec:
...
+apiVersion: networking.k8s.io/v1beta1
+kind: Ingress
```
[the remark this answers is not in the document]
jakerobb: Doing so meant that I was forced to match on path, which will be a bit fragile as Stargate evolves (e.g. `v1`, `v2` are sure to change someday). One thing we could do is require distinct hostnames for each backend, but that would be annoying to default. Thoughts?

charts/k8ssandra/templates/traefik.ingressroutes.yaml
bradfordcp: Possibly, we could have a parameter for `ingressProvider` with a value of `traefik`, `nginx`, `kong`, etc. Then in the template match on that value.

FWIW I think we will most likely be moving the ingress configs under each sub-chart for example:

```yaml
kube-prometheus-stack
  grafana:
    ingress:
      enabled: false
      # Uncomment to listen for Grafana traffic under a route prefix
      # path: /grafana
```

so we may end up with something like

```yaml
stargate
  ingress:
    enabled: true
    tcp_ingress_provider: traefik
    host: stargate.example.com
```

What do you think?

bradfordcp:
The lint failure doesn't make any sense. It passes locally when I run the same logic locally.

```console
$ ls | while read c; do
>             if [[ -d $c ]]; then
>               helm lint $c
> 
>               if [[ $? -ne 0 ]]; then
>                 EXIT_CODE=$?
>                 echo "::error file=charts/$c::Error linting $c, check job output"
>                 exit $EXIT_CODE
>               fi
>             fi
>           done
==> Linting backup
[INFO] Chart.yaml: icon is recommended

1 chart(s) linted, 0 chart(s) failed
==> Linting cass-operator
[INFO] Chart.yaml: icon is recommended

1 chart(s) linted, 0 chart(s) failed
==> Linting k8ssandra
[INFO] Chart.yaml: icon is recommended

1 chart(s) linted, 0 chart(s) failed
==> Linting k8ssandra-common
[INFO] Chart.yaml: icon is recommended

1 chart(s) linted, 0 chart(s) failed
==> Linting restore
[INFO] Chart.yaml: icon is recommended

1 chart(s) linted, 0 chart(s) failed
```

charts/k8ssandra/templates/traefik.ingressroutes.yaml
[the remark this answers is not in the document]
bradfordcp: If I specifically set it to an empty string and the value of `localhost` is rendered it would be confusing. I expect failure in the case of invalid values files.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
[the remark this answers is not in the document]
bradfordcp: Looking at the Kubernetes [concept docs](https://kubernetes.io/docs/concepts/services-networking/ingress/) for ingress I see can see a number of realistic configurations.

1. Developer installation
    * Most likely a locally installed k8s
    * Very little control over DNS, maybe /etc/hosts access, probably not
 2. Multi-node / Off-host k8s cluster
     * Most likely from a cloud provider or externally manager
     * Ability to control DNS
 
With _option 1_ I feel like everything should be path-based. To that end, our default configuration for ingress (including monitoring) should do all routing based on path prefixes.

For _Option 2_ I could see a couple of configurations. If it's possible to influence DNS then that would be the preferred approach. We route to Stargate backends with hostname matching and don't need to bother with defining paths. We _could_ specify paths, but it doesn't really add much IMHO (this assumption may be wrong). Should DNS be locked down then we can still deploy a config just like _option 1_.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
```diff
@@ -43,3 +45,92 @@ spec:
         - name: {{ $clusterName }}-{{ $datacenterName }}-service
           port: 9042
 {{- end }}
+{{- if $traefik.stargate.enabled }}
+---
+{{- if and $traefik.stargate.cassandra.enabled $traefik.cassandra.enabled }}
+  {{- fail "ingress.traefik.stargate.cassandra.enabled and ingress.traefik.cassandra.enabled cannot both be enabled." }}
```
bradfordcp, line 51: The error message here isn't necessarily true. It's possible to enable both, they would just each need their own external port. To be clear users do not have to use 9042 for this traffic, it's just an adjustment client-side to the AddressTranslator used in the driver or SniProxy.

With that being said I struggle to understand a deployment where you would want to expose both the Stargate CQL service _and_ the main cluster's native protocol port. At this point I don't really have a recommendation, just want to comment on the situation and kick off a convo _IF_ we want to change something.

bradfordcp: @jsanda feel free to chime in here.

jsanda: Sorry, I have not gone through the PR yet :(

>With that being said I struggle to understand a deployment where you would want to expose both the Stargate CQL service and the main cluster's native protocol port. 

@adejanovski didn't you do some benchmarking with Stargate? I would like to hear your thoughts on this.

adejanovski: I agree with @bradfordcp, if you deploy a cluster with Stargate then you probably meant to use it as a coordination layer (whatever the protocol) and don't need to access the data nodes native port.
It would be good IMO to take an opinionated approach: either you deploy Stargate nodes and they are your only access points to the cluster or you don't and then have to go through the Cassandra nodes.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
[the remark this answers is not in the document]
jsanda: If we are defaulting to `localhost` here, then I think it makes sense to go ahead put that in `values.yaml`.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
jsanda: >FWIW I think we will most likely be moving the ingress configs under each sub-chart for example:

Yes, this is exactly what I have in mind. And we can put common code/templates into k8ssandra-common. 

For now, I wouldn't worry too much about the placement and organization of these properties since it will change when we create a separate Stargate chart.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
jsanda: I am in agreement with @bradfordcp and @adejanovski.

charts/k8ssandra/templates/stargate/service.yaml
```diff
@@ -1,8 +1,9 @@
 {{- if .Values.stargate.enabled }}
+{{- $datacenterName := (index .Values.cassandra.datacenters 0).name -}}
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ .Release.Name }}-k8ssandra-stargate-service
+  name: {{ .Release.Name }}-{{ $datacenterName }}-stargate-service
```
jsanda, line 6: When we create a separate chart for Stargate it will be easier (hopefully) to standarize on naming. This will just become:

```
name: {{ include "k8ssandra-common.fullname" . }}
```

charts/k8ssandra/templates/stargate/service.yaml
```diff
@@ -17,12 +18,15 @@ spec:
...
     - name: metrics
       port: 8085
       protocol: TCP
-    - name: cql
+    - name: cassandra
       protocol: TCP
       port: 9042
   selector:
-    app: {{ .Release.Name }}-k8ssandra-stargate
+    app: {{ .Release.Name }}-{{ $datacenterName }}-stargate
```
jsanda, line 31: Similar to the name this will hopefully become something like this:

```yaml
selector:
    matchLabels:
      {{- include "k8ssandra-common.selectorLabels" . | nindent 6 }}
```

charts/k8ssandra/templates/traefik.ingressroutes.yaml
```diff
@@ -43,3 +45,92 @@ spec:
...
+{{- if $traefik.stargate.rest.enabled }}
+    - host: {{ $stargateHost }}
+      http:
+        paths:
+          - path: /v2/
+            backend:
+              serviceName: {{ $clusterName }}-{{ $datacenterName }}-stargate-service
+              servicePort: 8082
+{{- end }}
+{{- if $traefik.stargate.health.enabled }}
```
jsanda, line 97: Do we need to expose `/check`? I am inclined to think no assuming it is primarily intended to be exposed internally for readiness/liveness probes.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
```diff
@@ -43,3 +45,92 @@ spec:
...
+    - host: {{ $stargateHost }}
+      http:
+        paths:
+          - path: /checker
+            backend:
+              serviceName: {{ $clusterName }}-{{ $datacenterName }}-stargate-service
+              servicePort: 8084
+{{- end }}
+{{- if $traefik.stargate.metrics.enabled }}
```
jsanda, line 106: Do we need to expose `/metrics`? I am inclined to think no as it is primarily used internally for collecting metrics.

jsanda (changes requested):
In addition to my other feedback it would great to see some unit tests.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
```diff
@@ -43,3 +45,92 @@ spec:
         - name: {{ $clusterName }}-{{ $datacenterName }}-service
```
jsanda, line 45: `$clusterName` needs to be replaced with `$releaseName`.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
```diff
@@ -43,3 +45,92 @@ spec:
...
+  labels:
+{{ include "k8ssandra.labels" . | indent 4 }}
+spec:
+  rules:
+    - host: {{ $stargateHost }}
+      http:
+        paths:
+          - path: /v1/auth
+            backend:
+              serviceName: {{ $clusterName }}-{{ $datacenterName }}-stargate-service
```
jsanda, line 67: `$clusterName` needs to be replaced with `$releaseName`.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
```diff
@@ -43,3 +45,92 @@ spec:
...
+              servicePort: 8081
+{{- if $traefik.stargate.graphql.enabled }}
+    - host: {{ $stargateHost }}
+      http:
+        paths:
+          - path: /graphql-schema
+            backend:
+              serviceName: {{ $clusterName }}-{{ $datacenterName }}-stargate-service
```
jsanda, line 75: `$clusterName` needs to be replaced with `$releaseName`.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
```diff
@@ -43,3 +45,92 @@ spec:
...
+              servicePort: 8080
+          - path: /graphql/
+            backend:
+              serviceName: {{ $clusterName }}-{{ $datacenterName }}-stargate-service
```
jsanda, line 79: `$clusterName` needs to be replaced with `$releaseName`.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
```diff
@@ -43,3 +45,92 @@ spec:
...
+              servicePort: 8080
+  {{- if $traefik.stargate.graphql.playground.enabled }}
+          - path: /playground
+            backend:
+              serviceName: {{ $clusterName }}-{{ $datacenterName }}-stargate-service
```
jsanda, line 84: `$clusterName` needs to be replaced with `$releaseName`.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
```diff
@@ -43,3 +45,92 @@ spec:
...
+              servicePort: 8080
+  {{- end }}
+{{- end }}
+{{- if $traefik.stargate.rest.enabled }}
+    - host: {{ $stargateHost }}
+      http:
+        paths:
+          - path: /v2/
+            backend:
+              serviceName: {{ $clusterName }}-{{ $datacenterName }}-stargate-service
```
jsanda, line 94: `$clusterName` needs to be replaced with `$releaseName`.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
```diff
@@ -43,3 +45,92 @@ spec:
...
+              servicePort: 8082
+{{- end }}
+{{- if $traefik.stargate.health.enabled }}
+    - host: {{ $stargateHost }}
+      http:
+        paths:
+          - path: /checker
+            backend:
+              serviceName: {{ $clusterName }}-{{ $datacenterName }}-stargate-service
```
jsanda, line 103: `$clusterName` needs to be replaced with `$releaseName`.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
```diff
@@ -43,3 +45,92 @@ spec:
...
+              servicePort: 8084
+{{- end }}
+{{- if $traefik.stargate.metrics.enabled }}
+    - host: {{ $stargateHost }}
+      http:
+        paths:
+          - path: /metrics
+            backend:
+              serviceName: {{ $clusterName }}-{{ $datacenterName }}-stargate-service
```
jsanda, line 112: `$clusterName` needs to be replaced with `$releaseName`.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
```diff
@@ -43,3 +45,92 @@ spec:
...
+{{ include "k8ssandra.labels" . | indent 4 }}
+spec:
+  entryPoints:
+  {{- range $traefik.stargate.cassandra.entrypoints }}
+  - {{ . }}
+  {{- end }}
+  routes:
+    - match: Host(`{{ $stargateHost }}`)
+      services:
+        - name: {{ $clusterName }}-{{ $datacenterName }}-stargate-service
```
jsanda, line 131: `$clusterName` needs to be replaced with `$releaseName`.

jsanda (changes requested):
When I fixed the service names, I was able to generate an auth token, access the document and REST apis. I was not able to access the graphql playground though.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
bradfordcp: Agreed, I would make sure this stays internal to the cluster. If you want external access `kubectl port-forward`

charts/k8ssandra/templates/traefik.ingressroutes.yaml
bradfordcp: Agreed, this should not be included in the ingress config.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
jakerobb: Thanks all -- my original thinking was:

1. I'm not aware of a reason to expose both at the same time
2. We don't currently have any precedent for customizing port numbers anywhere in k8ssandra.

I wrote the error message based on what is currently supported by the charts as written, not intending to imply that simultaneous support was technically impossible should we decide we wanted it.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
jakerobb: Okay. I had it in there for my own testing and settled on leaving it in with an off-by-default switch. Will remove.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
jakerobb: Oh, I've been using the same values for those in my testing. 🤦🏻‍♂️ Good catch, thanks.

charts/k8ssandra/templates/stargate/service.yaml
jsanda: https://helm.sh/docs/chart_template_guide/function_list/#nindent

It's relatively new to me. I am only familiar with a handful of functions. There are a lot 😄

charts/k8ssandra/templates/traefik.ingressroutes.yaml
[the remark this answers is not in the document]
jakerobb: I've changed values.yaml to localhost and replaced the `default` here with `required` so that it blows up if you override with empty.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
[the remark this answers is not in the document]
jakerobb: I learned a couple weeks ago from @jdonenine about [nip.io](https://nip.io/). It's a free, publicly available DNS system that reads the IP out of the hostname dynamically -- so you can do something like stargate.127.0.0.1.nip.io and it will resolve to 127.0.0.1. Extremely useful for devs who don't have /etc/hosts access (and even for those who do)!

charts/k8ssandra/templates/traefik.ingressroutes.yaml
jakerobb: Removed.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
jakerobb: Removed.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
jakerobb: I think we just agreed that it's fine as-is. Right? Anyone requesting changes here?

charts/k8ssandra/templates/traefik.ingressroutes.yaml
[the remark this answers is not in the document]
jakerobb: Done.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
jsanda: @jakerobb it sounds like we are all on the same page and the check enforces what we want. I think we are good to go here 👍

charts/k8ssandra/templates/traefik.ingressroutes.yaml
```diff
@@ -40,6 +40,77 @@ spec:
...
-        - name: {{ $clusterName }}-{{ $datacenterName }}-service
+        - name: {{ $releaseName }}-{{ $datacenterName }}-service
+          port: 9042
+{{- end }}
+{{- if $traefik.stargate.enabled }}
+---
+{{- if and $traefik.stargate.cassandra.enabled $traefik.cassandra.enabled }}
+  {{- fail "ingress.traefik.stargate.cassandra.enabled and ingress.traefik.cassandra.enabled cannot both be enabled." }}
+{{- end }}
+{{- $stargateHost := required $traefik.stargate.host }}
```
jsanda, line 51: The `required` function expects two arguments. This will cause a templating error.

jsanda (changes requested):
It would be good to see some unit tests as part of this PR as well.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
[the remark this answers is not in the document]
jsanda: After spending some time reviewing and letting things sink in, I like the idea of providing options to address both scenarios. 

For a local, dev setup, e.g., kind, path-based routing seems like a good default. It frees me from have to worry about hostnames and more complex ingress configurations.

For the second scenario where the user may have ability to control DNS, I like allowing the user to specify different hostnames for each service and using host+path routing. The user could choose to use the same or different hostnames across services.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
jakerobb: Okay, I'm going to leave this as-is and revise when we make a separate chart.

charts/k8ssandra/templates/traefik.ingressroutes.yaml
jakerobb: Yep, found and fixed. Thanks. :)

jakerobb:
I've bumped the version; this is ready for merge.

I'm creating followup issues for unit testing and to revisit the ingress routing once @bradfordcp has gotten back to us with a decision on hosts/ports/paths.

jsanda (approved):

PR #238, opened 2021-01-24
medusa-operator chart

### LINKED ISSUE #219: Create medusa-operator chart
We want to pull out the medusa-operator bits from the `k8ssandra` into their own separate chart. See https://github.com/k8ssandra/k8ssandra/pull/202 for an example how to go about doing this.

jsanda:
Fixes #219

#202 needs to be merged first.


burmanm:
Is this PR correct? Says medusa-operator chart, but only commits reaper-operator-chart and removes one role-binding yaml file?

burmanm (approved):

burmanm:
Chart versions probably need to be updated, but since we haven't published them lately it doesn't really matter.

PR #241, opened 2021-01-25
Remove obsolete cassandra.size property

jdonenine:
This property moved under the datacenters objects.


jdonenine:
@jsanda is it necessary to bump the chart versions for this?

jsanda (changes requested):
Can you please bump the chart version as well?

jsanda:
I think it is a common practice with chart development in general to change the version whenever there is a chart modification. We have for the most part been following that practice. With that said, we have not been pushing recent versions to the chart repo, so I don't think it is much of an issue either way in this case. 

I'll go ahead and merge. Thanks for fixing this 😄

jsanda (approved):

Issue #158, opened 2021-01-07
Make Cassandra heap size configurable


jsanda:
**Is your feature request related to a problem? Please describe.**
I should be able to configure the JVM heap size for Cassandra.

**Describe the solution you'd like**
There are a lot of heap properties that can be configured for the JVM. For this ticket, I want to focus on only the following three:

* -Xms - min heap size
* -Xmx - max heap size
* -Xmn - new generation size

I prefer to add support for other heap settings in separate tickets. I want to make sure sufficient investigation is devoted to each of the settings since heap and GC settings is a complex topic.

For a good overview on GC tuning and regions, see the [GC Tuning](https://thelastpickle.com/blog/2018/04/11/gc-tuning.html) post on the TLP blog.

The min and max sizes should be the same, so we can have a single property for them, e.g., `heapSize`.

For the new generation, Cassandra uses this formula to calculate a default value - `min(100 * num_cores, 1/4 * heap size)`. This is often times a bad default that can lead to a lot of promotions to the old generations. In terms of GC, promotions are expensive. In my experience I have seen that allocating 50% of the heap to the new generation, particularly in ready heavy workloads yields much better performance. I suggestion we go with half the total heap as a default for the new generation.

We might have properties that look something like this:

```yaml
# example values.yaml
k8ssandra:
  configuration:
    jvm:
      heapSize: 8GB
      newGenerationSize: 4GB
```
When we add multi-DC support (#83) we will want to make this configurable at the DC level as well.

jsanda:
My example in the description of how the properties might look is **not** good. We are going to want to have a top-level `cassandra` property that will have various properties. Auth is a good example. For auth we might have this:

```yaml
cassandra:
  auth:
    enabled: true
```
We want to be careful about throwing stuff under a `jvm` property because it could easily wind up being a dumping ground. If there were just a handful of JVM settings, that might be fine, but there are lots. There are are a lot for each of the garbage collectors. Then we have setting for heap memory. There are properties to configure GC logging. We want to provide some organization and be opinionated about it.

With all that said, I might lean towards something like this:

```yaml
cassandra:
  jvm:
     heapSize:
     heapNewGenerationSize: 
```

I realize it is not all that different from the initial example, but it is import to convey that we should spend time thinking about what should be exposed and how.

Miles-Garnsey:
I was thinking it would be nice to be able to pull these settings out of configMaps if people want to customise them. Would there be interest in that? 

[...]

PR #242, opened 2021-01-25
JVM options heap inclusion

jeffbanks:
Provides cluster, dc, and default jvm heap option configurations as described in #158.

[Design Notes](https://github.com/k8ssandra/k8ssandra/issues/158#issuecomment-765010492)

### DIFF
--- a/charts/k8ssandra/templates/cassandra/cassdc.yaml
+++ b/charts/k8ssandra/templates/cassandra/cassdc.yaml
@@ -43,7 +43,7 @@ spec:
 {{- else if .Values.cassandra.resources }}
   resources:
 {{ toYaml .Values.cassandra.resources | indent 6 }}
-{{- end}}
+{{- end }}
 {{- if .Values.cassandra.auth.superuser.secret }}
   superuserSecretName: {{ .Values.cassandra.auth.superuser.secret }}
 {{- else if .Values.cassandra.auth.superuser.username }}
@@ -66,9 +66,19 @@ spec:
       credentials_validity_in_ms: {{ .Values.cassandra.auth.cacheValidityPeriodMillis }}
       credentials_update_interval_in_ms: {{ .Values.cassandra.auth.cacheUpdateIntervalMillis }}
     jvm-options:
-      initial_heap_size: "800M"
-      max_heap_size: "800M"
-
+{{- if (index .Values.cassandra.datacenters 0).heap }}
+      initial_heap_size: {{ default "800M" (index .Values.cassandra.datacenters 0).heap.size }}
+      max_heap_size: {{ default "800M" (index .Values.cassandra.datacenters 0).heap.size }}
+      heap_size_young_generation: {{ default "1.6Gi" (index .Values.cassandra.datacenters 0).heap.newGenSize }}
+{{- else if .Values.cassandra.heap }}
+      initial_heap_size: {{ default "800M" .Values.cassandra.heap.size }}
+      max_heap_size: {{ default "800M" .Values.cassandra.heap.size }}
+      heap_size_young_generation: {{ default "1.6Gi" .Values.cassandra.heap.newGenSize}}
+{{- else }}
+      initial_heap_size: {{ "800M" }}
+      max_heap_size: {{ "800M" }}
+      heap_size_young_generation: {{ "1.6Gi" }}
+{{- end}}
       additional-jvm-opts:
 {{- if .Values.cassandra.auth.enabled }}
         - "-Dcassandra.system_distributed_replication_dc_names={{ (index .Values.cassandra.datacenters 0).name }}"
--- a/charts/k8ssandra/templates/cassandra/cassdc.yaml
+++ b/charts/k8ssandra/templates/cassandra/cassdc.yaml
@@ -67,13 +67,21 @@ spec:
       credentials_update_interval_in_ms: {{ .Values.cassandra.auth.cacheUpdateIntervalMillis }}
     jvm-options:
 {{- if (index .Values.cassandra.datacenters 0).heap }}
-      initial_heap_size: {{  (index .Values.cassandra.datacenters 0).heap.size }}
-      max_heap_size: {{  (index .Values.cassandra.datacenters 0).heap.size }}
-      heap_size_young_generation: {{  (index .Values.cassandra.datacenters 0).heap.newGenSize }}
-{{- else if .Values.cassandra.heap }}
-      initial_heap_size: {{  .Values.cassandra.heap.size }}
+  {{- if (index .Values.cassandra.datacenters 0).heap.size }}
+      initial_heap_size: {{ (index .Values.cassandra.datacenters 0).heap.size }}
+      max_heap_size: {{ (index .Values.cassandra.datacenters 0).heap.size}}
+  {{- end }}
+  {{- if (index .Values.cassandra.datacenters 0).heap.newGenSize }}
+      heap_size_young_generation: {{ (index .Values.cassandra.datacenters 0).heap.newGenSize }}
+  {{- end }}
+  {{- else if .Values.cassandra.heap }}
+  {{- if .Values.cassandra.heap.size  }}
+      initial_heap_size: {{ .Values.cassandra.heap.size }}
       max_heap_size: {{ .Values.cassandra.heap.size }}
-      heap_size_young_generation: {{  .Values.cassandra.heap.newGenSize}}
+  {{- end }}
+  {{- if  .Values.cassandra.heap.newGenSize }}
+      heap_size_young_generation: {{ .Values.cassandra.heap.newGenSize }}
+  {{- end }}
 {{- end}}
       additional-jvm-opts:
 {{- if .Values.cassandra.auth.enabled }}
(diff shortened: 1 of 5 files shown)


charts/k8ssandra/templates/cassandra/cassdc.yaml
```diff
@@ -66,9 +66,19 @@ spec:
       credentials_validity_in_ms: {{ .Values.cassandra.auth.cacheValidityPeriodMillis }}
       credentials_update_interval_in_ms: {{ .Values.cassandra.auth.cacheUpdateIntervalMillis }}
     jvm-options:
-      initial_heap_size: "800M"
-      max_heap_size: "800M"
-
+{{- if (index .Values.cassandra.datacenters 0).heap }}
+      initial_heap_size: {{ default "800M" (index .Values.cassandra.datacenters 0).heap.size }}
+      max_heap_size: {{ default "800M" (index .Values.cassandra.datacenters 0).heap.size }}
+      heap_size_young_generation: {{ default "1.6Gi" (index .Values.cassandra.datacenters 0).heap.newGenSize }}
```
jsanda, line 72: The young generation is only part of the overall heap, so it cannot be larger than the total heap. It looks like you are setting the young generation to double the total size. My suggestion was to set the young generation to half the total heap size.

charts/k8ssandra/templates/cassandra/cassdc.yaml
```diff
@@ -66,9 +66,19 @@ spec:
...
+{{- else if .Values.cassandra.heap }}
```
jsanda, line 73: If we define defaults in `k8ssandra/values.yaml` for `cassandra.heap.size` and for `cassandra.heap.newGenSize`, then lines 73 - 79 can be simplified to an `else` block that simply assigns those values. Note that even if the user does not set these properties, then the chart will inherit the values defined in `values.yaml`.

tests/unit/template_cassdc_test.go
```diff
@@ -97,6 +98,14 @@ var _ = Describe("Verify CassandraDatacenter template", func() {
 			// Server version and mgmt-api image specified
 			Expect(cassdc.Spec.ServerVersion).ToNot(BeEmpty())
 			Expect(cassdc.Spec.ServerImage).ToNot(BeEmpty())
+
+			// JVM heap options -- default to settings as defined in cassdc.yaml
+			var config Config
+			Expect(json.Unmarshal(cassdc.Spec.Config, &config)).To(Succeed())
+			Expect(config.JvmOptions).ToNot(BeNil())
+			Expect(config.JvmOptions.InitialHeapSize).To(Equal("800M"))
```
jsanda, line 106: As mentioned above, let's have defaults defined in `values.yaml` and the young generation has to be smaller than the total heap size.

tests/unit/template_cassdc_test.go
```diff
@@ -423,11 +431,83 @@ var _ = Describe("Verify CassandraDatacenter template", func() {
...
+			var config Config
+			Expect(json.Unmarshal(cassdc.Spec.Config, &config)).To(Succeed())
+			Expect(config.JvmOptions).ToNot(BeNil())
+			Expect(config.JvmOptions.InitialHeapSize).To(Equal("300M"))
+			Expect(config.JvmOptions.MaxHeapSize).To(Equal("300M"))
+			Expect(config.JvmOptions.YoungGenSize).To(Equal("600M"))
+		})
+
+		// Note: currently only one DC supported, to be expanded in future release.
+		It("setting JVM heap settings at datacenter WITHOUT cluster-level settings", func() {
```
jsanda, line 490: If we define cluster-level defaults in `values.yaml`, this test case may not be relevant.

jsanda (changes requested):

charts/k8ssandra/templates/cassandra/cassdc.yaml
burmanm: Do we really want to set these? The options we want to expose differ based on the selected GC.

jsanda: Probably not, for the same reasons we do not want to set container/pod resources. At deployment time, i.e., when `helm install` is run and the manifests are generated we don't know how much memory is available.

jsanda (changes requested):

charts/k8ssandra/templates/cassandra/cassdc.yaml
jeffbanks: Adding this block to `values.yaml` under `cassandra`.
```yaml
heap:
    size: 800M
    newGenSize: 400M
```

charts/k8ssandra/templates/cassandra/cassdc.yaml
jeffbanks: Yes, the young_generation is incorrect, it should have been the other way around.

Moving the defaults to values.yaml and out of cassdc.yaml.

tests/unit/template_cassdc_test.go
jeffbanks: 👍

charts/k8ssandra/templates/cassandra/cassdc.yaml
jsanda: @burmanm made a good point earlier in a separate review that we may not want to explicitly set these. I agree. I do think we should have commented out and documented though in `values.yaml` to make it more clear what config options are available. What do you think?

jeffbanks: I agree, it may be more reasonable to offer the customization options as opposed to explicitly providing/setting any values.

I will insert the comments into the values.yaml for the `cluster` and `dc` level settings available.

tests/unit/template_cassdc_test.go
jeffbanks: 👍

jeffbanks:
Updated values.yaml to include description of optional `dc-level` and `cluster-level` settings. 

- The `default` settings for min, max, and newGen sizes are no longer being provided.
- The recommended values for `min` and `max` are to be set to the same value.
- The recommended `newGen` size is half of the min and max.

charts/k8ssandra/templates/cassandra/cassdc.yaml
```diff
@@ -66,9 +66,15 @@ spec:
       credentials_validity_in_ms: {{ .Values.cassandra.auth.cacheValidityPeriodMillis }}
       credentials_update_interval_in_ms: {{ .Values.cassandra.auth.cacheUpdateIntervalMillis }}
     jvm-options:
-      initial_heap_size: "800M"
-      max_heap_size: "800M"
-
+{{- if (index .Values.cassandra.datacenters 0).heap }}
+      initial_heap_size: {{  (index .Values.cassandra.datacenters 0).heap.size }}
```
jsanda, line 70: You need to check that the DC `heap.size` is set.

charts/k8ssandra/templates/cassandra/cassdc.yaml
```diff
@@ -66,9 +66,15 @@ spec:
...
+      max_heap_size: {{  (index .Values.cassandra.datacenters 0).heap.size }}
+      heap_size_young_generation: {{  (index .Values.cassandra.datacenters 0).heap.newGenSize }}
```
jsanda, line 72: You need to check that the DC `heap.newGenSize` is set.

charts/k8ssandra/templates/cassandra/cassdc.yaml
```diff
@@ -66,9 +66,15 @@ spec:
...
+{{- else if .Values.cassandra.heap }}
+      initial_heap_size: {{  .Values.cassandra.heap.size }}
```
jsanda, line 74: You need to check that the cluster-level `heap.size` is set.

charts/k8ssandra/templates/cassandra/cassdc.yaml
```diff
@@ -66,9 +66,15 @@ spec:
...
+      max_heap_size: {{ .Values.cassandra.heap.size }}
+      heap_size_young_generation: {{  .Values.cassandra.heap.newGenSize}}
```
jsanda, line 76: You need to check that the cluster-level `heap.newGenSize` is set.

jsanda (changes requested):

charts/k8ssandra/templates/cassandra/cassdc.yaml
jeffbanks: Adjusted in update; only setting when specified.

charts/k8ssandra/templates/cassandra/cassdc.yaml
jeffbanks: Adjusted in update; only setting when specified.

charts/k8ssandra/templates/cassandra/cassdc.yaml
jeffbanks: Adjusted in update; only setting when specified.

charts/k8ssandra/templates/cassandra/cassdc.yaml
jeffbanks: Adjusted in update; only setting when specified.

jsanda (approved):
@jeffbanks nice work!

PR #243, opened 2021-01-25
Return error code if removal failed

burmanm:
@jsanda This is for testing the docker image process


jsanda:
Thanks!

jsanda (approved):

jsanda:
@burmanm can you merge and check that the image is built and pushed? If it is, can you then open another PR to update the cleaner image in the templates?

Issue #226, opened 2021-01-22
Backups do not work with Cassandra 3.11.8 or 3.11.9


jsanda:
## Bug Report

**Describe the bug**
I ran into this issue while trying to test https://github.com/k8ssandra/k8ssandra/issues/193. Creating a backup fails with a validation error in medusa-operator like this:

```
2021-01-22T04:08:30.225Z	ERROR	controllers.CassandraBackup	failed to patch status with CassdcTemplateSpec	{"CassandraDatacenter": "default/dc1", "error": "CassandraBackup.cassandra.k8ssandra.io \"backup-1\" is invalid: status.cassdcTemplateSpec.spec.serverVersion: Unsupported value: \"3.11.9\": supported values: \"6.8.0\", \"6.8.1\", \"6.8.2\", \"6.8.3\", \"3.11.6\", \"3.11.7\", \"4.0.0\""}
github.com/go-logr/zapr.(*zapLogger).Error
	/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128
github.com/k8ssandra/medusa-operator/controllers.(*CassandraBackupReconciler).Reconcile
	/workspace/controllers/cassandrabackup_controller.go:108
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.2/pkg/internal/controller/controller.go:235
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.2/pkg/internal/controller/controller.go:209
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.2/pkg/internal/controller/controller.go:188
k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1
	/go/pkg/mod/k8s.io/apimachinery@v0.18.6/pkg/util/wait/wait.go:155
k8s.io/apimachinery/pkg/util/wait.BackoffUntil
	/go/pkg/mod/k8s.io/apimachinery@v0.18.6/pkg/util/wait/wait.go:156
k8s.io/apimachinery/pkg/util/wait.JitterUntil
[...]

PR #255, opened 2021-01-26
Upgrade medusa-operator

### LINKED ISSUE #229: Upgrade medusa-operator and restore template
This is a reminder for follow up changes that will be needed after https://github.com/k8ssandra/k8ssandra/issues/192 is merged. Two things are needed:

* Update medusa-operator image
* Update the `restore/templates/restore.yaml` template

jsanda:
Fixes #229 and #226

This PR has the following changes:

* Updates medusa-operator image
* Update CassandraBackup CRD
* Update CassandraRestore CRD
* Updates restore template with latest changes in medusa-operator

### DIFF
--- a/charts/k8ssandra/crds/medusa/backup.yaml
+++ b/charts/k8ssandra/crds/medusa/backup.yaml
@@ -60,6 +60,10 @@ spec:
                     canaryUpgrade:
                       description: Indicates that configuration and container image changes should only be pushed to the first rack of the datacenter
                       type: boolean
+                    canaryUpgradeCount:
+                      description: The number of nodes that will be updated when CanaryUpgrade is true. Note that the value is either 0 or greater than the rack size, then all nodes in the rack will get updated.
+                      format: int32
+                      type: integer
                     clusterName:
                       description: The name by which CQL clients and instances will know the cluster. If the same cluster name is shared by multiple Datacenters in the same Kubernetes namespace, they will join together in a multi-datacenter cluster.
                       minLength: 2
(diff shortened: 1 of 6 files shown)


charts/k8ssandra/crds/medusa/backup.yaml
```diff
@@ -3817,14 +3722,7 @@ spec:
...
                       description: Version string for config builder, used to generate Cassandra server configuration
-                      enum:
-                        - 6.8.0
-                        - 6.8.1
-                        - 6.8.2
-                        - 6.8.3
-                        - 3.11.6
-                        - 3.11.7
-                        - 4.0.0
+                      pattern: (6\.8\.\d+)|(3\.11\.\d+)|(4\.0\.\d+)
```
emerkle826, line 3725: This is more of a question, but will `4.0-beta4` match `4.0`? The pattern will only handle digits it seems, so I guess I'm not sure where the version here maps to an image tag.

emerkle826 (approved):
This looks ok to me.

charts/k8ssandra/crds/medusa/backup.yaml
jsanda: That is a great question. It looks like the answer is no. We will have to open an issue in cass-operator. I can do that.

emerkle826: will do

emerkle826: done: https://github.com/datastax/cass-operator/issues/373

PR #257, opened 2021-01-27
need to make sure the enabled flag for kubernetes is set

jsanda:
With the Medusa image upgrade, there is a new config setting that we need to make sure is set. It basically tells Medusa it is running in k8s which will in turn cause Medusa to use Jolokia (and later the management-api) instead of nodetool.


jsanda:
cc @jdonenine

burmanm (approved):

PR #264, opened 2021-01-28
add auth support for medusa

### LINKED ISSUE #154: Generate a Cassandra user for Medusa
If authentication is enabled (see #151), we should create a Cassandra user for Medusa.

There are a few things that need to be done:

* Generate C* credentials
* Create a secret that stores the credentials
* Specify the Medusa user and secret in the cassdc.yaml template
* Include credentials in medusa.ini
  * medusa.ini is generated from a configmap. We may need to get the credentials out of the secret and into that configmap.

jsanda:
Fixes #154


jsanda:
The following info is relevant for testing purposes.

See the comments [here](https://github.com/jsanda/k8ssandra/blob/medusa-auth/charts/k8ssandra/values.yaml#L98) of the new properties.

To obtain the credentials you can use the following:

```
$ kubectl get secret <secret name> -o json | jq -r '.data.username' | base64 --decode

$ kubectl get secret <secret name> -o json | jq -r '.data.password' | base64 --decode
```

I suggest then verifying those credentials can be used to access cqlsh one on the of C* pods as follows:

```
$ kubectl exec -it <cassandra-pod> -c cassandra -u <username> -p <password>
```

Also note that to actually verify that Medusa is configured correctly to use the credentials, you will need to run a backup. When the gRPC service starts up, Medusa does not create a connection to Cassandra.

tests/unit/template_cassdc_test.go
```diff
@@ -581,3 +753,44 @@ var _ = Describe("Verify CassandraDatacenter template", func() {
 		})
 	})
 })
+
```
jeffbanks, line 756: I like the addition of the helper functions.

jeffbanks (approved):
Comments added. Overall, looks great.

PR #268, opened 2021-01-29
remove obsolete medusa-operator templates

jsanda:
This PR removes the medusa-operator templates from the k8ssandra chart since medusa-operator is in its own chart now.


bradfordcp (approved):

bradfordcp:
LGTM

Issue #249, opened 2021-01-26
Stargate container memory calculated incorrectly


jsanda:
## Bug Report

**Describe the bug**
While testing #163 I observed that my Stargate container was repeatedly getting OOM killed. The problem is due to the container memory being set to the same value as the JVM heap size. The container memory is calculated as follows:

```
{{ mul 1.5 (.Values.stargate.heapMB | default 1024) }}Mi
```
The problem is that the math functions operate on int64 values (see [here](https://helm.sh/docs/chart_template_guide/function_list/#math-functions) in the Helm docs), so the 1.5 gets cast to an int64 which winds up as 1.

Another problem is that the `stargate.heapMB` property is not documented in `values.yaml` 

**To Reproduce**
Deploy k8ssandra with the following values:

```yaml
stargate:
  enabled: true
  heapMB: 2048
```

**Expected behavior**
The total container memory must be greater than the JVM heap size. We also need to document the `heapMB` property in `values.yaml`.

**Environment (please complete the following information):**

* Helm charts version info
0.31.0

jakerobb:
I'm not sure how to explain this, but I cannot reproduce this on my machine. The container is up and running fine on my machine and has not been OOM killed. I configured stargate.heapMB=2048, and this excerpt of `kubectl describe pod` shows that the container has a heap min and max of 2048M and a resource limit of 2Gi.

```
Containers:
  cluster1-k8ssandra-dc1-stargate:
    Container ID:   containerd://41d79305478af96b38a8717e64681e16fc627606a3a5b592b22bc19df045932e
    Image:          stargateio/stargate-3_11:v1.0.0
    Image ID:       docker.io/stargateio/stargate-3_11@sha256:810036d9e0018e151c8ec152f72e8d2bcf55a2980cdc5f1125f035b212aa6b12
    Ports:          8080/TCP, 8081/TCP, 8082/TCP, 8084/TCP, 8085/TCP, 8090/TCP, 9042/TCP, 8609/TCP, 7000/TCP, 7001/TCP
    Host Ports:     0/TCP, 0/TCP, 0/TCP, 0/TCP, 0/TCP, 0/TCP, 0/TCP, 0/TCP, 0/TCP, 0/TCP
    State:          Running
      Started:      Tue, 26 Jan 2021 22:12:10 -0500
    Ready:          True
    Restart Count:  0
    Limits:
      cpu:     1
      memory:  2Gi
    Requests:
      cpu:      200m
[...]

PR #282, opened 2021-01-30
#165: adding initial Grafana dashboard for Stargate

jakerobb:
**What this PR does**:
Creates an initial Grafana dashboard for Stargate, as shown below. It's not much, but it's functional and useful. There's certainly room to iterate on it more, but I needed at least this much in order to make progress on #249.

Note that the changes from #281 are also needed in order for there to be any data behind this dashboard.

image

**Which issue(s) this PR fixes**:
Fixes #165

**Checklist**
- [x] Changes manually tested
- [ ] Chart versions updated (if necessary)
- [ ] Automated Tests added/updated
- [ ] Documentation added/updated
- [x] CLA Signed:  [DataStax CLA](https://cla.datastax.com/)

Issue #258, opened 2021-01-27
What conventions should we adopt for setting default values for resources?


jsanda:
**Type of question**
We have (or had) multiple tickets that involve setting defaults for things like container memory and heap size. There has been scattered discussions on the topic, so I thought it would be good to try and capture that in a single place and try to figure out what conventions we want to adopt.

For reference here are some of the related tickets:

* #242 
* #160
* #249
* #173 

The discussion primarily focuses on Cassandra and Stargate, but it does apply to other application we deploy. Here are several questions that we need to consider:

* Should we specify defaults for memory, JVM heap, cpu?
* If we do not specify defaults, what settings will the application use?
  * In some cases, the application defaults are not good. Cassandra's default for the heap new generation size for example is typically bad.
  * If we specify defaults, what _type_ of defaults should we use? 
    * Should they be minimum requirements?
    * Should they target a dev environment? 
* Is it ok if we specify default settings in some places and not others? Would it be confusing for example if we specify default heap settings for Cassandra and not for Stargate (or vice versa)?

As we are quickly approaching a 1.0 release, I am not necessarily suggesting we have all this figured out in the 1.0 time frame. I do want to start the discussion now though as it is coming up multiple places.

I also want to note that there is a desire to be opinionated in k8ssandra. That is a way we can provide value add with the project.

jdonenine:
I would definitely support the idea that we should be opinionated and provide best-practice-based defaults where we can.

jakerobb:
In my opinion:
* We should specify defaults where the underlying thing we're configuring has unreasonable or inconsistent defaults by itself.
    * Cassandra's defaults, as John mentioned above, are what I'm thinking of when I say "unreasonable"
    * The JVM's behavior when heap size is not specified is what I'm thinking of when I say "inconsistent".
* We should include defaults for any setting required to get up and running.
* Our defaults should target a developer running k8ssandra on a local machine using kind/k3d/minikube.
    * One C*, one Stargate, ingress enabled, smallest reasonable memory footprint, etc.

My rationale for this is that one of the biggest obstacles to adoption of both Kubernetes and Cassandra is friction. Getting and environment configured and running _at all_, let alone _well_, is surprisingly hard. The more we do to streamline that experience, the more we are setting k8ssandra up to succeed. *I would argue that streamlining the experience of running C* on K8s is k8ssandra's prime directive and sole reason for existence.*

[...]

Issue #147, opened 2021-01-05
Remove the stargate.clusterVersion chart property


jsanda:
## Bug Report

**Describe the bug**
The k8ssandra-cluster chart has a `stargate.clusterVersion` property. It is used to determine which stargate image to use. The Stargate version should be determined by the Cassandra version. We currently only support Cassandra 3.11.7. I have created #144 to add support for 3.11.8 and 3.11.9. We will need to introduce a Cassandra version property for that ticket. That Cassandra version should ultimately determine which Stargate version is used.

We also have a `stargate.containerImage` property. When set to a value of `auto` the image version is determined by `stargate.clusterVersion`. I do like being able to override the image used. This can be useful for development/testing or advanced use cases that may require use of a custom image. I would like to see the following changes for it:

* Rename `containerImage to `image`.
* The `image` defaults to an empty string. In the default case the image version is determined by the Cassandra cluster version.
* If `image` is set, it overrides whatever default value would otherwise be used.
* Include comments to indicate that `image` should not normally be set. It might be done for development/test or other advanced use cases that require a custom image. 

Let's also add/update unit tests with the PR.

jsanda:
I am giving this a complexity of medium in large part because we do not yet have any automated tests around Stargate integration.

jsanda:
For this fix let's also make the image and imagePullPolicy configurable.

PR #283, opened 2021-01-30
#249: stargate bugfix and tuning

jakerobb:
**What this PR does**:
Resolving mathematical error in the Stargate pod's memory specs; adjusting memory defaults and also liveness/readiness probes for improved startup reliability.

Partially addresses #249 -- the mathematical error is gone. I have been unable to reproduce the issue John sees on his side; Stargate runs quite well for me with a 160MB heap and 320MB memory limit. I also added documentation for the Stargate tuning settings, which was called out in that issue.

**Checklist**
- [x] Changes manually tested


charts/k8ssandra/dashboards/stargate.json
```diff
@@ -0,0 +1,463 @@
+{
```
jakerobb, line 1: I'm not sure how or why this commit got included on this PR; I committed it on a different branch as part of #282.

charts/k8ssandra/templates/grafana/configmap.yaml
```diff
@@ -13,4 +13,8 @@ data:
 {{ .Files.Get "dashboards/overview.json" | indent 4 }}
```
jakerobb, line 13: This file is also from a commit on a different branch.

charts/k8ssandra/templates/stargate/stargate.yaml
```diff
@@ -77,16 +77,16 @@ spec:
           resources:
             limits:
               cpu: {{ .Values.stargate.cpuLimMillicores | default 1000 }}m
-              memory: {{ mul 1.5 (.Values.stargate.heapMB | default 1024) }}Mi
+              memory: {{ mul 4 (.Values.stargate.heapMB | default 256) }}Mi
```
bradfordcp, line 80: Can we move the default into `values.yaml`?

This removes the need for the commented out documentation line (or even a placeholder which is the style used in the `prometheusSpec` block). My preference is:

```yaml
  prometheus:
    # Provisions an instance of Prometheus as part of this release
    enabled: true
    prometheusSpec:
      # Uncomment to listen for Prometheus traffic under a route prefix
      # externalUrl: http://localhost:9090/prometheus
      # routePrefix: /prometheus
```

but it may make sense to align behind the format used in https://github.com/norwoodj/helm-docs

charts/k8ssandra/values.yaml
```diff
@@ -87,6 +87,9 @@ stargate:
   replicas: 1
   clusterVersion: "3.11" # this is the only supported value for now
   containerImage: "auto" # chooses image based on clusterVersion
+  # heapMB sets the heap size Stargate will use in megabytes. Default 256. Memory request and limit for the pod will be set to this value x2 and x4, respectively.
```
bradfordcp, line 90: See comment on `charts/k8ssandra/templates/stargate/stargate.yaml`

bradfordcp (changes requested):

charts/k8ssandra/templates/stargate/stargate.yaml
jsanda: This dovetails into #258.

In general though I do agree that if we are setting defaults we should set them in value.yaml. In some cases it may not be possible though, like if the default is a calculated value.

charts/k8ssandra/templates/stargate/stargate.yaml
```diff
@@ -77,16 +77,16 @@ spec:
...
             requests:
               cpu: {{ .Values.stargate.cpuReqMillicores | default 200 }}m
-              memory: {{ mul 1.5 (.Values.stargate.heapMB | default 1024) }}Mi
+              memory: {{ mul 2 (.Values.stargate.heapMB | default 256) }}Mi
```
jsanda, line 83: Can we put this in a variable to eliminate the duplication?

charts/k8ssandra/templates/stargate/stargate.yaml
```diff
@@ -77,16 +77,16 @@ spec:
...
           env:
             - name: JAVA_OPTS
               value: >-
                 -XX:+CrashOnOutOfMemoryError
-                -Xms{{ .Values.stargate.heapMB | default 1024 }}M
-                -Xmx{{ .Values.stargate.heapMB | default 1024 }}M
+                -Xms{{ .Values.stargate.heapMB | default 256 }}m
+                -Xmx{{ .Values.stargate.heapMB | default 256 }}m
```
jsanda, line 89: Can we put this in a variable to eliminate the duplication?

charts/k8ssandra/templates/stargate/stargate.yaml
```diff
@@ -77,16 +77,16 @@ spec:
           resources:
```
jsanda, line 77: I think it makes sense to make `resources` in general configurable. See [here](https://github.com/k8ssandra/k8ssandra/blob/main/charts/k8ssandra/templates/cassandra/cassdc.yaml#L53) for an example. 

Given that we have some insights into memory configuration I would be inclined to do as follows:

* If no values are set the apply defaults as you currently do
* else if heap/memory is specified, use those overrides
* else if `resources` is specified, use that.

The user directly setting `resources` would be a more advanced scenario IMO but one that we definitely want to support.

charts/k8ssandra/templates/stargate/stargate.yaml
jsanda: I have been following https://helm.sh/docs/chart_best_practices/values/ as a guide for how to document properties in `values.yaml`. 

What is not entirely clear to me is how best to document properties are commented out by default. Here is what we have in values.yaml for Stargate:

```yaml
stargate:
  # configuration regarding the Stargate API
  enabled: false
  replicas: 1
  clusterVersion: "3.11" # this is the only supported value for now
  containerImage: "auto" # chooses image based on clusterVersion
  # heapMB sets the heap size Stargate will use in megabytes. Default 256. Memory request and limit for the pod will be set to this value x2 and x4, respectively.
  # cpuReqMillicores sets the CPU request for the Stargate pod in millicores. default 200.
  # cpuLimMillicores sets the CPU limit for the Stargate pod in millicores. default 1000.
```
I think it could be easy to miss that those are properties. I think I would prefer to see this:

```yaml
stargate:
  # configuration regarding the Stargate API
  enabled: false
  replicas: 1
  clusterVersion: "3.11" # this is the only supported value for now
  containerImage: "auto" # chooses image based on clusterVersion
  # heapMB sets the heap size Stargate will use in megabytes. Default 256. Memory request and limit for the pod will be set to this value x2 and x4, respectively.
  # heapMB:
  
  # cpuReqMillicores sets the CPU request for the Stargate pod in millicores. default 200.
  # cpuReqMillicores:

  # cpuLimMillicores sets the CPU limit for the Stargate pod in millicores. default 1000.
  # cpuLimMillicores:
```
I realize that this is likely to highly subjective :)

jsanda (changes requested):
@jakerobb overall the changes look good!

Can you include a fix for #147 in this PR as well?

I tested your branch and Stargate was stable with the default settings. I did not however make any client requests? Did you test with any load?

charts/k8ssandra/templates/stargate/stargate.yaml
jakerobb: Makes sense to me, John. Chris's example includes reasonable values (possibly matching the default?) on the commented out lines. I think I like that better than leaving it blank. Thoughts?

jakerobb:
> Can you include a fix for #147 in this PR as well?

Will do.

> I tested your branch and Stargate was stable with the default settings. I did not however make any client requests? Did you test with any load?

Yes, I threw one million requests at two Stargate nodes over the course of three hours. The screenshot of the dashboard on #282 shows a shorter duration of a similar test.

jsanda:
>Yes, I threw one million requests at two Stargate nodes over the course of three hours. The screenshot of the dashboard on #282 shows a shorter duration (I think it was 100k requests) of a similar test running against an even smaller heap.

Awesome!

charts/k8ssandra/templates/stargate/stargate.yaml
jsanda: If we have default values that we are going to use, then yes definitely set them in `values.yaml`.

jsanda (changes requested):
Looks like we still need to address #147. I still see the `clusterVersion` and `containerImage` properties.

When you replace the `containerImage` property with `image` also be sure to add an `imagePullPolicy` property as well.

jakerobb:
Okay, I have:
1. Rebased on main
2. Moved defaults from template to values.yaml
3. documented options in values.yaml
4. removed stargate.clusterVersion property
5. changed behavior of stargate.image as described on #147
6. added stargate.imagePullPolicy
7. created a `utils` package and moved some utility functions there
8. added several unit tests for Stargate
9. updated our test suite to use unique, descriptive values for namespace, clusterName, and releaseName.

Re: #7 -- I initially planned to create a single "utils.go" file, but instead decided to separate them by subject (EnvVars, Containers, ...). This minimizes the chances for any one file to get big and difficult to manage, reduces namespace pollution when importing, and feels like a straightforward way to divide them up.

Re: #8 -- more tests are coming in separate PRs.

Re: #9 -- I'll be adding a test that, paired with this change, would have caught #293.

Tests are passing; I think this is good to merge. @bradfordcp and @jsanda, here's a poke because you guys requested changes.

tests/unit/utils/containers.go
```diff
@@ -0,0 +1,48 @@
+package utils
+
+import (
+	cassdcV1beta1 "github.com/datastax/cass-operator/operator/pkg/apis/cassandra/v1beta1"
+	. "github.com/onsi/gomega"
+	coreV1 "k8s.io/api/core/v1"
+)
+
+func GetInitContainer(cassdc *cassdcV1beta1.CassandraDatacenter, name string) *coreV1.Container {
```
jsanda, line 9: Can you change the name to `GetCassdcInitContainer`? You cannot have overloaded functions in Go :(

tests/unit/utils/containers.go
```diff
@@ -0,0 +1,48 @@
...
+	return GetContainerByName(cassdc.Spec.PodTemplateSpec.Spec.InitContainers, name)
+
+}
+
+func GetContainer(cassdc *cassdcV1beta1.CassandraDatacenter, name string) *coreV1.Container {
```
jsanda, line 14: Can you change the function name to `GetCassdcContainer`?

tests/unit/utils/containers.go
```diff
@@ -0,0 +1,48 @@
...
+	return GetContainerByName(cassdc.Spec.PodTemplateSpec.Spec.Containers, name)
+
+}
+
+func AssertInitContainerNamesMatch(cassdc *cassdcV1beta1.CassandraDatacenter, names ...string) {
```
jsanda, line 19: I was just thinking about my previous two comment. Maybe it makes more sense to put CassandraDatacenter-specific function into their own package, e.g., `cassdc`. Then we can just write `cassdc.GetContainer`. vs `GetCassdcContainer`? 

wdty?

Note that I am using `cassdc` because it is the abbreviation specified in the CassandraDatacenter CRD. It allows to me to do `kubectl get cassdc` for example.

tests/unit/utils/envVars.go
```diff
@@ -0,0 +1,14 @@
+package utils
+
+import (
+	coreV1 "k8s.io/api/core/v1"
+)
+
+func FindEnvVarByName(haystack []coreV1.EnvVar, needle string) *coreV1.EnvVar {
```
jsanda, line 7: While this made me laugh, it might be confusing to other users in the future. Maybe change these to something like `envVars` and `name` or `envVarName`.

jsanda (changes requested):

charts/k8ssandra/templates/stargate/stargate.yaml
```diff
@@ -8,7 +10,7 @@ metadata:
     app: {{ .Release.Name }}-{{ include "k8ssandra.datacenterName" . }}-stargate
 {{ include "k8ssandra.labels" . | indent 4 }}
 spec:
-  replicas: {{ default 1 .Values.stargate.replicas }}
+  replicas: {{ required "stargate.replicas is required" .Values.stargate.replicas }}
```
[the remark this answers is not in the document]
jakerobb: I thought it made sense to keep it in order to handle gracefully the scenario where the user overrides the value to empty (presumably inadvertently). Ditto for the other `required`s I added in this file.

tests/unit/utils/envVars.go
jakerobb: Heh, that was a leftover from when I was trying to write it generically. 

FWIW, I've used this nomenclature in the past (in Java), and never had it confuse anyone. It's both whimsical and very obvious. :)

Nonetheless, since it's not generic in any way, it makes sense to change it here.

tests/unit/utils/containers.go
jakerobb: I like the idea of having the functions in a separate package. It improves the focus of the package, shortens function names without losing clarity, and in places where you only need one, you can dot-alias the imports and not qualify them at all.

I made the change locally to move the cassdc functions to a separate package, but then found that `cassdc` is a very common (and reasonable) variable name, and using `cassdc` as the import alias creates a conflict. The conflict doesn't cause a compile failure; the variables just win. I conclude that `cassdc` is not a good choice for the alias. I changed it to use a dot alias in template_cassdc_test, and I think I like the result even more. :)

tests/unit/template_stargate_test.go
```diff
@@ -0,0 +1,186 @@
+package unit_test
+
+import (
+	"github.com/gruntwork-io/terratest/modules/helm"
+	. "github.com/k8ssandra/k8ssandra/tests/unit/utils"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	appsV1 "k8s.io/api/apps/v1"
```
jsanda, line 8: minor nit - can you make the alias all lower case. That seems to be the common convention.

tests/unit/utils/containers.go
```diff
@@ -0,0 +1,48 @@
+package utils
+
+import (
+	cassdcV1beta1 "github.com/datastax/cass-operator/operator/pkg/apis/cassandra/v1beta1"
```
jsanda, line 4: minor nit - can you make the aliasas all lower case. That seems to be the common convention.

tests/unit/utils/envVars.go
```diff
@@ -0,0 +1,14 @@
+package utils
+
+import (
+	coreV1 "k8s.io/api/core/v1"
```
jsanda, line 4: minor nit - can you make the alias all lower case. That seems to be the common convention.

jsanda, line 4: minor nit - can you make the alias all lower case. That seems to be the common convention.

tests/unit/utils/envVars.go
```diff
@@ -0,0 +1,14 @@
+package utils
```
jsanda, line 1: Can you rename the file to either `envvars.go` or `env_vars.go` to be consistent with naming conventions.

jsanda (changes requested):
Can you also bump the k8ssandra chart version.

tests/unit/utils/cassdc/containers.go
```diff
@@ -0,0 +1,32 @@
+package cassdc
+
```
jsanda, line 2: I like the reorganization! Just so you are aware a lot of these functions will likely be needed in reaper-operator and in medusa-operator. At some point soon we can pull the common test util code out into its own project.

Since I initially created these functions, I can add a follow up PR with some docs after yours is merged.

tests/unit/utils/kubeapi/ingress.go
```diff
@@ -0,0 +1,42 @@
+package kubeapi
+
+import (
+	. "fmt"
+	. "github.com/onsi/gomega"
+	networkingV1 "k8s.io/api/networking/v1beta1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+func FindIngressRuleByHttpPath(rules []networkingV1.IngressRule, path string) (*networkingV1.IngressRule, *networkingV1.HTTPIngressPath) {
```
jsanda: Can you add some comments for this?

tests/unit/utils/kubeapi/ingress.go
```diff
@@ -0,0 +1,42 @@
...
+		for _, pathCandidate := range ruleCandidate.HTTP.Paths {
+			if pathCandidate.Path == path {
+				return &ruleCandidate, &pathCandidate
+			}
+		}
+	}
+	return nil, nil
+}
+
+func VerifyNoRuleWithPath(rules []networkingV1.IngressRule, path string) {
```
jsanda: Can you add some comments for this?

tests/unit/utils/kubeapi/ingress.go
```diff
@@ -0,0 +1,42 @@
...
+	authRule, _ := FindIngressRuleByHttpPath(rules, path)
+	ExpectWithOffset(1, authRule).To(BeNil())
+}
+
+func VerifyIngressRule(rules []networkingV1.IngressRule, releaseName, path string, pathType *networkingV1.PathType, host string, port int) {
```
jsanda: Can you add some comments for this?

tests/unit/utils/traefik/httpRoutes.go
```diff
@@ -0,0 +1,39 @@
+package traefik
+
+import (
+	"fmt"
+	"github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	traefik "github.com/traefik/traefik/v2/pkg/provider/kubernetes/crd/traefik/v1alpha1"
+)
+
+func VerifyTraefikHTTPIngressRoute(ingress traefik.IngressRoute, entrypoint string, match string, backendService string, backendPort int) {
```
jsanda, line 10: Can you add some comments for this?

tests/unit/utils/traefik/httpRoutes.go
```diff
@@ -0,0 +1,39 @@
...
+	ExpectWithOffset(1, ingress.Spec.EntryPoints).To(ContainElement(entrypoint))
+	route := findTraefikHTTPRouteByMatch(&ingress.Spec, match)
+	ExpectWithOffset(1, route).ToNot(BeNil(), fmt.Sprintf("No Route found with match=%s for entrypoint=%s", match, entrypoint))
+	service := findTraefikHTTPServiceByName(route, backendService)
+	ExpectWithOffset(1, route).ToNot(BeNil(), fmt.Sprintf("No BackendService found with name=%s for match=%s and entrypoint=%s", backendService, match, entrypoint))
+	ExpectWithOffset(1, service.Port).To(Equal(int32(backendPort)))
+}
+
+func findTraefikHTTPRouteByMatch(spec *traefik.IngressRouteSpec, match string) *traefik.Route {
```
jsanda, line 19: Can you add some comments for this?

tests/unit/utils/traefik/tcpRoutes.go
```diff
@@ -0,0 +1,39 @@
+package traefik
+
+import (
+	"fmt"
+	"github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	traefik "github.com/traefik/traefik/v2/pkg/provider/kubernetes/crd/traefik/v1alpha1"
+)
+
+func VerifyTraefikTCPIngressRoute(ingress traefik.IngressRouteTCP, entrypoint string, match string, backendService string, backendPort int) {
```
jsanda, line 10: Can you add some comments for this?

jsanda (changes requested):
I made an initial pass through the first few commits. I will do another review later this afternoon.

jakerobb:
I've rebased this branch with the latest from main. Merging was fun! I've also added the requested doc comments, removed some debugging output, and taken care of a couple bits of fallout from the rebase.

jakerobb:
Rebasing has added dozens of unrelated commits to this PR. I think I did something wrong.

I am going to create a new branch, cherry pick the commits, and create a new PR. Leaving this open for now so that we don't lose track of the feedback and conversations in flight.

jakerobb:
This is now superseded by PR #333, which contains mostly the same changes but without all of the unrelated commits.

PR #288, opened 2021-01-30
Integration of helm-docs with our charts

### LINKED ISSUE #287: Helm chart reference documentation shipped as part of the chart and on k8ssandra.io
Users must currently either find and open the `values.yaml` files, use `helm show values ...` or look to k8ssandra.io. While the docs in our values.yaml files are pretty verbose the surfacing up of this content leaves something to be desired.

1. Tooling to automatically generate updated `README.md` files inside of each chart directory
2. Generated reference documentation on k8ssandra.io. This has already been started for some of the charts with symlinks and file includes. Let's get this stuff generated automatically and get rid of the symlinks.

[helm-docs](https://github.com/norwoodj/helm-docs) seems like a good starting point

**Additional context**
N/A

bradfordcp:
**What this PR does**:
* Visits `values.yaml` files for all charts and updating comments to that expected by `helm-docs`
* Updates `Chart.yaml` for all charts adding descriptions, website links, maintainer information, and repo urls
* Adds `scripts/generate-helm-docs.sh`
  * Iterates over all Helm charts generating a README.md
  * Creates a _generated.md file in `docs/content/en/docs/reference/CHART_NAME/`
* Updated _index.md for each Docs reference page to look for `_generated.md`
* Regenerated all docs given the new shell script

Fixes #287 

**Checklist**
- [x] Changes manually tested


scripts/generate-helm-docs.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# Generate docs for each chart
+ls charts | while read c; do
+  if [[ -d charts/$c ]]; then
+    helm-docs -c charts/$c -s file
+
+    mkdir -p docs/content/en/docs/reference/$c
+    helm-docs -c charts/$c -s file -t ../../docs/content/en/docs/reference/_generated.md.gotmpl -o ../../docs/content/en/docs/reference/$c/_generated.md
+  fi
+done
```
jakerobb, line 13: A few things:
* Did you mean to leave `set -x` in there? That'll be pretty chatty. I moved it to a subshell inside the loop so that the script only outputs the actual invocations of `helm-docs` and not all of the other junk.
* Using `ls` to iterate over files in a directory is not generally advisable (e.g. it'll fail if a directory name has a space); better to use [globbing](https://tldp.org/LDP/abs/html/globbingref.html).
* You're assuming the script is being invoked from the repo base directory. I've modified it such that the only assumption is that the script lives in a directory which shares a common parent with `charts/`, as is the case with `scripts/`.
* You're assuming `helm-docs` is installed. If it's not, the script will blow up very angrily. A pre-check seems advisable.

```suggestion
if ! command -v helm-docs &> /dev/null
then
    echo "helm-docs could not be found. Please ensure it is installed and on your PATH."
    echo "https://github.com/norwoodj/helm-docs"
    exit 1
fi

cd "$(dirname "$0")/.."

# Generate docs for each chart
for directory in charts/*; do
  if [[ -d "$directory" ]]; then
    chartName="$(basename ${directory})"
    mkdir -p "docs/content/en/docs/reference/${chartName}"
    (
      set -x
      helm-docs -c "${directory}" -s file
      helm-docs -c "${directory}" -s file -t ../../docs/content/en/docs/reference/_generated.md.gotmpl -o "../../docs/content/en/docs/reference/${chartName}/_generated.md"
    )
  fi
done
```

I didn't test the above; please verify locally.

What are your intentions with this script? Are we to execute it manually at release time? Or whenever we update any of the various values.yamls? Or can we bake in some automation somewhere?

jeffbanks (approved):
Nice work! It will be great to have this.  LGTM

jakerobb (commented):
Looks good! A couple small questions.

scripts/generate-helm-docs.sh
bradfordcp: Great question. IMHO (looking for feedback here) if you are planning a PR with a change to `Chart.yaml` or `values.yaml` you should run this script. Our CI will run this script and if there are any changes (IE you didn't run the script and should have) it will fail the PR. We _could_ make this step run automatically and push the changes back into the branch. The downside here is that if you open a Draft PR and this runs pushing a commit. Then you spend some cycles working through a couple more commits locally. You then have divergence and have to reconcile before pushing again. A counter-argument here would be that you need to have `helm-docs` installed or leverage the Docker image that the project provides. 

I think this leaves us in an interesting spot. Do we require the docs get regenerated and pushed as part of a PR? Should we generate these automatically and commit the result on branches?

Personally, I think the onus should be on the developer to include it as part of a PR. I could see a middle of the road approach where we have automation to generate the docs, but it has to be manually triggered (IE if it's the last thing blocking a PR). Thoughts?

bradfordcp: Separate from the questions above thanks for the script update. The other shell scripts we have offer a bit more polish than the previous iteration. It worked locally without issue.

jsanda: If we have to run the script manually, I will probably lead the pack in forgetting to run it. I am 👍  for automation and 👎  running it on the branch. If developers want to run it on a branch, I think that they should be able to do so. I think the automation should be run on main.

jakerobb: I agree with John -- automation on main only, and no requirements for PRs.

If I'm not mistaken, running it locally and pushing those changes as part of a PR will increase the odds of merge conflicts with others doing the same, which will be an annoying, pointless chore since we'd know the automation would trample the result post-merge. So, as a general habit, I'd advise not including the generated docs as part of PRs.

jakerobb (approved):

jsanda (approved):

PR #296, opened 2021-02-02
Fix trademark violation

michaelsembwever:
Projects and domain names can not use "Cassandra".
It is "Reaper" or "Reaper for Apache Cassandra".

https://www.apache.org/foundation/marks/faq/#products
https://apache.org/foundation/marks/#domains


jsanda (approved):

jdonenine:
So then how do we update the entire Reaper website?

image

PR #311, opened 2021-02-04
add support for C* 3.11.10 and make it the default version

jsanda:
**What this PR does**:
Adds support for Cassandra 3.11.10

**Which issue(s) this PR fixes**:
Fixes #308

PR #333, opened 2021-02-07
#249: stargate tuning and tests

### LINKED ISSUE #147: Remove the stargate.clusterVersion chart property

### LINKED ISSUE #249: Stargate container memory calculated incorrectly

jakerobb:
**What this PR does**:
I botched the branch for PR #283 in an attempt to rebase on main. I created a new branch, cherry picked as many of the commits as I could reasonably find, and remade some changes manually.

I also found that PR #320 had been trampled by PR #311; I've remade that change here.

Partially addresses #249 -- the mathematical error is gone. I have been unable to reproduce the issue John sees on his side; Stargate runs quite well for me with a 160MB heap and 320MB memory limit. I also added documentation for the Stargate tuning settings, which was called out in that issue.

Fixes #147

**Checklist**
- [x] Changes manually tested


tests/unit/template_stargate_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+			javaOpts := FindEnvVarByName(container.Env, "JAVA_OPTS")
+			Expect(javaOpts.Value).To(ContainSubstring("-Xms256M"))
+			Expect(javaOpts.Value).To(ContainSubstring("-Xmx256M"))
+
+			clusterName := FindEnvVarByName(container.Env, "CLUSTER_NAME")
+			Expect(clusterName.Value).To(Equal("k8ssandra-cluster"))
+
+			seed := FindEnvVarByName(container.Env, "SEED")
+			Expect(seed.Value).To(Equal("k8ssandra-cluster-seed-service.k8ssandra-namespace.svc.cluster.local"))
```
jsanda: another hard coded string

tests/unit/utils/kubeapi/ingress.go
```diff
@@ -0,0 +1,42 @@
+package kubeapi
+
+import (
+	. "fmt"
+	. "github.com/onsi/gomega"
+	networkingV1 "k8s.io/api/networking/v1beta1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+func FindIngressRuleByHttpPath(rules []networkingV1.IngressRule, path string) (*networkingV1.IngressRule, *networkingV1.HTTPIngressPath) {
```
jsanda: Can you add comments for this?

tests/unit/utils/kubeapi/ingress.go
```diff
@@ -0,0 +1,42 @@
...
+		for _, pathCandidate := range ruleCandidate.HTTP.Paths {
+			if pathCandidate.Path == path {
+				return &ruleCandidate, &pathCandidate
+			}
+		}
+	}
+	return nil, nil
+}
+
+func VerifyNoRuleWithPath(rules []networkingV1.IngressRule, path string) {
```
jsanda: Can you add comments for this?

tests/unit/utils/kubeapi/ingress.go
```diff
@@ -0,0 +1,42 @@
...
+	authRule, _ := FindIngressRuleByHttpPath(rules, path)
+	ExpectWithOffset(1, authRule).To(BeNil())
+}
+
+func VerifyIngressRule(rules []networkingV1.IngressRule, releaseName, path string, pathType *networkingV1.PathType, host string, port int) {
```
jsanda: Can you add comments for this?

tests/unit/utils/traefik/httpRoutes.go
```diff
@@ -0,0 +1,39 @@
+package traefik
+
+import (
+	"fmt"
+	"github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	traefik "github.com/traefik/traefik/v2/pkg/provider/kubernetes/crd/traefik/v1alpha1"
+)
+
+func VerifyTraefikHTTPIngressRoute(ingress traefik.IngressRoute, entrypoint string, match string, backendService string, backendPort int) {
```
jsanda, line 10: Can you add comments for this?

tests/unit/utils/traefik/tcpRoutes.go
```diff
@@ -0,0 +1,39 @@
+package traefik
+
+import (
+	"fmt"
+	"github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	traefik "github.com/traefik/traefik/v2/pkg/provider/kubernetes/crd/traefik/v1alpha1"
+)
+
+func VerifyTraefikTCPIngressRoute(ingress traefik.IngressRouteTCP, entrypoint string, match string, backendService string, backendPort int) {
```
jsanda, line 10: Can you add comments for this?

jsanda (changes requested):

tests/unit/utils/kubeapi/ingress.go
```diff
@@ -22,25 +22,30 @@ func FindIngressRuleByHttpPath(rules []networkingv1.IngressRule, path string) (*
 // VerifyNoRuleWithPath asserts that none of the given IngressRules have a given HTTP path entry.
 func VerifyNoRuleWithPath(rules []networkingv1.IngressRule, path string) {
 	authRule, _ := FindIngressRuleByHttpPath(rules, path)
-	ExpectWithOffset(1, authRule).To(BeNil())
```
jsanda: How come you changed from `ExpectWithOffset` to `Expect`? The former reports the line number at which `VerifyNoRuleWithPath` is called which I think is pretty helpful.

tests/unit/utils/kubeapi/ingress.go
```diff
@@ -22,25 +22,30 @@ func FindIngressRuleByHttpPath(rules []networkingv1.IngressRule, path string) (*
...
 // it exists and has the correct pathType, host, serviceName, and port.
-func VerifyIngressRule(rules []networkingv1.IngressRule, path string, pathType *networkingv1.PathType, host, serviceName string, port int) {
+func VerifyIngressRule(rules []networkingv1.IngressRule, path string, pathType *networkingv1.PathType, host *string, serviceName string, port int) {
 	rule, httpPath := FindIngressRuleByHttpPath(rules, path)
 
 	description := Sprintf("rule not found for path %v", path)
-	ExpectWithOffset(1, rule).ToNot(BeNil(), description)
-	ExpectWithOffset(1, httpPath).ToNot(BeNil(), description)
+	Expect(rule).ToNot(BeNil(), description)
+	Expect(httpPath).ToNot(BeNil(), description)
```
jsanda, line 35: Same as above - How come you changed from `ExpectWithOffset` to `Expect`?

tests/unit/utils/kubeapi/ingress.go
```diff
@@ -22,25 +22,30 @@ func FindIngressRuleByHttpPath(rules []networkingv1.IngressRule, path string) (*
...
 	if pathType == nil {
-		ExpectWithOffset(1, httpPath.PathType).To(BeNil(), description)
+		Expect(httpPath.PathType).To(BeNil(), description)
 	} else {
-		ExpectWithOffset(1, httpPath.PathType).To(Equal(pathType), description)
+		Expect(httpPath.PathType).To(Equal(pathType), description)
 	}
-	ExpectWithOffset(1, httpPath.Backend.ServiceName).To(Equal(serviceName), description)
-	ExpectWithOffset(1, httpPath.Backend.ServicePort).To(Equal(intstr.FromInt(port)), description)
+	Expect(httpPath.Backend.ServiceName).To(Equal(serviceName), description)
```
jsanda, line 49: Same as above - How come you changed from `ExpectWithOffset` to `Expect`?

jsanda (changes requested):

tests/unit/template_stargate_test.go
```diff
@@ -0,0 +1,185 @@
...
+			Expect(renderTemplate(options)).To(Succeed())
+			Expect(deployment.Kind).To(Equal("Deployment"))
+
+			templateSpec := deployment.Spec.Template.Spec
+			Expect(len(templateSpec.InitContainers)).To(Equal(1))
+			initContainer := templateSpec.InitContainers[0]
+			Expect(string(initContainer.ImagePullPolicy)).To(Equal("IfNotPresent"))
+
+			Expect(initContainer.Args[0]).To(Equal("-c"))
+			Expect(initContainer.Args[1]).To(ContainSubstring("nslookup k8ssandra-cluster-seed-service.k8ssandra-namespace.svc.cluster.local;"))
```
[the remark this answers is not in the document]
jakerobb: The intent is to confirm that it's looking for the expected hostname. I can change to use variables.

tests/unit/utils/kubeapi/ingress.go
jakerobb: During debugging, I found that `...WithOffset(1,...` in this specific case obscured useful information. The line number you're referring to is also in the output and easily identified.

tests/unit/utils/kubeapi/ingress.go
jakerobb: Weird, I did. Guess they got lost in the cherry picking. I'll dig them up and restore.

charts/k8ssandra/templates/stargate/stargate.yaml
```diff
@@ -78,16 +78,16 @@ spec:
...
               cpu: {{ .Values.stargate.cpuReqMillicores | default 200 }}m
-              memory: {{ mul 1.5 (.Values.stargate.heapMB | default 1024) }}Mi
+              memory: {{ mul 2 (.Values.stargate.heapMB | default 256) }}Mi
           env:
             - name: JAVA_OPTS
               value: >-
                 -XX:+CrashOnOutOfMemoryError
-                -Xms{{ .Values.stargate.heapMB | default 1024 }}M
-                -Xmx{{ .Values.stargate.heapMB | default 1024 }}M
+                -Xms{{ .Values.stargate.heapMB | default 256 }}m
```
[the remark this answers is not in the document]
jakerobb: Oh, it is. Another change that got lost in the cherry picking. :/

tests/unit/tests_suite_test.go
```diff
@@ -16,10 +16,10 @@ const (
 	CassOperatorChartsPath      = "../../charts/cass-operator"
 	MedusaOperatorChartsPath    = "../../charts/medusa-operator"
 	ReaperOperatorChartsPath    = "../../charts/reaper-operator"
-	DefaultTestNamespace        = "k8ssandra"
+	DefaultTestNamespace        = "k8ssandra-namespace"
```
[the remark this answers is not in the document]
jakerobb: Good idea. The use of "namespace", "release", and "cluster" in the corresponding variables is intended to make the nature of various failures obvious (e.g. when we inadvertently use releaseName instead of clusterName) and make it obvious when we use the wrong one.

tests/unit/utils/kubeapi/ingress.go
jsanda: >The line number you're referring to is also in the output and easily identified.

Maybe in this particular case but not always. For example, it won't be easily identifiable if I call the verify function multiple times in the test. 

To illustrate I modified the `VerifyNoRuleWithPath` function to fail. Here is what is the output looks like:

```
• Failure [0.317 seconds]
Verify Stargate ingress template
/Users/jsanda/Development/k8ssandra/k8ssandra/tests/unit/template_stargate_ingress_test.go:14
  by rendering it with options
  /Users/jsanda/Development/k8ssandra/k8ssandra/tests/unit/template_stargate_ingress_test.go:95
    using only default options [It]
    /Users/jsanda/Development/k8ssandra/k8ssandra/tests/unit/template_stargate_ingress_test.go:96

    Expected
        <*v1beta1.IngressRule | 0x0>: nil
    not to be nil

    /Users/jsanda/Development/k8ssandra/k8ssandra/tests/unit/utils/kubeapi/ingress.go:26
```
Line 96 is the declaration of the test spec. It does not tell me the exact line where the failure is. 

Next I modified `VerifyNoRuleWithPath` to fail and use `ExpectWithOffset(1, ...)`. Here is the output:

```
• Failure [0.313 seconds]
Verify Stargate ingress template
/Users/jsanda/Development/k8ssandra/k8ssandra/tests/unit/template_stargate_ingress_test.go:14
  by rendering it with options
  /Users/jsanda/Development/k8ssandra/k8ssandra/tests/unit/template_stargate_ingress_test.go:95
    using only default options [It]
    /Users/jsanda/Development/k8ssandra/k8ssandra/tests/unit/template_stargate_ingress_test.go:96

    Expected
        <*v1beta1.IngressRule | 0x0>: nil
    not to be nil

    /Users/jsanda/Development/k8ssandra/k8ssandra/tests/unit/template_stargate_ingress_test.go:178
```
Line 178 points to the call made from within `verifyIngressRules`.

But if I change it to use `ExpectWithOffset(2, ...)` my output is what I want:

```
• Failure [0.323 seconds]
Verify Stargate ingress template
/Users/jsanda/Development/k8ssandra/k8ssandra/tests/unit/template_stargate_ingress_test.go:14
  by rendering it with options
  /Users/jsanda/Development/k8ssandra/k8ssandra/tests/unit/template_stargate_ingress_test.go:95
    using only default options [It]
    /Users/jsanda/Development/k8ssandra/k8ssandra/tests/unit/template_stargate_ingress_test.go:96

    Expected
        <*v1beta1.IngressRule | 0x0>: nil
    not to be nil

    /Users/jsanda/Development/k8ssandra/k8ssandra/tests/unit/template_stargate_ingress_test.go:107
```
Now I know precisely at what line in the spec where the failure occurred. I think this is better but it certainly isn't perfect. We could refactor the the verify function to take an offset argument but as code gets refactored that offset could easily wind up pointing to the wrong place in the call stack.

I think the best solution and probably most idiomatic Go solution is to have the verify functions return an `error`. This will make it trivial to know where in the test spec the failure happens and good, descriptive error message should let us know what exactly failed.

jsanda: @jakerobb Let's leave the functions as is. I don't want to hold up the PR on this. I will create a follow up ticket where we can look at this more closely and decide what to do. It can be done post-merge.

tests/unit/template_stargate_test.go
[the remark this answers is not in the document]
jsanda: If we use a random value or suffix as I suggested above, won't we need to use variables here?

jdonenine:
@jakerobb this isn't change in this PR but I noticed when I was testing that the name for the ingress resource deployed for stargate is hard coded to `k8ssandra-ingress`

https://github.com/k8ssandra/k8ssandra/pull/333/files#diff-558b00f6344013f47fc9634cf5fb8546ad6c8a78e6f6f34f22c086ba56d57399R16

Should we swap that to include the release name like we typically do?

Figured we could work that change in here quickly since you're already touching the files - if you don't want to include I here that's ok as well, we can add an issue and cover it when we swapt the defaults.

jsanda:
> Should we swap that to include the release name like we typically do?

yes

charts/k8ssandra/templates/stargate/ingress.yaml
```diff
@@ -78,39 +40,25 @@ spec:
               servicePort: 8080
   {{- if $traefik.stargate.graphql.playground.enabled }}
           - path: /playground
+            pathType: Exact
```
jsanda, line 43: Out of curiosity why did you use `pathType: Exact` here?

charts/k8ssandra/templates/stargate/stargate.yaml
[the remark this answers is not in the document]
jakerobb: Actually, you were looking one commit at a time, I think. In the final commit, this is already done.

tests/unit/utils/kubeapi/ingress.go
jakerobb: Here again, I think you were just looking at non-latest commits. The comments are present on the latest commit.

tests/unit/utils/kubeapi/ingress.go
jakerobb: Okay. Note that, in later revisions, I added a mismatch description to the Expect calls. This provides sufficient context to identify the line in the calling function.

JeremiahDJordan:
> Stargate runs quite well for me with a 160MB heap and 320MB memory limit.

FYI the Stargate team has been doing a lot of testing and tuning work in the last few weeks around lowering the memory requirements for it.  Many of the things they are tweaking are DSE backend specific, but you might catch up with Michael Penick on his work there.

charts/k8ssandra/templates/stargate/ingress.yaml
jakerobb: My goal here was to minimize the footprint that this route makes in the overall ingress landscape, to leave as much room as possible for other software to coexist with k8ssandra. It's a hedge against the fact that native ingress routes are not port-sensitive.

It's a small thing, and if you wanted to argue that I should remove it, I'd not put up a fight.

jsanda (approved):

charts/k8ssandra/templates/stargate/ingress.yaml
bradfordcp: IIRC it's a single page app, there should be no routing beyond `/playground`. With `pathType: Exact` a request to `/playground/foo` will _not_ be routed to Stargate.

With that being said I have not personally tested the routes Stargate uses for the playground section, thus I don't know if its usage is appropriate here.

PR #340, opened 2021-02-08
337 heap settings fix

### LINKED ISSUE #337: Heap settings are not applied to Cassandra
## Bug Report

**Describe the bug**
Heap settings are not set correctly for Cassandra despite being customized in the charts.

**To Reproduce**
1. Run `helm template test charts/k8ssandra --set cassandra.heap.size=500`.
2. Check the config by looking at the output of `kubectl get cassdc dc1 -o yaml`.
3. Heap size isn't explicitely set there although it should be.

**Expected behavior**
Cassandra nodes should be customized with the heap settings from the charts.
It looks like leaving empty heap settings at the rack level causes the issue.
The following gives the buggy behavior:

```
  heap:
   size: 500M
   newGenSize: 200M
  resources: {}
  datacenters:
  - 
    name: dc1
    size: 3
    racks:
    - 
      name: rack1
      affinityLabels: {}
    heap:
      size: 
      newGenSize:
```

but this works as expected:

```
  size: 500M
   newGenSize: 200M
  resources: {}
  datacenters:
  - 
    name: dc1
    size: 3
    racks:
    - 
      name: rack1
      affinityLabels: {}
```

* Helm charts version info

```
NAME	NAMESPACE	REVISION	UPDATED                                	STATUS  	CHART           	APP VERSION
test	default  	1       	2021-02-08 14:35:23.241510784 +0000 UTC	deployed	k8ssandra-0.44.0	3.11.7
```

```
USER-SUPPLIED VALUES:
backupRestore:
  medusa:
    bucketName: k8ssandra-medusa-dev
    bucketSecret: medusa-bucket-key
    cassandraUser:
      secret: cassandra
      username: cassandra
    enabled: true
    image:
      pullPolicy: IfNotPresent
      repository: docker.io/k8ssandra/medusa
      tag: 6ab6a55541e9
    multiTenant: false
    storage: s3
cass-operator:
  enabled: true
cassandra:
  allowMultipleNodesPerWorker: false
  auth:
    enabled: false
    superuser:
      secret: cassandra
      username: cassandra
  cassandraLibDirVolume:
    size: 5Gi
    storageClass: standard
  clusterName: k8ssandra
  datacenters:
  - name: dc1
    size: 3
  heap:
    newGenSize: 200M
    size: 500M
  resources: {}
  version: 3.11.10
  versionImageMap:
    3.11.10: datastax/cassandra-mgmtapi-3_11_10:v0.1.19
cleaner:
  image: k8ssandra/k8ssandra-cleaner:618b8ff9d368
kube-prometheus-stack:
  alertmanager:
    enabled: false
    serviceMonitor:
      selfMonitor: false
  coreDns:
    enabled: false
  enabled: true
  grafana:
    adminPassword: secret
    adminUser: admin
    defaultDashboardsEnabled: false
    enabled: true
    grafana.ini: {}
    ingress:
      enabled: false
      path: null
    plugins:
    - grafana-polystat-panel
    serviceMonitor:
      selfMonitor: false
  kubeApiServer:
    enabled: false
  kubeControllerManager:
    enabled: false
  kubeDns:
    enabled: false
  kubeEtcd:
    enabled: false
  kubeProxy:
    enabled: false
  kubeScheduler:
    enabled: false
  kubeStateMetrics:
    enabled: false
  kubelet:
    enabled: false
  nodeExporter:
    enabled: false
  prometheus:
    enabled: true
    ingress:
      enabled: false
      paths: []
    prometheusSpec:
      externalUrl: ""
      routePrefix: /
    serviceMonitor:
      selfMonitor: false
  prometheusOperator:
    enabled: true
    namespaces:
      additional: []
      releaseNamespace: true
    serviceMonitor:
      selfMonitor: false
monitoring:
  grafana:
    provision_dashboards: true
  prometheus:
reaper-operator:
  enabled: true
repair:
  reaper:
    autoschedule: false
    cassandraUser:
      secret: cassandra
      username: cassandra
    enabled: true
    image:
      repository: docker.io/thelastpickle/cassandra-reaper
      tag: 2.1.3
    jmx:
      password: ""
      username: ""
stargate:
  cassandraUser:
    secret: ""
    username: ""
  clusterVersion: "3.11"
  containerImage: auto
  enabled: false
  replicas: 1
```

```
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.2", GitCommit:"faecb196815e248d3ecfb03c680a4507229c2a56", GitTreeState:"clean", BuildDate:"2021-01-13T13:28:09Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.11", GitCommit:"ea5f00d93211b7c80247bf607cfa422ad6fb5347", GitTreeState:"clean", BuildDate:"2020-09-14T07:50:38Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

* Kubernetes cluster kind:
kind

jeffbanks:
**What this PR does**:
Resolves issue reported found with unexpected empty entries in values.yaml of k8ssandra chart for heap settings at cluster and dc levels.  This was found to be a result of documentation additions in a recent PR.

Fixes #337 

**Checklist**
- [x] Changes manually tested


charts/k8ssandra/values.yaml
```diff
@@ -333,7 +333,7 @@ monitoring:
     provision_dashboards: true
 
   prometheus:
-    # -- Enabes the creation of Prometheus Operator ServiceMonitor custom
+    # -- Enables the creation of Prometheus Operator ServiceMonitor custom
```
bradfordcp, line 336: Nice catch!

charts/k8ssandra/values.yaml
```diff
@@ -74,9 +74,9 @@ cassandra:
   allowMultipleNodesPerWorker: false
 
   # -- Cluster-level heap configuration
-  heap:
-   size:
-   newGenSize:
+  #heap:
```
bradfordcp, line 77: Interesting, I thought we checked to see if the value was set and calculated if not. It could be we check to see if it is defined, not empty. In either case, I like the inclusion of the keys here as it drives the reference docs (see the regenerated files from the GH action). If we don't want to leverage empty keys here we should update the inline docs to let the user know which keys are available (again, reference the previously generated docs).

charts/k8ssandra/values.yaml
```diff
@@ -122,9 +122,9 @@ cassandra:
 
     # -- Optional datacenter-level heap setting, overrides cluster-level setting
     # `cassandra.heap`
-    heap:
-      size:
-      newGenSize:
+    #heap:
```
bradfordcp, line 125: Same as previous comment.

bradfordcp (changes requested):
I'm curious to see if we will go with empty keys (and adjust the default calculation logic) or comment them out and update the docs.

charts/k8ssandra/values.yaml
jeffbanks: @bradfordcp right, the calculated heap feature is something post 1.0 as there are some thoughts about including an up-front `automated survey` of the configuration, which _should_ provide a more accurate calculated set of heap defaults. 

For `v1.0`, we are planning to stay consistent with no empty keye.  

I'll update the inline documentation to reference the `k8ssandra.io` docs.  If not specified in the `values.yaml`, 
the recommended default values reference the cass-operator's [jvm_server_configuration](https://github.com/datastax/cass-operator/blob/master/docs/user/jvm_server_configuration.md).

@jsanda - keep me honest here based on our past discussions in Slack, etc. :smile:

charts/k8ssandra/values.yaml
jeffbanks: Added additional reference as noted in the commit.

charts/k8ssandra/values.yaml
jeffbanks: Added additional documentation reference as noted in the comment.

jakerobb: Am I understanding correctly that, due to these changes, our automatically generated settings docs will not contain specific entires for any of `cassandra.heap`, `cassandra.heap.size`, `cassandra.heap.newGenSize`, `cassandra.datacenter[].heap`, `cassandra.datacenter[].heap.size`, and `cassandra.datacenter[].heap.newGenSize`?

That seems pretty undesirable to me. I'd rather see a fix that modifies the template to make the right decision in the _presence_ of the empty settings. Would that be difficult? (Note that I'm not suggesting we calculate anything -- just do what it already does today.)

charts/k8ssandra/README.md
```diff
@@ -42,14 +42,12 @@ Provisions and configures an instance of the entire K8ssandra stack. This includ
...
 | cassandra.cassandraLibDirVolume.size | string | `"5Gi"` | Size of the provisioned persistent volume per node. It is recommended to keep the total amount of data per node to approximately 1 TB. With room for compactions this value should max out at ~2 TB. This recommendation is highly dependent on data model and compaction strategies in use. Consider testing with your data model to find an optimal value for your usecase.  |
 | cassandra.allowMultipleNodesPerWorker | bool | `false` | Permits running multiple Cassandra pods per Kubernetes worker. If enabled resources.limits and resources.requests **must** be defined. |
-| cassandra.heap | object | `{"newGenSize":null,"size":null}` | Cluster-level heap configuration |
-| cassandra.resources | object | `{}` | Resource requests for each Cassandra pod. |
+| cassandra.resources | object | `{}` | Cluster-level heap configuration size: newGenSize: -- Resource requests for each Cassandra pod. |
```
jakerobb, line 45: Note that leaving the doc-style comment on a commented-out property causes the doc to be concatenated with the docs for the next property. Now it looks like `cassandra.resources` is for heap settings. If those properties are going to be commented out, then their doc comments must _not_ start with `--`.

jakerobb (changes requested):

charts/k8ssandra/README.md
jakerobb: Here's how it looks in the generated doc:
image

charts/k8ssandra/values.yaml
bradfordcp: > Am I understanding correctly that, due to these changes, our automatically generated settings docs will not contain specific entires for any of cassandra.heap, cassandra.heap.size, cassandra.heap.newGenSize, cassandra.datacenter[].heap, cassandra.datacenter[].heap.size, and cassandra.datacenter[].heap.newGenSize?

This is correct (at least in my testing), if the field is commented out the tooling doesn't see something that needs to be annotated.

bradfordcp: > That seems pretty undesirable to me. I'd rather see a fix that modifies the template to make the right decision in the presence of the empty settings. Would that be difficult? (Note that I'm not suggesting we calculate anything -- just do what it already does today.)

I couldn't agree with this more. That's my desired behavior.

charts/k8ssandra/README.md
bradfordcp: Nice catch

charts/k8ssandra/values.yaml
```diff
@@ -73,10 +73,11 @@ cassandra:
   # resources.limits and resources.requests **must** be defined.
   allowMultipleNodesPerWorker: false
 
-  # -- Cluster-level heap configuration
-  heap:
-   size:
-   newGenSize:
+  # -- Cluster-level heap configuration (optional).
```
jakerobb, line 76: To be clear with my requested changes, unless we can uncomment `heap` and its children and modify the template to tolerate their blankness: this line needs to lose the `--`.

charts/k8ssandra/README.md
```diff
@@ -42,14 +42,12 @@ Provisions and configures an instance of the entire K8ssandra stack. This includ
...
 | cassandra.cassandraLibDirVolume.size | string | `"5Gi"` | Size of the provisioned persistent volume per node. It is recommended to keep the total amount of data per node to approximately 1 TB. With room for compactions this value should max out at ~2 TB. This recommendation is highly dependent on data model and compaction strategies in use. Consider testing with your data model to find an optimal value for your usecase.  |
 | cassandra.allowMultipleNodesPerWorker | bool | `false` | Permits running multiple Cassandra pods per Kubernetes worker. If enabled resources.limits and resources.requests **must** be defined. |
-| cassandra.heap | object | `{"newGenSize":null,"size":null}` | Cluster-level heap configuration |
-| cassandra.resources | object | `{}` | Resource requests for each Cassandra pod. |
+| cassandra.resources | object | `{}` | Cluster-level heap configuration (optional). See reference: https://k8ssandra.io/docs/reference/k8ssandra/#values for additional information. size: newGenSize: -- Resource requests for each Cassandra pod. |
```
jsanda, line 45: This is out place. `resources` is for setting the pod's cpu and/or memory.

charts/k8ssandra/values.yaml
jeffbanks: @bradfordcp @jakerobb @jsanda To clarify, we now want the properties to be existing at both the `cluster` and `datacenter` levels to support the `documentation generator` so our docs illustrate the` optional jvm heap` settings.

This was not the [original requirement](https://github.com/k8ssandra/k8ssandra/pull/242#discussion_r564590891) prior to this fix, which is why there is some confusion on my part.

My opinion at this point, I would like to see the heap config keys listed (without default values) available knowing that for now (v 1.0), the startup may fail due to unspecified values.  This is why I opened this item #348 today to consider post-1.0 and catch these using linting rules.  Once something like #348 were in place, the issue of missing values and/or missing configuration keys could be identified early on with some clarity.

jsanda: Let me provide a bit of background. I went back to look at the initial PR #242 to see what was discussed. We decided that we did not want to set defaults, i.e., non-empty values. There is a comment in which I suggest commenting them out after deciding we did not want to set defaults.

I also would like to see generated docs for all of the properties. With that said, if commenting out some or all of the properties make the code easier/simpler to development and maintain, then I will take easier/simpler code.

I have looked at some other chart projects for comparison. I see a common pattern for complex properties, i.e., having subproperties. The top-level property is set to an empty map or array. Comments are included to describe the subproperties. Here are a couples examples for reference:

* https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml#L523
* https://github.com/bitnami/charts/blob/master/bitnami/metrics-server/values.yaml#L198

In general I think we can work with initializing the top-level property to an empty map unless it makes sense to set default values for nested properties. With this approach the doc tooling will pick up the top-level property. We can include comments with examples of nested properties. It also simplifies code because we can avoid having to check any number of nested properties.

@jeffbanks wdyt?

docs/content/en/docs/reference/k8ssandra/_generated.md
```diff
@@ -38,14 +38,12 @@
...
 | cassandra.cassandraLibDirVolume.size | string | `"5Gi"` | Size of the provisioned persistent volume per node. It is recommended to keep the total amount of data per node to approximately 1 TB. With room for compactions this value should max out at ~2 TB. This recommendation is highly dependent on data model and compaction strategies in use. Consider testing with your data model to find an optimal value for your usecase.  |
 | cassandra.allowMultipleNodesPerWorker | bool | `false` | Permits running multiple Cassandra pods per Kubernetes worker. If enabled resources.limits and resources.requests **must** be defined. |
-| cassandra.heap | object | `{"newGenSize":null,"size":null}` | Cluster-level heap configuration |
-| cassandra.resources | object | `{}` | Resource requests for each Cassandra pod. |
+| cassandra.resources | object | `{}` | Cluster-level heap configuration (optional). See reference: https://k8ssandra.io/docs/reference/k8ssandra/#values for additional information. size: newGenSize: -- Resource requests for each Cassandra pod. |
```
jsanda, line 41: Similar to above this is out of place.

charts/k8ssandra/README.md
bradfordcp: Right, this goes to @jakerobb's comment about the comment getting merged into the next field. Removing the `--` fixes the issue, but the comment is not rendered in the docs (which may be what we want).

docs/content/en/docs/reference/k8ssandra/_generated.md
bradfordcp: Right, this goes to @jakerobb's [comment](https://github.com/k8ssandra/k8ssandra/pull/340#discussion_r574079468) about the comment getting merged into the next field. Removing the `--` fixes the issue, but the comment is not rendered in the docs (which may be what we want).

charts/k8ssandra/values.yaml
jsanda: @jeffbanks You bring up a good point about DC-level properties. I would do the same. Set the DC-level `heap` to an empty map.

I have not yet looked at #348 but post-1.0 I want to really look at moving more of the implementation out of Helm charts and into CRDs+controllers.

jeffbanks: @jsanda @bradfordcp 

I like that we would be consistent with others for management of `complex properties`.  As such, for k8ssandra I see the changes scoped to release 1.0 being the following:

1) `values.yaml` updated to include top-level heap property (at cluster and dc levels) with empty map value.
The heap's child properties will be commented out w/ no default values.

```yaml
cassandra:
  # cluster-level settings
  heap: {}
    # size:
    # newGenSize:
  # dc specific settings
  dataCenters:
    - name: dc1
      size: 1
      heap: {}
        # size:
        # newGenSize:
    - name: dc2
      size: 1
```

2) The comment format will follow the other comment entries to avoid any issues with the doc generator rules.  That is, usage of the '--' prefix.  This would resolve issues pointed out by @jakerobb in the generated docs.

3) Unit tests will be adjusted to allow for empty `heap` property values and with or without specified child properties at both levels.

4) If the k8ssandra user doesn't configure the heap settings, there are no default values provided.

charts/k8ssandra/values.yaml
jeffbanks: See latest comment for impact to this.

charts/k8ssandra/README.md
jeffbanks: See latest comment to clarify handling in next PR update.

jeffbanks:
Fixed as a result of requirements stated in #360.

PR #349, opened 2021-02-10
Enable Stargate By Default

### LINKED ISSUE #266: Enable Stargate by default
I've heard that Stargate is part of the K8ssandra distribution, but it did not come up as expected with a vanilla `helm install k8ssandra k8ssandra/k8ssandra`

Stargate should be enabled by default

I've considered writing a values.yaml override or passing the flag via CLI, but it seems like the default should be adjusted.

**Additional context**

jdonenine:
**What this PR does**:

* Enable Stargate by default.
* Enable Stargate Cassandra ingress by default
* Enable Stargate GraphQL Playground ingress by default
* Disable direct Cassandra ingress by default

Docs change removes the blurb in the getting-started section specific to Stargate and instead adds a section generally discussing our approach to the default values.  This is a very simple approach for the moment that I expect we'll expand on as we get into some of the meaty doc tasks for GA.

Fixes #266

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/charts/k8ssandra/Chart.yaml
+++ b/charts/k8ssandra/Chart.yaml
@@ -3,7 +3,7 @@ name: k8ssandra
 description: |
   Provisions and configures an instance of the entire K8ssandra stack. This includes Apache Cassandra, Stargate, Reaper, Medusa, Prometheus, and Grafana.
 type: application
-version: 0.45.0
+version: 0.46.0
 appVersion: 3.11.7
 dependencies:
   - name: cass-operator
--- a/charts/k8ssandra/Chart.yaml
+++ b/charts/k8ssandra/Chart.yaml
@@ -3,8 +3,8 @@ name: k8ssandra
 description: |
   Provisions and configures an instance of the entire K8ssandra stack. This includes Apache Cassandra, Stargate, Reaper, Medusa, Prometheus, and Grafana.
 type: application
-version: 0.46.0
-appVersion: 3.11.7
+version: 0.47.0
+appVersion: 3.11.10
 dependencies:
   - name: cass-operator
     version: 0.28.0
--- a/charts/k8ssandra/Chart.yaml
+++ b/charts/k8ssandra/Chart.yaml
@@ -3,7 +3,7 @@ name: k8ssandra
 description: |
   Provisions and configures an instance of the entire K8ssandra stack. This includes Apache Cassandra, Stargate, Reaper, Medusa, Prometheus, and Grafana.
 type: application
-version: 0.47.0
+version: 0.48.0
 appVersion: 3.11.10
 dependencies:
   - name: cass-operator
--- a/tests/unit/template_cassdc_test.go
+++ b/tests/unit/template_cassdc_test.go
@@ -375,6 +375,7 @@ var _ = Describe("Verify CassandraDatacenter template", func() {
 					"cassandra.auth.enabled":       "true",
 					"backupRestore.medusa.enabled": "true",
 					"repair.reaper.enabled":        "false",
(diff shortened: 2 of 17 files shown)


bradfordcp (commented):
These changes look good.

charts/k8ssandra/Chart.yaml
```diff
@@ -3,8 +3,8 @@ name: k8ssandra
 description: |
   Provisions and configures an instance of the entire K8ssandra stack. This includes Apache Cassandra, Stargate, Reaper, Medusa, Prometheus, and Grafana.
 type: application
-version: 0.46.0
-appVersion: 3.11.7
+version: 0.47.0
+appVersion: 3.11.10
```
jakerobb, line 7: What is the significance of changing this version? Guessing: it's just metadata, and that its intent is to say which version of C* we run by default?

jakerobb (approved):
Looks good to me. Just one question that should not block merging.

burmanm (changes requested):
There are also tests that assume Stargate was disabled by default. They need to be verified and modified to check that it actually is disabled for those cases. Previously tests assumed that if Reaper is disabled - all the extras are. 

For example, the test:

```go
It("enabling auth and medusa with user-defined secret", func() {
```

Checks that "users" part exists, but it will - even if medusa's part fails, since now Stargate adds it. When the test was created, it would have failed, since it disabled reaper.

All tests that assumed something Stargate enablement could modify have to be modified and new tests added if such case with the enabled changes the template rendering.

charts/k8ssandra/Chart.yaml
jdonenine: Yeah, sorry, this is unrelated, I just noticed that when we switched the default to .10 we didn't update this here.

jdonenine:
Thanks for the heads up on those auth related tests, I had fixed the one that was broken because it was testing all of the accounts created, but these definitely not.  I updated all of the tests that were checking the `cassdcv1beta1.CassandraUser` just for existence, in those cases I disabled stargate in those tests, so they continue to test exactly what they were before.

jakerobb:
I resolved the version merge conflict for you. :)

tests/unit/template_cassdc_test.go
```diff
@@ -338,8 +343,8 @@ var _ = Describe("Verify CassandraDatacenter template", func() {
 			Expect(cassdc.Spec.PodTemplateSpec.Spec.Containers[0].Env).To(BeNil())
 			// No initcontainers slice should be present
 			Expect(cassdc.Spec.PodTemplateSpec.Spec.InitContainers).To(BeNil())
-			// No users should exist
-			Expect(cassdc.Spec.Users).To(BeNil())
+			// With Stargate enabled by default, only the stargate user should exist
+			Expect(cassdc.Spec.Users).To(ConsistOf(cassdcv1beta1.CassandraUser{Superuser: true, SecretName: clusterName + "-stargate"}))
```
jdonenine, line 347: @burmanm I updated the test you modified in #357 as discussed over there.

burmanm: Ah, the test was designed to verify that empty users never exists (nil, not size 0), since that fails the CassandraDatacenter deployment. Other tests can verify that the correct user appears.

jdonenine: Ah, OK, so then we actually want this one to disable stargate as well.

jdonenine: OK, swapped it to disable stargate for the test and switched back the check to be the nil check.

burmanm (approved):

PR #355, opened 2021-02-11
Make garbage collection configurable

### LINKED ISSUE #220: Make garbage collection in Cassandra configurable
Because garbage collection tuning can have a major impact on Cassandra's performance, I think we need to expose some GC configuration for the 1.0 release. As with other configuration settings, we should take into consideration best practices and strive to be opinionated. We also need to changes around 4.0. My understanding is that C* 4.0 will support Java 11, so we may want to expose config settings that specific to 4.0 as well.

This may involve a bit of research and we should create additional tickets as necessary.

jsanda:
**What this PR does**:
Makes GC configurable

Fixes #220

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,9 +23,6 @@ test: fmt vet
 integ-test:
 	go test -v -test.timeout=5m ./tests/integration/... -coverprofile cover.out
 
-test-gc: fmt vet
-	go test -v -test.timeout=3m ./tests/unit/... -coverprofile cover.out -args -ginkgo.focus=".*garbage\ collection.*"
-
 fmt:
 	go fmt ./pkg/...
 	go fmt ./tests/...
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,9 @@ test: fmt vet
 integ-test:
 	go test -v -test.timeout=5m ./tests/integration/... -coverprofile cover.out
 
+test-gc: fmt vet
+	go test -v -test.timeout=3m ./tests/unit/... -coverprofile cover.out -args -ginkgo.focus=".*garbage\ collection.*"
+
 fmt:
 	go fmt ./pkg/...
 	go fmt ./tests/...
--- a/charts/k8ssandra/templates/_helpers.tpl
+++ b/charts/k8ssandra/templates/_helpers.tpl
@@ -199,9 +199,17 @@ Creates Cassandra auth environment variables if authentication is enabled.
 {{- end }}
 
 {{- define "k8ssandra.configureGc" -}}
-{{- if .Values.cassandra.gc.cms.enabled -}}
+{{- $datacenter := (index .Values.cassandra.datacenters 0) }}
+{{- $gc := "" -}}
+{{- if $datacenter.gc }}
+  {{ $gc = $datacenter.gc }}
+{{- else -}}
+  {{ $gc = .Values.cassandra.gc }}
+{{- end -}}
+
+{{- if $gc.cms.enabled -}}
   {{- nindent 6 "garbage_collector: CMS" -}}
-  {{ with .Values.cassandra.gc.cms }}
+  {{ with $gc.cms }}
     {{- if .survivorRatio }}
       survivor_ratio: {{ .survivorRatio }}
     {{- end }}
@@ -215,9 +223,9 @@ Creates Cassandra auth environment variables if authentication is enabled.
       cms_wait_duration: {{ .waitDuration }}
     {{- end }}
   {{- end }}
-{{- else if .Values.cassandra.gc.g1.enabled -}}
+{{- else if $gc.g1.enabled -}}
   {{- nindent 6 "garbage_collector: G1" -}}
-  {{ with .Values.cassandra.gc.g1 }}
+  {{ with $gc.g1 }}
     {{- if .setUpdatingPauseTimePercent }}
       g1r_set_updating_pause_time_percent: {{ .setUpdatingPauseTimePercent }}
     {{- end }}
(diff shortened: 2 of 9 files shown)


charts/k8ssandra/templates/_helpers.tpl
```diff
@@ -62,4 +62,59 @@ Create the name of the service account to use
 
 {{- define "k8ssandra.datacenterSize" -}}
 {{ (index .Values.cassandra.datacenters 0).size }}
+{{- end }}
+
+{{- define "k8sandra.checkGcEnabled" -}}
+{{- if .Values.cassandra.gc }}
+  {{- if and .Values.cassandra.gc.cms.enabled .Values.cassandra.gc.g1.enabled }}
+    {{- fail "Only one of the CMS and G1 garbage collectors can be enabled" }}
```
burmanm, line 70: Why?

charts/k8ssandra/values.yaml
```diff
@@ -80,6 +80,21 @@ cassandra:
    size:
    newGenSize:
 
+  gc:
```
burmanm, line 83: Why this sort of structure? Now we prevent use of ZGC or Shenandoah. I think I'd rather see a ConfigMap or such to make sure the user has to actually understand what they're doing (by having to define the parameters themselves).

jsanda: We cannot simply use a ConfigMap. We have to work within the constraints of the config-builder since that is what generate the `jvm-options` for 3.11 or the `jvm8-server.options` or `jvm11-server.options` for 4.0. We will need to update config-builder to support additional garbage collectors.

jsanda:
This PR only adds support for CMS and G1. Config files are created by cass-config-builder and that is all it currently supports. 

This PR got a lot more involved than I was expecting. See [here](http://github.com/k8ssandra/k8ssandra/blob/master/charts/k8ssandra/templates/_helpers.tpl#L92-L92) for an explanation of the logic.

Makefile
```diff
@@ -23,6 +23,9 @@ test: fmt vet
 integ-test:
 	go test -v -test.timeout=5m ./tests/integration/... -coverprofile cover.out
 
+test-gc: fmt vet
```
burmanm, line 26: Any reason we would need to separate these tests?

jsanda: I need to add a comment. I didn't want or need to run all the tests and I didn't want to manually keep typing `-args -ginkgo.focus=".*garbage\ collection.*"`. It might be helpful to help additional Makefile targets for running subsets of the tests. I am also fine with removing if you think it is confusing.

jdonenine: Actually, as a total newbie to running these tests for the first time last week I would have loved to at least have an example of how to target the test suite so that I wouldn't have to wait through and sift through the logging for all of them.  I think this would be a good add if for nothing else than that purpose.

jsanda: I would like to add similar, targets for other tests. The verbose output is painful at times. I looked before briefly at cut down the output to just failures but I was unsuccessful.

burmanm (approved):

Makefile
jsanda: >The alternative (while doing development) is to make the test use a F infront of the test

Do I need to remove the `F` if I want to remove the focus? If so, that is not a great general solution.

charts/k8ssandra/values.yaml
```diff
@@ -81,6 +81,25 @@ cassandra:
    #size:
    #newGenSize:
 
+  # -- Optional cluster-level garbage collection configuration. It can be overridden at
+  # the datacenter level.
+  gc:
+    # -- GC configuration for the CMS collector.
+    cms: {}
```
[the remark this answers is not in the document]
jsanda: These are good questions. I could have used an empty string instead. I chose `{}` because I have seen that in other charts. To me it indicates that the property has child properties. Will that be clear to the k8ssandra users, particularly those with little or no prior Helm experience? Doubtful. We need to document it.

tests/unit/testdata/cluster-g1-dc-cms-values.yaml
```diff
@@ -0,0 +1,26 @@
...
+      concurrentGcThreads: 18
+    cms:
+      enabled: false
+      survivorRatio: 6
+      maxTenuringThreshold: 1
+      initiatingOccupancyFraction: 75
+      waitDuration: 11000
+
+  datacenters:
+    - name: dc1
```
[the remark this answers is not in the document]
jsanda: I see the upside of a more dynamic approach. In some regards it is less maintenance. The user is free to add arbitrary properties which don't require any code updates in k8ssandra. It gives the user more flexibility and reduces maintenance burden on the code base in some regards.

There are some downsides though. First, the user has to understand what values are supported by cass-config-builder. It does not expect arbitrary values from cassandra.yaml and from jvm.options. 

Secondly, a yaml map would make parsing harder. While users could supply arbitrary values, we would need to verify that those values match what is expected/allow by cass-config-builder.

Ultimately I do think we should provide a hook of some sorts to allow users to set arbitrary values even if it does mean understanding cass-config-builder. I would consider this applicable for advanced use cases.

Lastly, I hope we will be able to put some higher level abstractions in place that could free the user from worrying about some of these settings.

Makefile
burmanm: Yeah. It's intended for development time focusing.

jsanda:
@burmanm @adejanovski can you do another review of this?

Note that the PR exposes what is available through [cass-config-definitions](https://github.com/datastax/cass-config-definitions) which is a subset of all available options.

PR #370, opened 2021-02-12
Removing the pathType as suggested by Michael to resolve backwards co…

jdonenine:
…mpatibility issues with older versions of kubernetes.

**What this PR does**:

Removes the pathType specification in the template deployed for the stargate ingress resources.  Per discussion in #366 it's not supported < 1.18.

I did some testing to make sure that after the change ingress was working as expected, things seem good:

```
% curl -L -X POST 'http://stargate.127.0.0.1.nip.io:8081/v1/auth' -H 'Content-Type: application/json' --data-raw '{"username": "k8ssandra-cluster-superuser", "password": "XOHBHcJs_f8CFcciczCJYjmZEvoJmpBKpHVTVjO3G386yqWV6HtrEA"}'
{"authToken":"f9e58ea3-8460-452a-8584-01e3479ffcac"}%                                  
```

```
% curl --location --request GET 'http://stargate.127.0.0.1.nip.io:8080/v2/schemas/namespaces' \
--header "x-cassandra-token: f9e58ea3-8460-452a-8584-01e3479ffcac"
{"data":[{"name":"medusa_test"},{"name":"system_distributed"},{"name":"system"},{"name":"data_endpoint_auth"},{"name":"system_schema"},{"name":"stargate_system"},{"name":"reaper_db","datacenters":[{"name":"dc1","replicas":1}]},{"name":"system_auth"},{"name":"system_traces"}]}%   
```

```                                                                    
% curl -L \
-X GET 'http://stargate.127.0.0.1.nip.io:8080/v2/schemas/keyspaces/'medusa_test'/tables/users' \
-H "X-Cassandra-Token: f9e58ea3-8460-452a-8584-01e3479ffcac" \
-H "accept: application/json"
{"data":{"name":"users","keyspace":"medusa_test","columnDefinitions":[{"name":"email","typeDefinition":"varchar","static":false},{"name":"name","typeDefinition":"varchar","static":false},{"name":"state","typeDefinition":"varchar","static":false}],"primaryKey":{"partitionKey":["email"],"clusteringKey":[]},"tableOptions":{"defaultTimeToLive":0,"clusteringExpression":[]}}}%    
```

```
% wget http://stargate.127.0.0.1.nip.io:8080/playground
--2021-02-12 10:58:40--  http://stargate.127.0.0.1.nip.io:8080/playground
Resolving stargate.127.0.0.1.nip.io (stargate.127.0.0.1.nip.io)... 127.0.0.1
[...]

PR #372, opened 2021-02-12
#361: refactor Stargate ingress configuration

### LINKED ISSUE #361: Rework ingress configuration in values.yaml
In #253 and #342, @bradfordcp described a new, cleaner approach to configuration of ingress for the various components which will become necessary when we split off separate subcharts.

Reaper:
```
repair:
  reaper:
    # ... existing configuration
    ingress:
      enabled: true
      # -- Optional hostname used to match requests
      host:
      # tls:
      #   # -- Hostnames to match as part of SNI routing
      #   hosts:
      #     - example.com
      #   # -- Secret containing the TLS certificate and key for encrypting requests to this endpoint
      #   secretName: my-secret
```

Stargate:
```
stargate:
  # ... existing stargate configuration
  
  # New pieces here
  ingress:
    auth:
      enabled: true
      # -- Optional hostname used to match requests
      host:
      # tls:
      #   # -- Hostnames to match as part of SNI routing
      #   hosts:
      #     - example.com
      #   # -- Secret containing the TLS certificate and key for encrypting requests to this endpoint
      #   secretName: my-secret

    rest:
      # -- Enables REST
      enabled: true
      # -- Optional hostname used to match requests
      host:
      # tls:
      #   # -- Hostnames to match as part of SNI routing
      #   hosts:
      #     - example.com
      #   # -- Secret containing the TLS certificate and key for encrypting requests to this endpoint
      #   secretName: my-secret

    graphql:
      # -- Enables GraphQL API endpoint ingress
      enabled: true
      # -- Optional hostname used to match requests
      host:
      # tls:
      #   # -- Hostnames to match as part of SNI routing
      #   hosts:
      #     - example.com
      #   # -- Secret containing the TLS certificate and key for encrypting requests to this endpoint
      #   secretName: my-secret

    cassandra:
      # -- Enables C* native protocol ingress with Traefik
      enabled: false
      # -- Determines which TCP-based ingress custom resources to template out. Currently only `traefik` is supported
      method: traefik
      # -- Parameters used by the Traefik IngressRoute custom resource
      traefik:
        entrypoint: cassandra
        # Future parameters
        # tls:
        #   client_auth:
        #     # -- Collection of Secrets containing TLS certificates. KEYS ARE NOT REQUIRED HERE
        #     ca_secrets:
        #       - ca_secret_1
        #   domains:
        #     - example.com
        #   # -- Secret containing the TLS certificate and key for the listener
        #   secretName: tls-secret
        
```

And so on. 

Leaving those issues open to tackle the Reaper- and Stargate-specific changes not related to restructuring values.yaml, I'm spinning this issue off to do just the restructuring part of the work.

### LINKED ISSUE #368: Stargate deploys resources even when disabled
## Bug Report

**Describe the bug**
Even if Stargate is disabled, it deploys resources such as the ingress:

➜  k8ssandra git:(remove_grafana_operator_crds) ✗ kubectl get ingress
NAME                                    HOSTS                             ADDRESS   PORTS   AGE
demo-grafana                            grafana.192.168.1.112.xip.io                80      9m50s
demo-kube-prometheus-stack-prometheus   prometheus.192.168.1.112.xip.io             80      9m50s
k8ssandra-stargate                      *                                           80      9m50s
➜  k8ssandra git:(remove_grafana_operator_crds) ✗

➜  k8ssandra git:(remove_grafana_operator_crds) ✗ kubectl get ingressroutetcp
NAME                AGE
demo-dc1-stargate   11m
➜  k8ssandra git:(remove_grafana_operator_crds) ✗

This is because the Stargate ingress only verifies if Stargate ingress is enabled / disabled, not if Stargate itself is deployed.

**To Reproduce**
1. Disable stargate
2. Deploy k8ssandra

**Expected behavior**
If Stargate is disabled, nothing extra is deployed.

### LINKED ISSUE #371: Stargate Cassandra ingress is configured incorrectly
## Bug Report

**Describe the bug**

When deploying stargate with ingress (the new default) the Cassandra ingress has errors and cannot be used:

image

**To Reproduce**
1. Install traefik

(For Kind):
```
---
providers:
  kubernetesCRD:
    namespaces:
      - default
      - traefik
  kubernetesIngress:
    namespaces:
      - default
      - traefik

ports:
  traefik:
    expose: true
    nodePort: 32090
  web:
    nodePort: 32080
  websecure:
    nodePort: 32443
  cassandra:
    expose: true
    port: 9042
    nodePort: 32091
  cassandrasecure:
    expose: true
    port: 9142
    nodePort: 32092
  sg-graphql:
    expose: true
    port: 8080
    nodePort: 30080
  sg-auth:
    expose: true
    port: 8081
    nodePort: 30081
  sg-rest:
    expose: true
    port: 8082
    nodePort: 30082

service:
  type: NodePort
```

2. Deploy k8ssandra:

```
cassandra:
  version: "3.11.10"
  auth:
    enabled: false
  clusterName: k8ssandra-cluster
  datacenters:
  - name: dc1
    size: 1
kube-prometheus-stack:
  prometheus:
    enabled: true
    prometheusSpec:
      externalUrl: http://localhost:9090/prometheus
      routePrefix: /prometheus
    ingress:
      enabled: true
      paths:
        - /prometheus
  grafana:
    enabled: true
    ingress:
      enabled: true
      path: /grafana
    adminUser: admin
    adminPassword: admin123
    grafana.ini:
      server:
        root_url: http://localhost:3000/grafana
        serve_from_sub_path: true
ingress:
  traefik:
    enabled: true
    repair:
      host: repair.127.0.0.1.nip.io
    stargate:
      host: stargate.127.0.0.1.nip.io
backupRestore:
  medusa:
    enabled: true
    storage: s3
    bucketName: "jdinoto-k8ssandra-test"
    bucketSecret: "medusa-bucket-key"
```
3. Populate the cluster with some data

test-data.cql
```
CREATE KEYSPACE medusa_test WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 3};
USE medusa_test;
CREATE TABLE users (email text primary key, name text, state text);
insert into users (email, name, state) values ('john@gamil.com', 'John Smith', 'NC');
insert into users (email, name, state) values ('joe@gamil.com', 'Joe Jones', 'VA');
insert into users (email, name, state) values ('sue@help.com', 'Sue Sas', 'CA');
insert into users (email, name, state) values ('tom@yes.com', 'Tom and Jerry', 'NV');
```

```
kubectl exec -it k8ssandra-cluster-dc1-default-sts-0 -c cassandra -- mkdir /tmp/scripts/
```

```
kubectl cp ./test-data.cql k8ssandra-cluster-dc1-default-sts-0:/tmp/scripts -c cassandra
```

```
kubectl exec -it k8ssandra-cluster-dc1-default-sts-0 -c cassandra -- cqlsh -f /tmp/scripts/test-data.cql
```

```
kubectl exec -it k8ssandra-cluster-dc1-default-sts-0 -c cassandra -- cqlsh -e "select * from medusa_test.users;"
```
4. Observe the Traefik dashboard has errors as shown above and that access via CQL fails:

get-data.py
```
from cassandra.cluster import Cluster
from cassandra.auth import PlainTextAuthProvider

auth_provider = PlainTextAuthProvider(username='k8ssandra-cluster-superuser', password='XOHBHcJs_f8CFcciczCJYjmZEvoJmpBKpHVTVjO3G386yqWV6HtrEA')
cluster = Cluster(['stargate.127.0.0.1.nip.io'], auth_provider=auth_provider)
session = cluster.connect('medusa_test')
print(session.execute("SELECT * FROM users").one())
```

```
python get-data.py
```

**Expected behavior**

Access to Cassandra via the python driver used in the example should work as if it was the natively exposed Cassandra CQL access.

**Screenshots**

* Helm charts version info

```
% helm ls -A
NAME            NAMESPACE       REVISION        UPDATED                                 STATUS          CHART                   APP VERSION
k8ssandra       default         1               2021-02-12 09:44:05.487311 -0500 EST    deployed        k8ssandra-0.49.0        3.11.10    
traefik         default         1               2021-02-12 09:43:50.885753 -0500 EST    deployed        traefik-9.14.2          2.4.2  
``` 

```
% helm get values k8ssandra
USER-SUPPLIED VALUES:
backupRestore:
  medusa:
    bucketName: jdinoto-k8ssandra-test
    bucketSecret: medusa-bucket-key
    enabled: true
    storage: s3
cassandra:
  auth:
    enabled: false
  clusterName: k8ssandra-cluster
  datacenters:
  - name: dc1
    size: 1
  version: 3.11.10
ingress:
  traefik:
    enabled: true
    repair:
      host: repair.127.0.0.1.nip.io
    stargate:
      host: stargate.127.0.0.1.nip.io
kube-prometheus-stack:
  grafana:
    adminPassword: admin123
    adminUser: admin
    enabled: true
    grafana.ini:
      server:
        root_url: http://localhost:3000/grafana
        serve_from_sub_path: true
    ingress:
      enabled: true
      path: /grafana
  prometheus:
    enabled: true
    ingress:
      enabled: true
      paths:
      - /prometheus
    prometheusSpec:
      externalUrl: http://localhost:9090/prometheus
      routePrefix: /prometheus
```

```
% k version
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T12:50:19Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.17", GitCommit:"f3abc15296f3a3f54e4ee42e830c61047b13895f", GitTreeState:"clean", BuildDate:"2021-01-22T22:20:06Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

* Kubernetes cluster kind:
Kind

**Additional context**
@jakerobb confirmed that he was already aware of this and would include it in his current work.

jakerobb:
**What this PR does**:
Converts to the new values.yaml structure for ingress config. 

Progress:
 - [x] Stargate
 - [x] Stargate Cassandra
 - [x] Cassandra
 - [x] Reaper

Fixes #361.
Fixes #371, which Jeff handed off to me from #370.
Fixes #368 

**Checklist**
- [x] Changes manually tested


charts/k8ssandra/templates/stargate/cassandra-ingress.yaml
```diff
@@ -1,14 +1,14 @@
...
-{{- if $traefik.enabled }}
+{{- $stargateIngress := .Values.stargate.ingress -}}
+{{- if $stargateIngress.enabled }}
 {{- $releaseName := .Release.Name -}}
 {{- $datacenterName := include "k8ssandra.datacenterName" . -}}
 
-{{- if $traefik.stargate.enabled }}
-{{- if and $traefik.stargate.cassandra.enabled $traefik.cassandra.enabled }}
-  {{- fail "ingress.traefik.stargate.cassandra.enabled and ingress.traefik.cassandra.enabled cannot both be enabled." }}
+{{- if and $stargateIngress.cassandra.enabled .Values.ingress.traefik.cassandra.enabled }}
```
bradfordcp, line 6: Is it safe to assume `.Values.ingress.traefik.cassandra.enabled` will be something like `$cassandra.ingress.enabled` as part of a separate ticket?

charts/k8ssandra/templates/stargate/ingress.yaml
```diff
@@ -1,14 +1,15 @@
...
-{{- if $traefik.stargate.enabled }}
+{{- define "k8ssandra.datacenterSize" -}}
+  {{ (index .Values.cassandra.datacenters 0).size }}
+{{- end }}
+{{- $stargateIngress := .Values.stargate.ingress -}}
+{{- if $stargateIngress.enabled }}
 
-{{- if and $traefik.stargate.cassandra.enabled $traefik.cassandra.enabled }}
-  {{- fail "ingress.traefik.stargate.cassandra.enabled and ingress.traefik.cassandra.enabled cannot both be enabled." }}
+{{- if and .Values.ingress.traefik.cassandra.enabled $stargateIngress.cassandra.enabled }}
```
bradfordcp, line 7: Is it safe to assume `.Values.ingress.traefik.cassandra.enabled` will be something like `$cassandra.ingress.enabled` as part of a separate ticket?

charts/k8ssandra/values.yaml
```diff
@@ -169,6 +169,66 @@ stargate:
     secret: ""
     username: ""
 
+  ingress:
+    # -- Optional hostname used to match requests
+    host:
+    auth:
+      enabled: true
```
bradfordcp, line 176: Should this have a comment?

```suggestion
      # -- Enables authentication ingress
      enabled: true
```

bradfordcp: Follow-up question, should we fail if authentication is disabled, but the other endpoints are enabled?

bradfordcp (commented):
Awesome changes, I left a couple comments from an initial scan through.

charts/k8ssandra/templates/stargate/cassandra-ingress.yaml
bradfordcp: > Reaper, and Non-Stargate Cassandra will come in later commits to this PR.

I'm going to close this question since you already answered it in the PR description.

charts/k8ssandra/templates/stargate/ingress.yaml
bradfordcp: > Reaper, and Non-Stargate Cassandra will come in later commits to this PR.

I'm going to close this question since you already answered it in the PR description.

charts/k8ssandra/templates/stargate/cassandra-ingress.yaml
jakerobb: Yes, part of this PR, coming in later commits.

charts/k8ssandra/templates/stargate/ingress.yaml
jakerobb: Yes, part of this PR, coming in later commits.

charts/k8ssandra/values.yaml
jakerobb: I considered that as well. Prior to this refactor, it was not possible to enable Stargate ingress without enabling it for auth. I had made that choice deliberately at the time, thinking that if `ingress.traefik.stargate.enabled=true` was sufficient. Some experience showed that layers of `enabled` are pretty obnoxious, and so I didn't want to keep that aspect of the structure. In the new setup, there is no `stargate.ingress.enabled` -- only `stargate.ingress.[auth|rest|graphql|cassandra].enabled`. They're all on by default (and, of course, they have no effect unless you've also installed an ingress controller).

It would be weird to disable auth but enable other aspects of Stargate ingress, but I don't think it would be invalid. Consider the possibility that auth could be handled by some kind of second- or third-party SSO service running in the same Kubernetes cluster, and having direct access to Stargate auth is actively undesirable.

bradfordcp: I need to dig into the stargate auth service a bit more. If it's not required to talk with stargate then I don't care, but if it's required to use graphql or rest, maybe we tie whether it is enabled with something like `or stargate.ingress.rest.enabled stargate.ingress.graphql.enabled` instead of a dedicated field / toggle.

jsanda: When we first introduced support for Stargate, authentication had to be enabled. It did not work when it was disabled. @jakerobb do you know if this has changed?

jsanda (changes requested):
There is a template error:

```
$ helm template test ./k8ssandra --debug
install.go:172: [debug] Original chart version: ""
install.go:189: [debug] CHART PATH: /Users/jsanda/Development/k8ssandra/k8ssandra/charts/k8ssandra


Error: parse error at (k8ssandra/templates/_helpers.tpl:140): unexpected EOF
helm.go:81: [debug] parse error at (k8ssandra/templates/_helpers.tpl:140): unexpected EOF
```
Looks like there is an `{{- end }}` missing at line 140 in _helpers.tpl.

charts/k8ssandra/values.yaml
```diff
@@ -169,6 +169,67 @@ stargate:
     secret: ""
     username: ""
 
+  ingress:
+    # -- Optional hostname used to match requests
```
jsanda, line 173: The `stargate.ingress.enabled` property is missing.

jsanda: >Some experience showed that layers of enabled are pretty obnoxious, and so I didn't want to keep that aspect of the structure. 

I have mixed feelings on this and it probably merits more discussion. 

>In the new setup, there is no stargate.ingress.enabled -- only stargate.ingress.[auth|rest|graphql|cassandra].enabled. They're all on by default (and, of course, they have no effect unless you've also installed an ingress controller).

As things are currently implemented though in your PR, `stargate.ingress.enabled` has to be set; otherwise, no Ingress objects will be created. You do set the property in the unit tests.

jsanda (changes requested):

jsanda (changes requested):
The `name` property in k8ssandra/templates/stargate/ingress.yaml is hardcoded as `k8ssandra-stargate`.

If we only created the one Ingress object I would just set it to the release name. Since we create multiple Ingress objects that won't work. My suggestion then is `<releasName>-stargate`.

I noticed that the Reaper ingress name is also hardcoded to the same value. I would set it `<releaseName>-reaper`.

charts/k8ssandra/values.yaml
jakerobb: @jsanda this is just whether to enable ingress for Stargate's auth resource, not whether Stargate is configured to use or not use authentication.

charts/k8ssandra/values.yaml
jakerobb: > As things are currently implemented though in your PR, stargate.ingress.enabled has to be set; otherwise, no Ingress objects will be created. You do set the property in the unit tests.

Yep, hence draft mode. I believe it's normalized now. I'm willing to have the discussion about whether is is the right move.

jakerobb:
> The `name` property in k8ssandra/templates/stargate/ingress.yaml is hardcoded as `k8ssandra-stargate`.
> 
> If we only created the one Ingress object I would just set it to the release name. Since we create multiple Ingress objects that won't work. My suggestion then is `<releasName>-stargate`.
> 
> I noticed that the Reaper ingress name is also hardcoded to the same value. I would set it `<releaseName>-reaper`.

Can do. Just checked: Stargate's ingress is named as you said, but Reaper's is `<releaseName>-k8ssandra-repair`.

Whatever we do, I'd like to be consistent. Which do we want?
1. `<releaseName>-k8ssandra-<component>`
2. `<releaseName>-<component>`
3. `<releaseName>-k8ssandra-<componentRole>`
4. `<releaseName>-<componentRole>`

("component" is `reaper" and "componentRole" is "repair")

If #3 or #4, I'm assuming Stargate is the right name either way? Otherwise we'd have to come up with a name for _what Stargate is_, i.e. some hopefully-much-more-concise version of "hybrid coordinator and API gateway." And if my assumption is correct, this is really just about the distinction between `repair` and `reaper`.

jdonenine:
> EDIT: I've tentatively gone with option 3, updating the Stargate ingress resource's name to `<releaseName>-k8ssandra-stargate` and leaving Reaper's as `<releaseName>-k8ssandra-repair`.

I like the `-componentRole` approach you have here, that I guess we had previously with Reaper.  Decouples us a bit incase we were to ever swap in a new component to fill that roll down the road or something like that.

jakerobb:
> > EDIT: I've tentatively gone with option 3, updating the Stargate ingress resource's name to `<releaseName>-k8ssandra-stargate` and leaving Reaper's as `<releaseName>-k8ssandra-repair`.
> 
> I like the `-componentRole` approach you have here, that I guess we had previously with Reaper. Decouples us a bit incase we were to ever swap in a new component to fill that roll down the road or something like that.

Agreed. That stance rules out options 1 and 2. Anybody prefer option 4 over option 3?

jsanda:
@jdonenine @jakerobb with the respect to the naming, my vote is for `<releaseName>-<component>`. In the case of Reaper, it is a pretty well-known and established tool among SREs. I feel like `reaper` in the name provides more clarity than `repair` in this regard.

>Decouples us a bit incase we were to ever swap in a new component to fill that roll down the road or something like that.

Note that we had moved forward with putting the Reaper templates in their own chart, the chart name would be `reaper` and the resources created would mostly likely therefore have `reaper` in their names.

I am confident that we will be able to better address the naming when we move to more of an operator based implementation.

jdonenine:
> @jdonenine @jakerobb with the respect to the naming, my vote is for `<releaseName>-<component>`. In the case of Reaper, it is a pretty well-known and established tool among SREs. I feel like `reaper` in the name provides more clarity than `repair` in this regard.

Thinking a bit more about it, while I do like the decoupling of the direct components, I realize that very quickly breakdown with stargate itself.  If it's `-stargate` that's the component, not the role, I don't even know what the "role" would be, `api-gateway` or something maybe?  But that's a super loaded term I wouldn't mess with.

It's perhaps just simpler and more understandable to go with the component names `-stargate` `-reaper` etc like you suggest @jsanda

jsanda (changes requested):
I made an initial pass through the Reaper changes and everything looks good. I still need to do some manual testing.

I think that there are only a couple other outstanding things left:

* reintroduce `stargate.ingress.enabled`
* add comments in values.yaml about how cqlsh resolves hostnames
* set the `host` properties for the cassandra ingress to empty string.

jakerobb:
I've pushed changes to fix #368 and reintroduce `stargate.ingress.enabled`.

charts/k8ssandra/values.yaml
```diff
@@ -129,6 +129,35 @@ cassandra:
...
+    method: traefik
+
+    # -- Optional hostname used to match requests. Warning: many native Cassandra clients, notably including cqlsh, initialize their connection by querying for
+    # the cluster's contactPoints, and thereafter communicate to the cluster using those names/IPs rather than whatever host was specified to the client. In
+    # order for clients to work correctly through ingress with a host filter, this means that the host filter must match the hostnames specified in the
+    # contactPoints. This value must be a DNS-resolvable hostname and not an IP address. To avoid this issue, leave this setting blank.
+    host:
+
+    traefik:
+      # -- Traefik entrypoint where traffic is sourced.
```
jsanda, line 148: This comment is referencing a property from the traefik chart installation. We should reference the docs to make it clear where it is coming from.

charts/k8ssandra/values.yaml
```diff
@@ -169,53 +198,148 @@ stargate:
...
+      # the cluster's contactPoints, and thereafter communicate to the cluster using those names/IPs rather than whatever host was specified to the client. In
+      # order for clients to work correctly through ingress with a host filter, this means that the host filter must match the hostnames specified in the
+      # contactPoints. This value must be a DNS-resolvable hostname and not an IP address. To avoid this issue, leave this setting blank, or if
+      # stargate.ingress.host is set, override it here to "" (empty string). This note does not apply to clients of Stargate's auth, REST, or GraphQL APIs.
+      host:
+
+      # -- Parameters used by the Traefik IngressRoute custom resource
+      traefik:
+
+        # -- Traefik entrypoint where traffic is sourced.
```
jsanda, line 267: This comment is referencing a property from the traefik chart installation. We should reference the docs to make it clear where it is coming from.

charts/k8ssandra/values.yaml
```diff
@@ -169,53 +198,148 @@ stargate:
...
+    # -- Hostname to use for routing requests to the repair UI. If
+    # using a local deployment consider leveraging dynamic DNS services like
+    # xip.io. Example: `repair.127.0.0.1.xip.io` will return `127.0.0.1` for
+    # DNS requests routing requests to your local machine.
+    host:
+
+    method: traefik
+
+    traefik:
+      # -- Traefik entrypoint where traffic is sourced.
```
jsanda, line 340: This comment is referencing a property from the traefik chart installation. We should reference the docs to make it clear where it is coming from.

charts/k8ssandra/values.yaml
```diff
@@ -169,53 +198,148 @@ stargate:
...
+
+  ingress:
+    # -- Enables Reaper ingress definitions.
+    enabled: false
+
+    # -- Hostname to use for routing requests to the repair UI. If
+    # using a local deployment consider leveraging dynamic DNS services like
+    # xip.io. Example: `repair.127.0.0.1.xip.io` will return `127.0.0.1` for
+    # DNS requests routing requests to your local machine.
+    host:
```
jsanda, line 335: The comments here should indicate that `host` is required when reaper ingress is enabled.

charts/k8ssandra/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: k8ssandra
 description: |
   Provisions and configures an instance of the entire K8ssandra stack. This includes Apache Cassandra, Stargate, Reaper, Medusa, Prometheus, and Grafana.
 type: application
-version: 0.49.1
+version: 0.50.0
```
jsanda, line 6: You need to bump the version to 0.52.0.

charts/k8ssandra/templates/reaper/ingress.yaml
```diff
@@ -12,15 +13,14 @@ metadata:
 {{ include "k8ssandra.labels" . | indent 4 }}
```
jsanda, line 13: As per the discussion in https://github.com/k8ssandra/k8ssandra/pull/372#issuecomment-780879442 can you update the name to be `{ $releaseName }}-reaper`? 

I know we haven't been consistent on the naming, but I feel like we have been in the process of figuring out. I don't necessarily want to go through all the code and make sweeping name changes but it would be good to make changes along the way to get things more consistent.

jsanda (changes requested):

charts/k8ssandra/templates/reaper/ingress.yaml
jakerobb: Done.

charts/k8ssandra/Chart.yaml
jakerobb: Done.

charts/k8ssandra/values.yaml
jakerobb: Done.

charts/k8ssandra/values.yaml
jakerobb: Done.

charts/k8ssandra/values.yaml
jakerobb: Done.

charts/k8ssandra/values.yaml
jakerobb: Done.

jakerobb:
All feedback has been addressed. I think this is ready for final testing and merge.

jsanda (approved):
@jakerobb Thanks for all the hard work on this!

PR #382, opened 2021-02-17
Set the default value of tokens for each supported version

### LINKED ISSUE #324: Provide a better default for num_tokens
The default behavior of cass-operator is to create the Cassandra cluster with vnodes enabled and a single token per node. As an aside cass-operator does not support manual token assignments.

The out of box default for Cassandra is `num_tokens: 256`. Experience over time has shown that this is not a good default. In fact a couple Netflix engineers published a white paper titled [Cassandra Availability with Virtual Nodes](https://jolynch.github.io/pdf/cassandra-availability-virtual.pdf) that discusses how 256 tokens actually decreases availability.

Given the problems with using 256 tokens, it makes sense that cass-operator chooses a different value. The `nodetool` output from a couple test clusters demonstrate the problem with a single token:

```
# 3 node cluster

$ kubectl exec -it test-dc1-default-sts-0 -c cassandra -- nodetool status
Datacenter: dc1
===============
Status=Up/Down
|/ State=Normal/Leaving/Joining/Moving
--  Address     Load       Tokens       Owns (effective)  Host ID                               Rack
UN  10.40.0.8   70.94 KiB  1            31.8%             42f9ff07-927e-4e81-8de7-59232d28de89  default
UN  10.40.1.67  71.07 KiB  1            88.9%             f4729051-86be-42f1-9d0b-a09d04761795  default
UN  10.40.2.20  85.29 KiB  1            79.3%             37c7b5bb-a573-439e-bcc8-8b1c9fd08217  default
```

```
# 9 node cluster

$ kubectl exec -it test-dc1-default-sts-0 -c cassandra -- nodetool status
Datacenter: dc1
===============
Status=Up/Down
|/ State=Normal/Leaving/Joining/Moving
--  Address     Load       Tokens       Owns (effective)  Host ID                               Rack
UN  10.40.0.10  90.23 KiB  1            77.5%             5af74306-ec08-4176-aaad-6dae7ad9f950  default
UN  10.40.2.26  90.23 KiB  1            39.9%             a3a22bcf-6e61-4135-a017-3c797fd876f7  default
UN  10.40.5.3   90.36 KiB  1            79.4%             78d82669-2cf9-4a7e-841d-4bc9ffaebbd0  default
UN  10.40.6.3   30.4 KiB   1            83.7%             718f3945-0bad-4c81-bd0c-06680daeec62  default
UN  10.40.7.3   71.07 KiB  1            78.3%             324c3fc5-fe3d-4c3d-929a-9409c8370eee  default
UN  10.40.8.3   90.23 KiB  1            36.6%             c373a0fa-f208-4b80-bc03-fcbb41a89ab9  default
UN  10.40.4.5   71.07 KiB  1            73.0%             edc40ba7-d48d-4170-b60a-68995cf4f2ca  default
UN  10.40.1.21  90.36 KiB  1            31.5%             407f920f-89ad-4e69-8274-6cacaf6baf6a  default
```

Token range ownership is way out of balance which can cause major issues as the load in the cluster increases.
 
We need a more sensible default for `num_tokens` for the 1.0 release.

@adejanovski you have done a good bit of analysis in this area. What are your thoughts? Keep in mind we are looking for a better default for 3.11.x in particular at the moment and also that the even token distribution algorithm is not an option (at least not in the 1.0 time frame).

burmanm:
**What this PR does**:
Overrides the num_tokens in cass-operator by setting it to a default value as defined in the used version. This would be 256 for 3.11.* and 16 for 4.0*. Also allows overriding to set a defined value.

Fixes #324 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/charts/k8ssandra/values.yaml
+++ b/charts/k8ssandra/values.yaml
@@ -98,7 +98,7 @@ cassandra:
     size: 1
 
     # -- Number of tokens within the datacenter. If not defined, the default values will be
-    # used. 
+    # used, which are 256 tokens for 3.11 releases and 16 tokens for 4.0 releases.
     # num_tokens: 16
 
     # -- Specifies the racks for the data center, if unset the datacenter will
--- a/charts/k8ssandra/values.yaml
+++ b/charts/k8ssandra/values.yaml
@@ -97,6 +97,10 @@ cassandra:
     # environments.
     size: 1
 
+    # -- Number of tokens within the datacenter. If not defined, the default values will be
+    # used. 
+    # num_tokens: 16
+
     # -- Specifies the racks for the data center, if unset the datacenter will
     # be composed of a single rack named `default`. The number of racks should
     # equal the replication factor of your application keyspaces. Cassandra will
--- a/charts/k8ssandra/Chart.yaml
+++ b/charts/k8ssandra/Chart.yaml
@@ -3,7 +3,7 @@ name: k8ssandra
 description: |
   Provisions and configures an instance of the entire K8ssandra stack. This includes Apache Cassandra, Stargate, Reaper, Medusa, Prometheus, and Grafana.
 type: application
-version: 0.50.0
+version: 0.51.0
 appVersion: 3.11.10
 
 dependencies:
--- a/charts/k8ssandra/templates/_helpers.tpl
+++ b/charts/k8ssandra/templates/_helpers.tpl
@@ -87,3 +87,19 @@ Create the jvm options based on heap properties specified.
   {{- end }}
 {{- end }}
 {{- end }}
+
+{{/*
+Set default num_tokens based on the server version
+*/}}
+{{- define "k8ssandra.default_num_tokens" -}}
+{{- $datacenter := (index .Values.cassandra.datacenters 0) -}}
+    {{- if $datacenter.num_tokens }}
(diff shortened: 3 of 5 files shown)


charts/k8ssandra/values.yaml
```diff
@@ -97,6 +97,10 @@ cassandra:
     # environments.
     size: 1
 
+    # -- Number of tokens within the datacenter. If not defined, the default values will be
```
jsanda, line 100: I would be explicit about the defaults. Go ahead and mention that for 3.11 the default is 256 and for 4.0 the default is 16.

charts/k8ssandra/values.yaml
```diff
@@ -97,6 +97,10 @@ cassandra:
...
+    # used. 
+    # num_tokens: 16
```
jsanda, line 102: I am a bit torn on the property name. Thinking out loud here...

On the one hand we are for the most part using camel case, so to be consistent it ought to be `numTokens`. On the other hand the fact that the property name is the same as it is in cassandra.yaml makes it crystal clear what the property is. 

I say leave it as `num_tokens`. If we are adding a property that maps directly to a property in cassandra.yaml, then using the same name adds clarity.

jsanda (changes requested):

jsanda (approved):
Thanks for knocking this out!

PR #409, opened 2021-02-22
add support for configOverride property

### LINKED ISSUE #387: Provide a way to apply advanced configuration settings for Cassandra
k8ssandra currently exposes a limited set of Cassandra's configuration options from cassandra.yaml and from jvm.options. This is by design, and we will introduce support for more settings over time. There may be situations where users need to apply settings that are not exposed. 

Let's say I have run into a Cassandra bug that is causing major instability in my cluster. The work around involves applying some change in cassandra.yaml, but the property in question is not exposed. What are my options? I can directly edit the CassandraDatacenter to apply the change. This will work, but the change will be lost when I run a `helm upgrade`. That just compounds the problem.

k8ssandra should provide some "hooks" to allow users to define arbitrary properties for cassandra.yaml and jvm.options. Since the configuration files are provisioned by cass-config-builder, it will be up to the user to understand what is supported and allowed by cass-config-builder. 

**Additional context**
A big motivation for this is customer support scenarios. We don't want this to turn into a backdoor that users rely on for all of their configuration. If that happens, then k8ssandra is falling short.

jsanda:
**What this PR does**:

Fixes #387 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/charts/k8ssandra/Chart.yaml
+++ b/charts/k8ssandra/Chart.yaml
@@ -3,7 +3,7 @@ name: k8ssandra
 description: |
   Provisions and configures an instance of the entire K8ssandra stack. This includes Apache Cassandra, Stargate, Reaper, Medusa, Prometheus, and Grafana.
 type: application
-version: 0.56.0
+version: 0.57.0
 appVersion: 3.11.10
 
 dependencies:
--- a/charts/k8ssandra/templates/cassandra/cassdc.yaml
+++ b/charts/k8ssandra/templates/cassandra/cassdc.yaml
@@ -81,15 +81,23 @@ spec:
   {{- end }}
 {{- end }}
   config:
+  {{- /*
+    configOverride is an undocumented property that is intended for advanced use cases like
+    production support. General use is discouraged If there is a particular Cassandra or
+    JVM setting you need exposed, please open an issue for it.
+   */ -}}
+  {{- if .Values.cassandra.configOverride }}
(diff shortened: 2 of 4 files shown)


charts/k8ssandra/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: k8ssandra
 description: |
   Provisions and configures an instance of the entire K8ssandra stack. This includes Apache Cassandra, Stargate, Reaper, Medusa, Prometheus, and Grafana.
 type: application
-version: 0.55.0
+version: 0.56.0
```
jakerobb, line 6: You'll need to bump to 57.

jakerobb (approved):
Looks good to me.

charts/k8ssandra/Chart.yaml
jdonenine: Yeah we're definitely at 56 already, strange that it doesn't seem to show that here, hmm.

https://github.com/k8ssandra/k8ssandra/blob/df1f5f44925b247a483014176995d4d691ada417/charts/k8ssandra/Chart.yaml#L6

jdonenine (changes requested):
Just need to bump the version and we're good.

emerkle826 (approved):

burmanm (approved):

jdonenine (approved):

PR #414, opened 2021-02-23
do not generate new passwords if the secrets already exist

### LINKED ISSUE #413: Running helm upgrade when auth is enabled can cause a cluster rolling restart
## Bug Report

**Describe the bug**
I observed that scaling my Cassandra cluster up or down via `helm upgrade` when auth and Reaper are enabled (which they are by default) causes a cluster-wide rolling restart after the scaling operation. More specifically, there is a StatefulSet update which has the effect of a rolling restart. I also observed the StatefulSet update doing a `helm upgrade` that made no changes to the CassandraDatacenter.

https://github.com/helm/charts/issues/5167 describes the root of the problem. We auto generate random passwords for secrets. cass-operator watches the secrets for the roles that it creates. When `helm upgrade` runs, the password in the secret changes. This in turns triggers a reconciliation request in cass-operator. I think cass-operator detects a change in the StatefulSet due to this:

```yaml
        env:
        - name: REAPER_JMX_USERNAME
          valueFrom:
            secretKeyRef:
              key: username
              name: test-reaper-secret-k8ssandra
        - name: REAPER_JMX_PASSWORD
          valueFrom:
            secretKeyRef:
              key: password
              name: test-reaper-secret-k8ssandra
```

This environment variable is from the podTemplateSpec of the StatefulSet that cass-operator creates. cass-operator add the `cassandra.datastax.com/resource-hash` annotation on the StatefulSet. During reconciliation the operator recreates the StatefulSet based on the current spec. It compares the new hash to the existing hash to determine if anything has changed. If the hashes are different cass-operator updates the StatefulSet

I confirmed the StatefulSet update and hash update simply by enabling Stargate.

**To Reproduce**
1. `helm install test k8ssandra/k8ssandra -f values.yaml`

**values.yaml:**
```yaml
cassandra:
  auth:
    enabled: true
  datacenters:
  - name: dc1
    size: 1
reaper:
  enabled: true
stargate:
  enabled: false
  replicas: 1
```

2. Wait for the CassandraDatacenter to become ready
3. Make a note of the StatefulSet's resource hash 

```
$ kubectl get sts test-dc1-default-sts -o=jsonpath="{.metadata.annotations.cassandra\.datastax\.com\/resource\-hash}"
```

4. Enable Stargate and upgrade
5. `helm upgrade test /k8ssandra/k8ssandra -f values.yaml`

**values.yaml:**
```yaml
cassandra:
  auth:
    enabled: true
  datacenters:
  - name: dc1
    size: 1
reaper:
  enabled: true
stargate:
  enabled: true
  replicas: 1
```

6. Wait a minute or two and repeat step 3.

Note that this happens frequently but not consistently. You may to repeat multiple times to reproduce.

**Expected behavior**
Scaling the cluster should not cause a cluster-wide rolling restart. Apply a `helm upgrade` that does not update the CassandraDatacenter should not cause a StatefulSet update.

* Helm charts version info

`$ helm ls -A` 

0.56.0

jsanda:
**What this PR does**:

Fixes #413

**Checklist**
- [x] Changes manually tested


jsanda:
When reviewing and testing changes, check the passwords of the secrets that get created. Make sure that they do not change after running `helm upgrade`.

charts/k8ssandra/templates/reaper/reaper-jmx-secret.yaml
```diff
@@ -14,6 +22,6 @@ data:
   {{- if .Values.reaper.jmx.password }}
   password: {{ .Values.reaper.jmx.password | b64enc | quote }}
   {{- else }}
-  password: {{ include "k8ssandra-common.password" . }}
+  password: {{ $password }}
```
burmanm, line 25: So this tackles the password, but is the username guaranteed to not be regenerated?

jsanda: Nice catch :) I missed that. I will update that now.

jdonenine (approved):
I was able to test and confirm that the secrets remained unchanged between `helm upgrade` attempts.

I'm not sure if @burmanm was still reviewing or not, but I'll go ahead and approve so that you can move forward if you already talked to him @jsanda before he went offline for the night over there.

charts/k8ssandra/templates/cassandra/superuser-secret.yaml
```diff
@@ -1,11 +1,19 @@
+{{- $password := (include "k8ssandra-common.password" .) }}
+{{- $secret := (lookup "v1" "Secret" .Release.Namespace (include "k8ssandra.superuserSecretName" .)) }}
+
 {{- if and .Values.cassandra.auth.superuser.username (not .Values.cassandra.auth.superuser.secret) .Values.cassandra.auth.enabled }}
+
+{{- if $secret }}
+{{- $password = index $secret.data "password" }}
```
burmanm, line 7: This question applies to this, medusa, reaper and stargate.. if I had a username in the Values-file, but changed it - won't my secrets be all messed up in the upgrade process? Currently, only the existing password is fetched in these, but not the existing username. So instead of existing username, we use the one from the values file, but old password. I'll be honest, I have no idea how messed up we could make the system if secret exists, but these are targetting a different username.

jsanda: That is a really good point. I was focused solely on the password. I will update the other secrets so that they reuse the existing username as well.

jsanda: For a first test, I installed with this configuration:

```yaml
cassandra:
  auth:
    enabled: true
  datacenters:
  - name: dc1
    size: 1
reaper:
  enabled: true
stargate:
  enabled: true
  replicas: 1
reaper:
  enabled: true
  cassandraUser:
```
That results in the creation of a secret named `test-reaper` and a Cassandra role/user named `reaper` with password `U59R30EzNxcNUjqi1koK`. I then changed my configuration to:

```yaml
cassandra:
  auth:
    enabled: true
  datacenters:
  - name: dc1
    size: 1
reaper:
  enabled: true
stargate:
  enabled: true
  replicas: 1
reaper:
  enabled: true
  cassandraUser:
    username: reaper-admin
```

After upgrading the username in the secret was changed to `reaper-admin`. The password in the secret did not. A new role was created, `reaper-admin`, same password as for `reaper` role.

I think a really good question is what if anything should happen to the `reaper` role. Should it be deleted in this case? Even if the answer is yes, it would be a post-1.0 change and probably require changes in cass-operator.

There is an additional change that I think should be made. When I change the username, I think a different password should be generated.

jdonenine: I think I agree on all fronts there.  I would generally expect that if the username is changed, that should effectively create a new user, not reuse the password from the older account.  I also think the cleanup of the left-over role can wait until post-1.0.0.

jsanda: https://github.com/datastax/cass-operator/issues/382 is related and something with which we need to be involved.

burmanm: I'm fine with postponing these to post-1.0

jdonenine (approved):
@jsanda and I have been working together through each series of changes and I just wrapped up some end-to-end testing and things look good.  Approving and we'll let @burmanm have another pass over things when he comes online and we should be able to merge!

PR #436, opened 2021-02-25
added lifecycle preStop to Stargate container

jakerobb:
**What this PR does**:
Adds a twenty-second (configurable) `preStop` delay to the Stargate container, giving it time to finish processing in-flight requests after Kube stops routing requests to it and before Kube sends SIGTERM.

I've also tossed in:
* an upgrade to Stargate ~1.0.11~ 1.0.13.
* removal of a docs reference that referred to `stargate.clusterVersion`, which is purposefully undocumented.
* fixed a docs issue related to num_tokens

**Checklist**
- [x] Changes manually tested
- [x] Changelog updated


charts/k8ssandra/templates/stargate/stargate.yaml
```diff
@@ -48,6 +48,10 @@ spec:
         - name: {{ .Release.Name }}-{{ include "k8ssandra.datacenterName" . }}-stargate
           image: {{ $image }}
           imagePullPolicy: {{ .Values.stargate.imagePullPolicy }}
+          lifecycle:
+            preStop:
+              exec:
+                command: [ "sleep", "20" ]
```
jakerobb, line 54: It might be good to make `20` the default value for an entry in values.yaml, perhaps something like `stargate.preStop.delay`.

jakerobb:
Note: Docs will need to be regenerated for this PR.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ When cutting a new release of the parent `k8ssandra` chart update the `main / un
 
 ## main / unreleased
 * [BUGFIX] #396 cqlsh commands show warnings
+* [ENHANCEMENT] #436 Upgrade Stargate to 1.0.13. Added `stargate.preStopDelaySecconds` with default of 20, which improves Stargate's availability during scale-down operations.
```
bradfordcp, line 16: Enhancements should go before bugfixes

```suggestion
* [ENHANCEMENT] #436 Upgrade Stargate to 1.0.13. Added `stargate.preStopDelaySecconds` with default of 20, which improves Stargate's availability during scale-down operations. 
* [BUGFIX] #396 cqlsh commands show warnings
```

bradfordcp (commented):
These changes look good. Minor fix to the changelog order.

jakerobb:
@bradfordcp I changed the changelog order -- just committed your suggestion as-is.

charts/k8ssandra/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: k8ssandra
 description: |
   Provisions and configures an instance of the entire K8ssandra stack. This includes Apache Cassandra, Stargate, Reaper, Medusa, Prometheus, and Grafana.
 type: application
-version: 1.0.1
+version: 1.0.3
```
jakerobb:
```suggestion
version: 1.0.5
```

jakerobb:
This will need another version bump in Chart.yaml before merging.

jakerobb:
@bradfordcp @jsanda I'd like to close this out. Are we waiting on anything for an approving review? (I'll fix the merge conflict and bump the version, but I'd rather do that post-approval to ensure I only need to do it once.)

bradfordcp:
No blockers from me. I defer to @jsanda for the approval.

jsanda (commented):
@jakerobb my apologies for being late to the game in reviewing this. Can you provide some context on why we want or need this?

jsanda:
I just read https://learnk8s.io/graceful-shutdown which is great article and provides the context I was looking for. I have some questions though:

* Can we run `nodetool drain` on Stargate nodes?
* Have you tested the preStop hook to get a feel for whether 20 sec is a good default?
* Pods have a default value of 30 sec for `terminationGracePeriodSeconds`. Would this increase the termination time to 50 secs?
* Could we simply use a higher value for `terminationGracePeriodSeconds`?

jdonenine:
What do we went to do with this PR at this point?  @jakerobb if you want to chime back in with answers to some of the later questions we can look at merging this in our next sprint, otherwise perhaps we just close this if we're not sure about some of those questions?

jakerobb:
Hey guys, sorry, wasn't checking my GH notifications there for a bit.

> Can we run nodetool drain on Stargate nodes?

I have genuinely no idea. 

> Have you tested the preStop hook to get a feel for whether 20 sec is a good default?

Yes; I used [stargate-load-test.sh](https://github.com/jakerobb/k8ssandra-scripts/blob/master/run/stargate-load-test.sh) to apply a continuous load, then reduced the number of Stargate replicas. Before this change, several requests would fail while nodes were shutting down. After, none fail. It's possible the default could be turned down; I didn't do a lot of testing with different values, as the load test script is pretty immature and doesn't send any complex requests that might need more time for a response, nor does it use a complex data model.

> Pods have a default value of 30 sec for terminationGracePeriodSeconds. Would this increase the termination time to 50 secs?

30 seconds is the maximum, used only if the container does not exit 30 seconds after SIGTERM is sent. My change delays SIGTERM by 20 seconds. Stargate tends to shut down in <10 seconds after SIGTERM. So worst-case termination time does increase, but actual termination time is not generally anywhere near 50 seconds.

> Could we simply use a higher value for terminationGracePeriodSeconds?

No. Here's the timeline:
1. Kubernetes decides to shut down a Stargate instance (e.g. because the deployment is being scaled down)
2. Asynchronous things happen that lead to both the preStop hook being invoked and the to-be-shut-down pod being removed from the Service.
3. Kubernetes stops routing requests to the pod via the Service.
4. When the preStop hook exits, SIGTERM is sent to the container's PID 1. (If there were no preStop hook, this would happen immediately in step 2). terminationGracePeriodSeconds timer starts.
5. Stargate JVM reacts to SIGTERM and begins to shut down. Stargate does not have any custom handling for the SIGTERM, and so it does not shut down gracefully. Any in-flight requests will be aborted and incoming connections closed unceremoniously.
6. Stargate JVM exits. Normally, terminationGracePeriodSeconds timer is nowhere near 30 seconds.
7. Container exits; pod goes away.

What preStop does is give Kubernetes a chance to stop routing traffic to the pod, and for any requests already sent to the pod to finish processing and return a response, _before_ we send SIGTERM to the JVM. As noted in #5, SIGTERM leads to the failure of any in-flight requests.

jakerobb:
Also: it seems like the cloud team intends to use 20 seconds when they get around to implementing this: https://datastax.jira.com/browse/C2-940

jsanda:
Closing due to inactivity. Please reopen if you want to revisit this.

PR #444, opened 2021-02-25
update to v1.6.0

jsanda:
**What this PR does**:
Upgrade to cass-operator 1.6.0

Fixes #<issue number>

**Checklist**
- [x] Changes manually tested


jdonenine (approved):
I finished up a full test plan run on 1.18 + 4.0.0 - looks good!  Was able to test size changes, backup, restore all as expected.

I will try to test some more version combos tonight, but I'm content here.

The only thing we need to add to this is the k8ssandra chart version, which wasn't updated here.  We can do that after the fact if you want @jsanda after we do some more testing and go straight to 1.0.0 if we're good later tonight.

jdonenine (approved):
1.0.0 🎉

PR #456, opened 2021-02-26
FAQs on multiple installs, Docker Desktop for Mac tip, cass-operator …

johnsmartco:
…1.6.0 version edit

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/docs/content/en/docs/faqs/_index.md
+++ b/docs/content/en/docs/faqs/_index.md
@@ -87,7 +87,25 @@ helm install demo k8ssandra/k8ssandra -n k8ssandra --create-namespace
 
 ### Can I install multiple releases of K8ssandra?
 
-Some of the objects installed by the K8ssandra chart are currently configured to be cluster-scoped; consequently, you should only install those components once. This should be fixed before version 1.0 to allow multiple installations. Other parts can be installed multiple times to allow creating multiple Cassandra clusters in a single k8s cluster.
+You can install multiple releases of K8ssandra in a Kubernetes environment, provided:
+
+* You wait for all the pods to reach a Ready or Completed state for a given K8ssandra install, before attempting the next K8ssandra install
+* **And** provided you use a unique cluster-name per install
+
+### How do I install K8ssandra using the K8 included with Docker Desktop for Mac?
+
+When installing K8ssandra in the K8 instance of Docker Desktop for Mac, you may encounter the following error:
+
+```bash
+Error: failed pre-install: timed out waiting for the condition
+```
+
+To solve the issue, install K8ssandra using the following Helm command:
+
+```bash
+helm install <cluster-name> k8ssandra/k8ssandra \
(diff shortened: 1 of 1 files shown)


docs/content/en/docs/faqs/_index.md
```diff
@@ -87,7 +87,25 @@ helm install demo k8ssandra/k8ssandra -n k8ssandra --create-namespace
 
 ### Can I install multiple releases of K8ssandra?
 
-Some of the objects installed by the K8ssandra chart are currently configured to be cluster-scoped; consequently, you should only install those components once. This should be fixed before version 1.0 to allow multiple installations. Other parts can be installed multiple times to allow creating multiple Cassandra clusters in a single k8s cluster.
+You can install multiple releases of K8ssandra in a Kubernetes environment, provided:
+
+* You wait for all the pods to reach a Ready or Completed state for a given K8ssandra install, before attempting the next K8ssandra install
```
jakerobb, line 92: I didn't know this. What happens if you attempt to install another before then?

jakerobb (approved):
LGTM.

jdonenine (approved):
I'm approving as well so we can get the general content in - but I did have the same question as @jakerobb, what happens if you don't wait? I took a look back through the notes that @burmanm had put on the issue we had about multiple installs and didn't see that mentioned?

johnsmartco:
Right, thanks guys. Michael @burmanm said :  "As long as the deployment doesn’t happen instantly, then it works. Deploying two instances with && causes errors on the Prometheus end occasionally." 

https://datastax.slack.com/archives/C01B6E1NMCH/p1614068584129600?thread_ts=1614014663.060600&cid=C01B6E1NMCH

docs/content/en/docs/faqs/_index.md
slayerjain: I thought K8ssandra is like a super operator over cass-operator. I didn't know that for installing another Cassandra cluster, I'll have to install the whole stack again.

jakerobb: @slayerjain k8ssandra is, for now at least, not an operator, but a set of Helm charts. It greatly simplifies the process of configuring cass-operator by establishing sane defaults. It also helps you configure the other components you need in order to have a production-ready Cassandra database: backup/restore, repair, coordinator-only nodes for offloading compute work from storage nodes (aka Stargate), ingress, monitoring, and so on.

If you want multiple Cassandra clusters in your Kubernetes cluster, then yes, you need to manage a values.yaml (or, if you're a masochist, a set of `--set` arguments) and invoke `helm install` once for each.  

I'm not sure of the current status of this effort (post-1.0 planning is happening today), but the plan is that some components (e.g. monitoring) can be shared across k8ssandra clusters.

slayerjain: @jakerobb thanks for the detailed explanation.

PR #508, opened 2021-03-08
Make TESTS and GO_FLAGS conditional so they can be set from cmd line

### LINKED ISSUE #507: TESTS and GO_FLAGS Makefile vars cannot be set from command line
## Bug Report

**Describe the bug**
The `TESTS` and `GO_FLAGS` variables were introduced to give more control over unit test execution, but they currently cannot be set from the command line.

**To Reproduce**
1. Run `GO_FLAGS="-v" TESTS=".*auth.*" make unit-test`
2. All unit tests are run with non-verbose output

**Expected behavior**
I should be able to set these vars from the command line.

**Screenshots**

* Helm charts version info

1.0.0

`$ helm get values RELEASE_NAME` 

`kubectl version`

* Kubernetes cluster kind:

**Additional context**

jsanda:
**What this PR does**:
Makes the `TESTS` and `GO_FLAGS` Makefile vars conditional so that they can be set from the cmd line.

Fixes #507

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ When cutting a new release of the parent `k8ssandra` chart update the `main / un
 
 ## main / unreleased
 * [BUGFIX] #396 cqlsh commands show warnings
+* [BUGFIX] #507 `TESTS` and `GO_FLAGS` Makefile vars cannot be set from command line
 
 ## v1.0.0 - 2021-02-26
 
--- a/Makefile
+++ b/Makefile
@@ -13,8 +13,8 @@ CLEANER_LATEST_IMAGE=$(CLEANER_IMAGE_BASE):latest
 # Image URL to use all building/pushing image targets
 CLEANER_IMG ?= $(CLEANER_LATEST_IMAGE)
 
-TESTS=all
-GO_FLAGS=
+TESTS?=all
+GO_FLAGS?=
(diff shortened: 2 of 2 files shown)


burmanm:
So what's the behavioral change? They were always settable from the command line.

burmanm:
Actually I can't find any difference when testing with ``make GO_FLAGS="-v" TESTS="wontmatch" test`, it functions the same with both versions.

jsanda:
> So what's the behavioral change? They were always settable from the command line.

No they were not settable from the command line. I tested and verified this by adding this target:

```make
MSG=foo
echo-test:
	echo $(MSG)
```

Running `MSG=bar make echo-test` still outputs `foo`.

But when I change it to

```make
MSG?=foo
echo-test:
	echo $(MSG)
```
and then run `MSG=bar make echo-test` the output is `bar` as expected.

jsanda:
Ok, now I did the following:

```
$ make MSG=bar echo-test
echo bar
bar
```
@burmanm do you know why there is a difference in behavior when setting the var before the `make` command?

burmanm:
If it's after make command then it's a parameter for make (it could be -t etc also and we could catch that). The one before make is a shell variable.

jsanda:
Then I can go ahead and close this. Thanks for the explanation!

PR #511, opened 2021-03-08
Update examples and docs in values.yaml

### LINKED ISSUE #510: Update examples and docs in values.yaml
## Bug Report

**Describe the bug**
The comments for the `affinityLabels` property in values.yaml mention an example to demonstrate how to pin racks to nodes, but there is no such example.

There are some other things noted by @jeffreyscarpenter including:

* Lack of docs for `cleaner`
* Confusing docs for using Medusa local storage for backups
* Explanation of some `resources`
* Explanation of lack of default value for heap settings
* And more...

**To Reproduce**
1. Go to https://github.com/k8ssandra/k8ssandra/blob/v1.0.0/charts/k8ssandra/values.yaml#L182

**Expected behavior**
There should be an example in the comments.

**Screenshots**

* Helm charts version info

1.0.0

jsanda:
**What this PR does**:

Fixes #510

**Checklist**
- [ ] Changes manually tested


jsanda:
@burmanm and/or @jdonenine can you review this?

charts/k8ssandra/values.yaml
```diff
@@ -482,8 +500,11 @@ medusa:
   # -- Enables usage of a bucket across multiple clusters.
   multiTenant: false
 
-  # -- API interface used by the object store. Supported values include `s3`, 's3_compatible'
-  # and `gcs`. For pod mapped storage, use 'local'
+  # -- API interface used by the object store. Supported values include `s3`,
+  # 's3_compatible' and `gcs`. For file system storage, i.e., a pod volume
+  # mount, use 'local'. Note that 'local' does not necessarily imply a local
+  # volume. It could also be network attached storage. It is simply accessed
+  # through the local file system.
```
burmanm, line 507: Or accessed through pod's mounts? "local file system" is a bit odd here as you just said it's not local necessarily.

jsanda: I will remove local.

charts/k8ssandra/values.yaml
```diff
@@ -511,17 +532,25 @@ medusa:
...
 
   prometheus:
     # -- Enables the creation of Prometheus Operator ServiceMonitor custom
-    # resources. If you are not using the kube prometheus stack sub-chart or do
+    # resources. If you are not using the kube-prometheus-stack subchart or do
     # not have the ServiceMonitor CRD installed on your cluster, set this value
     # to `false`.
     provision_service_monitors: true
 
+# -- The cleaner is a pre-delete hook that ensures the CassandraDatacenter
```
burmanm, line 548: I think after the CassandraBackup we could say it's a pre-delete hook that cleans up finalizers that will prevent the uninstallation. Not just cassdc.

jsanda: Not sure I understand. I reviewed the cleaner code and it just deletes the CassandraDatacenter that is installed with the release. What am I missing?

burmanm: The https://github.com/k8ssandra/medusa-operator/pull/26 will add a new finalizer, so the cleaner must be updated also to remove all these backups (or at least finalizers). Otherwise helm uninstall will no longer work. So it's going to be more common than just CassDC.

jsanda: How about something like:

 _The cleaner is a pre-delete hook that that ensures objects with finalizers get deleted. For example, cass-operator sets a finalizer on the CassandraDatacenter. Kubernetes blocks deletion of an object until all of its finalizers are cleared. In the case of the CassandraDatacenter object, cass-operator removes the finalizer. The problem is that there are no ordering guarantees with helm uninstall which means that the cass-operator deployment could be deleted before the CassandraDatacenter. The cleaner ensures that the CassandraDatacenter is deleted before cass-operator._

burmanm: Sounds good.

bradfordcp (approved):

bradfordcp:
Given the changes here we should run the script to regenerate the reference docs.

jsanda:
I have regenerated the docs. @bradfordcp are you good with my merging now?

PR #517, opened 2021-03-09
Removed prerelease automation from GH Actions

bradfordcp:
**What this PR does**: Removes prerelease automation from GH Actions until we figure out what this should look like.

**Checklist**
- [x] Changes manually tested


bradfordcp:
The latest run on main shows the automation is failing to push to Github Releases. It was pulling the version number from the "short version" which in this case is `main`. If we want this functionality in the future we either want to parse it from the Chart.yaml _or_ use a shortened commit hash. In either case, the behavior here is wrong and in the interim, this PR removes the functionality.

Issue #204, opened 2021-01-19
Write e2e test to verify full stack is deployed


jsanda:
I would like to see an e2e integration test that deploys the full stack. We  The test should verify that each of the components is initialized. To the extent possible this should be done through the status of each of the objects. 

The CassandraDatacenter object for example has the `cassandraOperatorProgress` field which is set to `true` when the operator is done doing work. There is also condition named `Ready` whose status will be set to `true` when all the Cassandra pods have reached the ready state.

The Reaper object simply has a `ready` field in its status that is set to true when Reaper reaches the ready state. 

We want perform similar verification through the statuses of Prometheus, Grafana, and Stargate as well. 

Before checking each of those objects, we should probably first check that each of the operators is successfully deployed. If cass-operator fails to deploy for example, there is not point to try and check the CassandraDatacenter.

For Prometheus, I think we additionally should verify that targets have been discovered. We have example of this already being done in [simple_integration_test.go](https://github.com/k8ssandra/k8ssandra/blob/main/tests/integration/simple_integration_test.go#L126).

For Grafana if possible I think we should verify that the Prometheus data source is registered and also check that the dashboards are imported.

We have people developing on both Mac and on Linux. The test should be able to run on both platforms. Furthermore, the test should not only be able to run on [kind](https://kind.sigs.k8s.io/) but other distros as well, e.g., k3d, GKE, etc. I am not saying we need to run the test against N different distros, but should be a general enough test that it should run on any (or most) distros provided it meets the minimim Kubernetes version requirements. 

[simple_integration_test.go](https://github.com/k8ssandra/k8ssandra/blob/main/tests/integration/simple_integration_test.go#L126) deploys Traefik and verifies ingress as well. If using Traefik will make it a lot more difficult to target different k8s distros, then I think we should consider tests with and without ingress.

jsanda:
`simple_integration_test.go` sets up the cluster and configures traefik as described in the docs [here](https://k8ssandra.io/docs/topics/ingress/traefik/kind-deployment/). 

Here is the traefik chart configuration:

```yaml
providers:
  kubernetesCRD:
    namespaces:
      - default
      - traefik
  kubernetesIngress:
    namespaces:
      - default
      - traefik

ports:
  traefik:
    expose: true
    nodePort: 32090
  web:
    nodePort: 32080
  websecure:
    nodePort: 32443
  cassandra:
    port: 9042
    nodePort: 32091
  cassandrasecure:
    port: 9142
[...]

PR #544, opened 2021-03-16
Implement integration tests with the testing package

### LINKED ISSUE #104: Research feasibility of using GitHub Actions for executing integration testing

adejanovski:
Use a Matrix to test different features and Cassandra versions in parallel.
Test each component in the stack:
- Reaper
- Medusa
- Monitoring (Prometheus, Grafana, MCAC)
- Stargate (Auth and Document APIs)

**What this PR does**:
Adds a first set of integration tests for each component in the stack and introduces a GHA workflow for running them on push/pull requests.

Fixes https://github.com/k8ssandra/k8ssandra/issues/204

**Checklist**
- [x] Changes manually tested


.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,93 @@
...
+      - name: Install Dependencies
+        run: |
+          set -x
+          sudo apt-get update || echo "something went slightly wrong with apt-get update"
+
+          # install Go
+          wget https://dl.google.com/go/go1.15.8.linux-amd64.tar.gz
+          sudo tar -C /usr/local -xzf go1.15.8.linux-amd64.tar.gz
+          export PATH=$PATH:/usr/local/go/bin
+          go version
```
emerkle826, line 31: Have you tried to setup Go using the GitHub Action like they do in cass-operator?
https://github.com/datastax/cass-operator/blob/master/.github/workflows/kindIntegTest.yml#L32-L35

Either way works for me, but I guess the Action is _cleaner_

adejanovski: Good point! I didn't think of searching for an action for this.
I'll update to use the action, it's cleaner indeed 👍

tests/integration/steps/integration_steps.go
```diff
@@ -0,0 +1,622 @@
...
+func MinioServiceName(t *testing.T) string {
+	kubectlOptions := k8s.NewKubectlOptions("", "", "minio")
+	minioService, err := k8s.RunKubectlAndGetOutputE(t, kubectlOptions, "get", "services", "-l", "app=minio", "-o", "jsonpath='{.items[0].metadata.name}'")
+	g(t).Expect(err).To(gomega.BeNil())
+
+	log.Printf("Minio service: %s", minioService)
+	return minioService
+}
+
+func CheckResourceIsPresent(t *testing.T, resourceType, label string) {
```
jeffbanks, line 199: Should we consider adding `WithLabel` to the name of the externally accessible function to indicate that used as a filter?

tests/integration/steps/integration_steps.go
```diff
@@ -0,0 +1,622 @@
...
+
+	customChartPath, err := filepath.Abs("charts/" + customValues)
+	g(t).Expect(err).To(gomega.BeNil())
+
+	helmOptions := &helm.Options{}
+
+	helmOptions = &helm.Options{
+		// Enable traefik to allow redirections for testing
+		SetValues:      helmValues,
+		KubectlOptions: k8s.NewKubectlOptions("", "", namespace),
```
jeffbanks, line 93: A couple of observations, possible for a future PR, but nonetheless.

1) Do we need to start thinking about a test `context` naming here for our integration tests w/ KubectlOptions?  I'm thinking about things like having concurrently running jobs that have the same namespace, issues with using an empty string, etc.

2) In a prior effort, I was building out a set of test structs/models for managing some of the redundancy in test setup and configuration.

In this example, I was allowing the test setup to prepare these values.  This also included an optional set of string arguments when needed to apply or create with overrides.  

```golang
type OptionsContext struct {
	HelmOptions *helm.Options
	KubeOptions *k8s.KubectlOptions
	Namespace   string
	ReleaseName string
	Args        []string
}
```
Since this is Go, I suspect we could still consider some of this for adoption.  

IMO, structs allows us to reduce some verbosity and build things more consistently.  In fact, [generics in Golang](https://blog.golang.org/generics-next-step) may be a thing in the near future, so that would blend well with this approach moving forward.  

Just a thought, not show stopper.

jeffbanks (approved):

tests/integration/integration_suite_test.go
```diff
@@ -1,13 +1,187 @@
...
+// - Install Traefik
+// - Create a namespace
+// - Register a cluster with 3 Cassandra nodes
+// - Verify that Reaper correctly initializes
+// - Start a repair on the reaper_db keyspace
+// - Wait for at least one segment to be processed
+// - Cancel the repair
+// - Terminate the namespace and delete the cluster
+func TestReaperDeploymentScenario(t *testing.T) {
+	t.Run("Initialize cluster", initializeCluster)
```
jsanda, line 49: I am not sure if makes sense to organize all of these as subtests. For instance, if `initializeCluster` fails, every test that follows will fail, so there is no point in running them in that case.

jsanda: GH wouldn't let me update my previous comment. It seems like it would make more sense to treat `initializeCluster` and `deployClusterForReaper` as setup. If either of those fail, there really is not much point in proceeding.

tests/integration/integration_suite_test.go
```diff
@@ -1,13 +1,187 @@
...
+// - Load 10 rows and check that we can read 10 rows
+// - Perform a backup with Medusa
+// - Load 10 rows and check that we can read 20 rows
+// - Restore the backup
+// - Verify that we can read 10 rows
+// - Terminate the namespace and delete the cluster
+func TestMedusaDeploymentScenario(t *testing.T) {
+	backends := []string{"Minio", "S3"}
+	for _, backend := range backends {
+		t.Run(fmt.Sprintf("%s: Initialize cluster", backend), initializeCluster)
```
jsanda, line 102: Same as above, if `initializeCluster` fails, there is no point in proceeding.

tests/integration/integration_suite_test.go
```diff
@@ -1,13 +1,187 @@
...
+			} else {
+				// S3
+				CreateMedusaSecretWithFile(t, "~/medusa_secret.yaml")
+			}
+		})
+
+		t.Run(fmt.Sprintf("%s: Deploy cluster", backend), func(t *testing.T) {
+			if backend == "Minio" {
+				// Minio
+				DeployClusterWithValues(t, "minio", "one_node_cluster_with_medusa_minio.yaml")
```
jsanda, line 117: Same as above, if `DeployClusterWithValues` fail, I don't see why we would bother with running subsequent subtests.

tests/integration/integration_suite_test.go
```diff
@@ -1,13 +1,187 @@
...
+			} else {
+				// S3
+				DeployClusterWithValues(t, "default", "one_node_cluster_with_medusa_s3.yaml")
+			}
+		})
+		t.Run(fmt.Sprintf("%s: Check expected resources are present", backend), CheckClusterExpectedResources)
+
+		t.Run(fmt.Sprintf("%s: Create a test table", backend), func(t *testing.T) {
+			CreateCassandraTable(t, medusaTestTable, medusaTestKeyspace)
```
jsanda, line 126: As with `TestReaperDeploymentScenario` I am not sure it makes sense to organize all of these as subtests. If `CreateCassandraTable` fails, then presumably all of the following tests that depend on that table will fail as well. You could conditionally execute the following tests only if `CreateCassandraTable` succeeds, but I introduces unnecessary complexity. If they are combined into a single test and `CreateCassandraTable` fails, then the test halts execution and no need for conditional logic.

tests/integration/steps/integration_steps.go
```diff
@@ -0,0 +1,622 @@
...
+		return podWithLabelIsReady(t, "app.kubernetes.io/name=cass-operator")
+	}, RETRY_TIMEOUT, RETRY_INTERVAL).Should(gomega.BeTrue())
+
+	// Wait for CassandraDatacenter to be ready..
+	WaitForCassDcToBeReady(t)
+}
+
+func WaitForCassDcToBeReady(t *testing.T) {
+	kubectlOptions := k8s.NewKubectlOptions("", "", namespace)
+	k8s.RunKubectl(t, kubectlOptions, "wait", "--for=condition=Ready", "cassandradatacenter/dc1", fmt.Sprintf("--timeout=%s", 15*time.Minute))
```
jsanda, line 112: As I mentioned in the other PR, I would prefer using `wait` or `Eventually` here so that we can log some status. 15 minutes (in the worst case scenario) is a long time to wait without any updates.

tests/integration/steps/integration_steps.go
```diff
@@ -0,0 +1,622 @@
...
+	g(t).Eventually(func() int {
+		return backupIsFinished(t, backupName)
+	}, 2*time.Minute, 5*time.Second).Should(gomega.Equal(1))
+}
+
+func backupIsFinished(t *testing.T, backupName string) int {
+	var nodes []string
+	log.Printf("Checking if backup %s is finished...", backupName)
+	kubectlOptions := k8s.NewKubectlOptions("", "", namespace)
+	output, err := k8s.RunKubectlAndGetOutputE(t, kubectlOptions, "get", "cassandrabackup", backupName, "-o", "jsonpath='{.status.finished}'")
```
jsanda, line 424: This check might not always be right. `.status.finished` gets updated as each backup completes. The `.status.finishTime` is set once all of the backups are done.

tests/integration/steps/integration_steps.go
```diff
@@ -0,0 +1,622 @@
...
+		},
+		KubectlOptions: k8s.NewKubectlOptions("", "", namespace),
+	}
+	helm.Install(t, helmOptions, restoreChartPath, "restore-test")
+
+	// Restore should trigger an update of the pod through cass-operator
+	g(t).Eventually(func() string {
+		return readCassandraOperatorProgress(t)
+	}, RETRY_TIMEOUT, 5*time.Second).Should(gomega.Equal("Updating"))
+
```
jsanda, line 449: Similar with backups, the CassandraRestore object has a `.status.finishTime` property that is set after the restore is complete. This happens after the CassandraDatacenter comes back to the ready state.

tests/integration/steps/integration_steps.go
```diff
@@ -0,0 +1,622 @@
...
+	g(t).Eventually(func() bool {
+		return clusterIsRegisteredInReaper(t, clusterName)
+	}, RETRY_TIMEOUT, RETRY_INTERVAL).Should(gomega.BeTrue(), "Cluster wasn't properly registered in Reaper")
+
+	t.Log("Cluster wasn't properly registered in Reaper")
+	t.FailNow()
+}
+
+func clusterIsRegisteredInReaper(t *testing.T, clusterName string) bool {
+	restClient := resty.New()
```
jsanda, line 476: I think @burmanm already mentioned https://github.com/k8ssandra/reaper-client-go. It is definitely complete but it is used in reaper-operator and would be nice to reuse it here.

jsanda (changes requested):

tests/integration/steps/integration_steps.go
```diff
@@ -0,0 +1,622 @@
...
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const RETRY_TIMEOUT = 15 * time.Minute
+const RETRY_INTERVAL = 30 * time.Second
+
+var namespace string
+var repairId string
+var releaseName string
```
burmanm, line 31: We should not have any global variables if we want to allow proper parallel execution. These should be hidden behind a struct.

tests/integration/steps/integration_steps.go
```diff
@@ -0,0 +1,622 @@
...
+	return false
+}
+
+func g(t *testing.T) *gomega.WithT {
+	return gomega.NewWithT(t)
+}
+
+func deployCluster(t *testing.T, customValues string, helmValues map[string]string) {
+	clusterChartPath, err := filepath.Abs("../../charts/k8ssandra")
+	g(t).Expect(err).To(gomega.BeNil())
```
burmanm, line 83: I guess you want to have ``.`` in the import segment for the gomega. This way the ``gomega.BeNil()`` can be written as ``BeNil()``, which makes sense for a matcher library (kinda like Java's static import)

tests/integration/steps/integration_steps.go
```diff
@@ -0,0 +1,622 @@
...
+}
+
+func GetStargateService(t *testing.T) v1.Service {
+	kubectlOptions := k8s.NewKubectlOptions("", "", namespace)
+	clientset, err := k8s.GetKubernetesClientFromOptionsE(t, kubectlOptions)
+	g(t).Expect(err).To(gomega.BeNil())
+
+	services, err := clientset.CoreV1().Services(namespace).List(context.Background(), metav1.ListOptions{})
+	for _, service := range services.Items {
+		if strings.HasSuffix(service.ObjectMeta.Name, "-stargate-service") {
```
burmanm, line 180: I think this should verify the annotation / labels instead of name.

tests/integration/steps/integration_steps.go
```diff
@@ -0,0 +1,622 @@
...
+			return service
+		}
+	}
+
+	t.Log("Couldn't find Stargate service")
+	t.FailNow()
```
burmanm, line 186: These helper methods shouldn't probably fail as they now don't allow reusing the same method to verify that Stargate service was NOT deployed. I'd say keep the matchers and assertions in the test methods and all these helpers free of them. Also, splitting these helpers to another file would make reading probably easier.

tests/integration/steps/integration_steps.go
```diff
@@ -0,0 +1,622 @@
...
+func namespaceIsAbsentOrTerminating(t *testing.T, ns string) bool {
+	kubectlOptions := k8s.NewKubectlOptions("", "", "default")
+	namespaceObject, _ := k8s.GetNamespaceE(t, kubectlOptions, namespace)
+	if namespaceObject == nil {
+		// namespace was deleted
+		return true
+	}
+
+	if namespaceObject.Status.Phase == v1.NamespaceTerminating {
+		// namespace is terminating, which is good enough
```
burmanm, line 303: Namespace deleted and terminating are not equal. Namespace can stay in terminating phase forever if there's finalizers blocking the deletion. That would be a bug in uninstall.

tests/integration/steps/integration_steps.go
```diff
@@ -0,0 +1,622 @@
...
+func UninstallTraefikHelmRelease(t *testing.T) {
+	err := RunShellCommand(exec.Command("helm", "uninstall", "traefik", "-n", "traefik"))
+	g(t).Expect(err).To(gomega.BeNil(), "Failed uninstalling Traefik Helm release")
+}
+
+func DeleteK8ssandraNamespace(t *testing.T) {
+	deleteNamespace(t, namespace)
+}
+
+func DeleteTraefikNamespace(t *testing.T) {
```
burmanm, line 330: To which packages are these functions meant to be exported to?

tests/integration/steps/integration_steps.go
```diff
@@ -0,0 +1,622 @@
...
+
+func LoadRowsInTable(t *testing.T, nbRows int, tableName, keyspaceName string) {
+	for i := 0; i < nbRows; i++ {
+		runCassandraQueryAndGetOutput(t, fmt.Sprintf("INSERT INTO %s.%s(id,val) values(now(), '%d');", keyspaceName, tableName, i))
+	}
+}
+
+// Medusa related functions
+func CreateMedusaSecretWithFile(t *testing.T, secretFile string) {
+	home, _ := os.UserHomeDir()
```
burmanm, line 394: I hope this isn't creating files to user's home directory who is running the tests? We don't want to pollute the user's filesystem with stuff.

tests/integration/steps/integration_steps.go
```diff
@@ -0,0 +1,622 @@
...
+	}, 2*time.Minute, 5*time.Second).Should(gomega.Equal(1))
+}
+
+func backupIsFinished(t *testing.T, backupName string) int {
+	var nodes []string
+	log.Printf("Checking if backup %s is finished...", backupName)
+	kubectlOptions := k8s.NewKubectlOptions("", "", namespace)
+	output, err := k8s.RunKubectlAndGetOutputE(t, kubectlOptions, "get", "cassandrabackup", backupName, "-o", "jsonpath='{.status.finished}'")
+	if err == nil && len(output) > 0 {
+		json.Unmarshal([]byte(strings.ReplaceAll(output, "'", "")), &nodes)
```
burmanm, line 426: Missing err check?

tests/integration/steps/integration_steps.go
```diff
@@ -0,0 +1,622 @@
...
+	helm.RunHelmCommandAndGetOutputE(t, helmOptions, "repo", "add", "minio", "https://helm.min.io/")
+
+	helm.RunHelmCommandAndGetOutputE(t, helmOptions, "install",
+		"--set", fmt.Sprintf("accessKey=minio_key,secretKey=minio_secret,defaultBucket.enabled=true,defaultBucket.name=%s", bucketName),
+		"--generate-name", "minio/minio", "-n", "minio", "--create-namespace")
+}
+
+func CheckClusterExpectedResources(t *testing.T) {
+	checkResourcePresence(t, "service", "k8ssandra-dc1-all-pods-service")
+	checkResourcePresence(t, "service", "k8ssandra-dc1-service")
```
burmanm, line 620: These names are only true if datacenter name can't be modified.

tests/integration/steps/integration_steps.go
```diff
@@ -0,0 +1,622 @@
...
+	sprint := func(args ...interface{}) string {
+		return fmt.Sprintf(colorString,
+			fmt.Sprint(args...))
+	}
+	return sprint
+}
+
+func timeTrack(start time.Time, name string) {
+	elapsed := time.Since(start)
+	log.Println(Info(fmt.Sprintf("%s took %s", name, elapsed)))
```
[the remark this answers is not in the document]
adejanovski: We are still using Terratest :) at least for now...
This `Info()` function allows to get colorized output for specific lines so they're easy to spot.

tests/integration/integration_suite_test.go
```diff
@@ -1,13 +1,187 @@
-package integration_test
+package integration
```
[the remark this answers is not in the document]
adejanovski: I think we still haven't decided if we were dealing with e2e or integration tests here 😅 
I'm checking a lot of intermediate steps and not with the all stack loaded each time, so maybe that doesn't qualify as e2e.

tests/integration/steps/integration_steps.go
adejanovski: Yes, that sounds right.

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,93 @@
...
+          curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+          sudo install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
+
+          # install kind
+          curl -Lo ./kind https://kind.sigs.k8s.io/dl/v0.10.0/kind-linux-amd64
+          chmod +x ./kind
+          sudo install -o root -g root -m 0755 ./kind /usr/local/bin/kind
+
+          # install Helm
+          curl https://baltocdn.com/helm/signing.asc | sudo apt-key add -
```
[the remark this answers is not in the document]
adejanovski: Sure thing, I'll use it instead.

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,93 @@
...
+      - name: Run Integration Tests
+        run: |
+          cd tests/integration
+
+          # Medusa on S3 requires a secret. Skip step if the secret is absent.
+          if [[ ( -n "${{ secrets.MEDUSA_SECRET }}" && "${{ matrix.scenario }}" == "Medusa on S3" ) \
+            || "${{ matrix.scenario }}" != "Medusa on S3" ]]
+          then
+            ./scripts/create_kind_cluster.sh
+            go test . -timeout 30m -v -failfast -run=${{ matrix.scenario }}
```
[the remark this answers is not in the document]
adejanovski: Agreed, I'll make the necessary changes.

tests/integration/integration_suite_test.go
adejanovski: They won't as we invoke the tests with the `-failfast` flag.
Organizing the steps in subtests gives us: 

- Readability for the test code
- Clean output with each subtest description being logged when it starts
- The ability to run single subtests when needed (for dev/troubleshooting purposes for example).

I'm -1 on not using subtests here since we can achieve the desired behavior and exit on failure.

Here's an example of a test failing in the "Initialize cluster" subtest:

```
% go test . -v -failfast -run=TestMedusaDeploymentScenario/"Minio:"
2021/03/17 08:38:28 Updating Helm dependencies...
=== RUN   TestMedusaDeploymentScenario
=== RUN   TestMedusaDeploymentScenario/Minio:_Initialize_cluster
TestMedusaDeploymentScenario/Minio:_Initialize_cluster 2021-03-17T08:38:34+01:00 logger.go:66: Running command kubectl with args [--namespace default get namespace default -o jsonpath={.metadata.name}]
TestMedusaDeploymentScenario/Minio:_Initialize_cluster 2021-03-17T08:38:34+01:00 logger.go:66: The connection to the server localhost:8080 was refused - did you specify the right host or port?
    TestMedusaDeploymentScenario/Minio:_Initialize_cluster: integration_steps.go:219:
        Expected
            <*shell.ErrWithCmdOutput | 0xc00027e040>: {
                Underlying: {
                    ProcessState: {
                        pid: 99393,
                        status: 256,
                        rusage: {
                            Utime: {
                                Sec: 0,
                                Usec: 40937,
                                Pad_cgo_0: [0, 0, 0, 0],
                            },
                            Stime: {
                                Sec: 0,
                                Usec: 20864,
                                Pad_cgo_0: [0, 0, 0, 0],
                            },
                            Maxrss: 24985600,
                            Ixrss: 0,
                            Idrss: 0,
                            Isrss: 0,
                            Minflt: 3428,
                            Majflt: 2903,
                            Nswap: 0,
                            Inblock: 0,
                            Oublock: 0,
                            Msgsnd: 0,
                            Msgrcv: 0,
                            Nsignals: 61,
                            Nvcsw: 29,
                            Nivcsw: 922,
                        },
                    },
                    Stderr: nil,
                },
                Output: {
                    stdout: {
                        Lines: nil,
                        merged: {
                            Mutex: {state: 0, sema: 0},
                            Lines: [
                                "The connection to the server localhost:8080 was refused - did you specify the right host or port?",
                            ],
                        },
                    },
                    stderr: {
                        Lines: [
                            "The connection to the server localhost:8080 was refused - did you specify the right host or port?",
                        ],
                        merged: {
                            Mutex: {state: 0, sema: 0},
                            Lines: [
                                "The connection to the server localhost:8080 was refused - did you specify the right host or port?",
                            ],
                        },
                    },
                    merged: {
                        Mutex: {state: 0, sema: 0},
                        Lines: [
                            "The connection to the server localhost:8080 was refused - did you specify the right host or port?",
                        ],
                    },
                },
            }
        to be nil
--- FAIL: TestMedusaDeploymentScenario (0.14s)
    --- FAIL: TestMedusaDeploymentScenario/Minio:_Initialize_cluster (0.14s)
FAIL
FAIL	github.com/k8ssandra/k8ssandra/tests/integration	6.177s
FAIL
```

You can see it stops after the first failure.

tests/integration/integration_suite_test.go
adejanovski: See [my previous answer](https://github.com/k8ssandra/k8ssandra/pull/544#discussion_r595794009).

tests/integration/integration_suite_test.go
adejanovski: See [my previous answer](https://github.com/k8ssandra/k8ssandra/pull/544#discussion_r595794009). 🙂

tests/integration/steps/integration_steps.go
adejanovski: ok, it's going to be more verbose but can be done.

tests/integration/steps/integration_steps.go
adejanovski: We're expecting a single backup here since the test runs with a single node.
I'll check `finishTime` though so we're sure to have the right behavior if we add nodes.

tests/integration/steps/integration_steps.go
adejanovski: Gotcha, will change the code accordingly.

tests/integration/steps/integration_steps.go
adejanovski: Yes, it's planned 👍

tests/integration/steps/integration_steps.go
```diff
@@ -0,0 +1,622 @@
...
+func Find(slice []string, val string) bool {
+	for _, item := range slice {
+		if item == val {
+			return true
+		}
+	}
+	return false
+}
+
+func g(t *testing.T) *gomega.WithT {
```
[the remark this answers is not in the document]
adejanovski: Thanks!

tests/integration/steps/integration_steps.go
adejanovski: Right, I had an issue initially with the dot import when invoking `NewWithT()` but I must have messed up as it's working now.
Good call 👍

tests/integration/steps/integration_steps.go
adejanovski: Agreed, I'll refactor accordingly.

tests/integration/steps/integration_steps.go
adejanovski: The problem then is the time scenarios will take as I've seen namespace deletions take several minutes to complete.
Do we want to test that the namespace was fully deleted this in all scenarios?
Maybe we could create a test scenario that would be specific to this in order to shorten the overall duration?

tests/integration/steps/integration_steps.go
adejanovski: They are used by [cleanupCluster](https://github.com/k8ssandra/k8ssandra/blob/5c018349a3fd6ee291a419fc082dfb85d04bdaee/tests/integration/integration_suite_test.go#L27) in the `integration` package.

tests/integration/steps/integration_steps.go
adejanovski: Nope, it's reading from it. This is to allow using `~` notation when referencing secrets that are not stored in the GH repo.

tests/integration/steps/integration_steps.go
adejanovski: I was rather considering the command could fail and that the repeated checks with `Eventually` would be enough to catch (later on) if there was a real problem.
Are you referring to the `err` from the above `RunKubectlAndGetOutputE()` call or a check on the 
`json.Unmarshall()` ?

tests/integration/integration_suite_test.go
jsanda: >I'm -1 on not using subtests here since we can achieve the desired behavior and exit on failure.

It is not exactly the desired behavior. I believe `-failfast` halts execution of all non-parallel tests after the first failure. Consider this example:

```go
// example1_test.go

func TestFoo(t *testing.T) {
	t.Run("TEST ONE", testOne)
	t.Run("TEST TWO", testTwo)
	t.Run("TEST THREE", testThree)
}

func TestBar(t *testing.T) {

}

func testOne(t *testing.T) {
	t.Fatal("test one failed")
}

func testTwo(t *testing.T) {
	t.Fatal("test two failed")
}

func testThree(t *testing.T) {
}

// example2_test.go
func TestAnotherFoo(t *testing.T) {
}

func TestAnotherBar(t *testing.T) {
	
}
```

Here is the output when I run with `-failfast`:

```
go test -v -failfast -timeout 1800s ./test/e2e/...
=== RUN   TestFoo
=== RUN   TestFoo/TEST_ONE
    example1_test.go:16: test one failed
--- FAIL: TestFoo (0.00s)
    --- FAIL: TestFoo/TEST_ONE (0.00s)
FAIL
FAIL	github.com/k8ssandra/medusa-operator/test/e2e	0.367s
```
Not only are the remaining subtests in `TestFoo` skipped but `TestBar`, `TestAnotherBar` and `TestAnotherFoo` are skipped as well. I don't think we want this behavior.

tests/integration/integration_suite_test.go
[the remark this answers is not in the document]
jeffbanks: Right, that is fine, I was thinking we could build out a folder structure with placeholders so developers know to think about a distinction.

tests/integration/integration_suite_test.go
jeffbanks: @jsanda In the case of integration tests, is it safe then to say that each test Go file will scope a single integration test run and will be run independent of the others?  What kind of resources will be required to do this effectively.  I'm thinking of things like snapshots of current k8ssandra state to determine root cause.  If I understand correctly, we would need to use appropriate test-name, namespace, context, etc.  

@adejanovski How would that work with your current approach of harvesting details after a test run?

tests/integration/integration_suite_test.go
burmanm: Yes, each subtest should be a "self-contained" test in a sense that it's a sequence of steps. So for example, these 8 steps could be one test. If you wish to have the logging, add t.Log() instead of t.Run(fmt.Sprintf( and so on.

adejanovski: It could be more efficient to have a live chat about this because we don't seem to consider the same requirements in terms of parallelism and why it would be useful at this stage. If we don't have the same goal that conversation will quickly heat up as we talk past each other 😅

jsanda: @adejanovski what you explained to me about the CI set up mostly addressed my questions around the test organization/structure in the context of CI. To be clear though what I had envisioned for subtests was more along the lines of what @burmanm is describing.

 The explanation does not really address running tests locally. In a [previous comment](https://github.com/k8ssandra/k8ssandra/pull/544#discussion_r595794009) you gave the following example to demonstrate that the test execution will stop after the first failure:

```
% go test . -v -failfast -run=TestMedusaDeploymentScenario/"Minio:"
```

This is fine if I just want to run one test and its subtests. I don't think this works when running multiple tests. Test execution will stop on the first failure and not execute subsequent tests. It is nice to have the `-failfast` option, but when I run multiple tests, I want all of them to run and see the failures at the end.

I feel like the current approach with the fail fast, fined-grained subtests puts us back into a situation like we are in with Ginkgo, and that is a reason I wanted to move from Ginkgo to `testing`. 

On a semi-related note, gomega, at least as used, does not provide the behavior I want. Let me illustrate with examples. Here is a test using gomega:

```go
func TestFoo(t *testing.T) {
	g(t).Expect(1).To(gomega.Equal(2))
	g(t).Expect(1).To(gomega.Equal(3))
}
```
and here is the output when I run it:

```
$ go test -v -run=TestFoo ./tests/integration/...
=== RUN   TestFoo
    integration_suite_test.go:86:
        Expected
            <int>: 1
        to equal
            <int>: 2
--- FAIL: TestFoo (0.00s)
```
The second statement does not get executed. Now look at this next example:

```go
func TestBar(t *testing.T) {
	t.Error("1 != 2")
	t.Error("1 != 3")
}
```
and here is the output when I run it:

```
$ go test -v -run=TestBar ./tests/integration/...
=== RUN   TestBar
    integration_suite_test.go:91: 1 != 2
    integration_suite_test.go:92: 1 != 3
--- FAIL: TestBar (0.00s)
```
The errors are logged and execution continues until the end of the test. This is critical functionality IMO for integration/e2e tests.

burmanm: A single test is something that verifies functionality. These steps are often not verifying a functionality, but more like a single assertions only. "CreateCassandraTable" is not a test for k8ssandra, we don't ship "CassandraTableCreator". "Verify Medusa works" is a test, it's one of the functions we ship. The distinction is a bit more complex between "create a backup and verify it works" and "create a backup, verify it works and then restore it". Those I could see as two different tests.

adejanovski: @burmanm, I find that definition a little dogmatic 🙂 . Anything that triggers an action and then performs checks through assertions could qualify as a test. In your example, creating a table allows us to check that Cassandra is reachable and behaves as expected. I guess the granularity is then a matter of personal taste and we need to commonly agree on the level we want.



Also, I'd like us to be on the same page: Wasn't the ability to run a single step one of your requirements? If we remove the subtests altogether and replace them with loggings, we'll only be able to run the whole scenario each time.

@jsanda, I don't think subtests are the main issue in the behavior you describe: it's the distinction between `Fail` and `Fatal` which can't be made when using Gomega. [It defaults](https://github.com/onsi/gomega/pull/215/files) to `Fatal` while the `testing` package provides both behaviors (`Fail` marking the test as failed but letting the execution continue). Your example with `t.Error()` isn't comparable because it calls `Log.Print()` and then `t.Fail()` underneath.
Now what we could do is use Gomega only for assertions that we consider fatal (or use Gomega only for its async assertions through `Eventually`).
Also both your snippets aren't equivalent as `t.Error()` is not performing an assertion, it's marking the running test case as failed without stopping its execution.

Then we can definitely discuss the organization of subtests and nest them accordingly to what is ok to run after a failed one. Let's take the following structure:
```
func TestReaperDeploymentScenario(t *testing.T) {
	ctx := initContext()
	t.Run("Initialize cluster", func(t *testing.T) {
		initializeCluster(t, ctx)
		t.Run("Deploy cluster", func(t *testing.T) {
			...
			...
			t.Run("Verify Reaper works as expected", func(t *testing.T) {
				t.Fatal("Fake failure")
				...
			})
			t.Run("Verify Prometheus is accessible", func(t *testing.T) {
				log.Println(Info("Prometheus is accessible, yay!"))
			})
			cleanupCluster(t, ctx)
		})
	})
}
```

Assuming we don't use `-failfast`, the `Verify Reaper works as expected` will fail but `Verify Prometheus is accessible` will be able to run:

```
=== RUN   TestReaperDeploymentScenario/Initialize_cluster/Deploy_cluster/Verify_Reaper_works_as_expected
    integration_suite_test.go:110: Fake failure
=== RUN   TestReaperDeploymentScenario/Initialize_cluster/Deploy_cluster/Verify_Prometheus_is_accessible
2021/03/18 10:05:59 Prometheus is accessible, yay!
2021/03/18 10:05:59 === Cleaning up cluster
TestReaperDeploymentScenario/Initialize_cluster/Deploy_cluster 2021-03-18T10:06:12+01:00 client.go:33: Configuring kubectl using config file /Users/adejanovski/.kube/config with context
TestReaperDeploymentScenario/Initialize_cluster/Deploy_cluster 2021-03-18T10:06:12+01:00 client.go:33: Configuring kubectl using config file /Users/adejanovski/.kube/config with context
TestReaperDeploymentScenario/Initialize_cluster/Deploy_cluster 2021-03-18T10:06:12+01:00 client.go:33: Configuring kubectl using config file /Users/adejanovski/.kube/config with context
--- FAIL: TestReaperDeploymentScenario (467.89s)
    --- FAIL: TestReaperDeploymentScenario/Initialize_cluster (467.89s)
        --- FAIL: TestReaperDeploymentScenario/Initialize_cluster/Deploy_cluster (462.21s)
            --- FAIL: TestReaperDeploymentScenario/Initialize_cluster/Deploy_cluster/Verify_Reaper_works_as_expected (0.00s)
            --- PASS: TestReaperDeploymentScenario/Initialize_cluster/Deploy_cluster/Verify_Prometheus_is_accessible (0.00s)
```

But it will allow cleanupCluster() to run, which is not really something we want as it will prevent us from collecting artifacts or investigate directly on the k8s cluster.


That structure though will prevent any nested subtest from running if the parent test fails, which allows us to avoid running tests that would fail for sure.
If I move the failure at the beginning of the `Deploy cluster` subtest, I can see that no nested test will run:

```
func TestReaperDeploymentScenario(t *testing.T) {
	ctx := initContext()
	t.Run("Initialize cluster", func(t *testing.T) {
		initializeCluster(t, ctx)
		t.Run("Deploy cluster", func(t *testing.T) {
			t.Fatal("Fake failure")
			...
			t.Run("Verify Reaper works as expected", func(t *testing.T) {
				...
			})
			t.Run("Verify Prometheus is accessible", func(t *testing.T) {
				log.Println(Info("Prometheus is accessible, yay!"))
			})
			cleanupCluster(t, ctx)
		})
	})
}
```
It gives the following execution:
```
=== RUN   TestReaperDeploymentScenario/Initialize_cluster
...
...
=== RUN   TestReaperDeploymentScenario/Initialize_cluster/Deploy_cluster
    integration_suite_test.go:105: Fake failure
--- FAIL: TestReaperDeploymentScenario (5.76s)
    --- FAIL: TestReaperDeploymentScenario/Initialize_cluster (5.76s)
        --- FAIL: TestReaperDeploymentScenario/Initialize_cluster/Deploy_cluster (0.00s)
```

Now, `t.Run()` returns a boolean which tells us if the subtests succeeded. We could do the following to only cleanup the cluster if both the lowest level tests succeeded:
```
func TestReaperDeploymentScenario(t *testing.T) {
	ctx := initContext()
	t.Run("Initialize cluster", func(t *testing.T) {
		initializeCluster(t, ctx)
		t.Run("Deploy cluster", func(t *testing.T) {
			...
			reaperSucceeded := t.Run("Verify Reaper works as expected", func(t *testing.T) {
				t.Fatal("Fake failure")
				...
			})
			promSucceeded := t.Run("Verify Prometheus is accessible", func(t *testing.T) {
				log.Println(Info("Prometheus is accessible, yay!"))
			})
			if reaperSucceeded && promSucceeded {
				t.Run("Cleanup cluster", func(t *testing.T) {
					cleanupCluster(t, ctx)
				})
			}
		})
	})
}
```

Which gives the following output, effectively skipping the cleanup cluster test:
```
=== RUN   TestReaperDeploymentScenario/Initialize_cluster
...
=== RUN   TestReaperDeploymentScenario/Initialize_cluster/Deploy_cluster
...
=== RUN   TestReaperDeploymentScenario/Initialize_cluster/Deploy_cluster/Verify_Reaper_works_as_expected
    integration_suite_test.go:110: Fake failure
=== RUN   TestReaperDeploymentScenario/Initialize_cluster/Deploy_cluster/Verify_Prometheus_is_accessible
2021/03/18 10:46:32 Prometheus is accessible, yay!
--- FAIL: TestReaperDeploymentScenario (475.83s)
    --- FAIL: TestReaperDeploymentScenario/Initialize_cluster (475.83s)
        --- FAIL: TestReaperDeploymentScenario/Initialize_cluster/Deploy_cluster (470.12s)
            --- FAIL: TestReaperDeploymentScenario/Initialize_cluster/Deploy_cluster/Verify_Reaper_works_as_expected (0.00s)
            --- PASS: TestReaperDeploymentScenario/Initialize_cluster/Deploy_cluster/Verify_Prometheus_is_accessible (0.00s)
```

The above code would provide, I think, the behavior that you want, which could be further extended by not using Gomega for non fatal checks (although I don't really get the point of such assertions):

- Remove the use of `-failfast`
- Allow to run subtests from the same parent subtest to run despite each other success
- Prevent nested subtest to run if the parent test failed (giving us an explicit failfast behavior) and allow conditional behavior on running tests at the same level.

What we lose though:

- A bit of readability and simplicity of the code (the flat structure is easier to read)
- The ability to run individual subtests if they are nested ones. All the parent tests will have to run in order for the target subtest to run. That pretty much brings us back to Ginkgo's "run all or nothing":
```
% go test  -test.timeout=30m ./tests/integration -v -run=TestReaperDeploymentScenario/Initialize_cluster/Deploy_cluster/"Verify Prometheus"
=== RUN   TestReaperDeploymentScenario
=== RUN   TestReaperDeploymentScenario/Initialize_cluster
...
=== RUN   TestReaperDeploymentScenario/Initialize_cluster/Deploy_cluster
...
=== RUN   TestReaperDeploymentScenario/Initialize_cluster/Deploy_cluster/Verify_Prometheus_is_accessible
2021/03/18 11:06:10 Prometheus is accessible, yay!
--- PASS: TestReaperDeploymentScenario (456.56s)
    --- PASS: TestReaperDeploymentScenario/Initialize_cluster (456.56s)
        --- PASS: TestReaperDeploymentScenario/Initialize_cluster/Deploy_cluster (450.84s)
            --- PASS: TestReaperDeploymentScenario/Initialize_cluster/Deploy_cluster/Verify_Prometheus_is_accessible (0.00s)
```

adejanovski: Another option could be to avoid nesting subtests and rely on testing the result of mandatory ones:
```
func TestReaperDeploymentScenario(t *testing.T) {
	ctx := initContext()
	t.Run("Initialize cluster", func(t *testing.T) {
		initializeCluster(t, ctx)
	})

	deploySucceeded := t.Run("Deploy cluster", func(t *testing.T) {
		deployClusterForReaper(t, ctx)
		checkResourcePresenceForReaper(t, ctx)
	})

	if deploySucceeded {
		t.Run("Wait for Reaper pod to be ready", func(t *testing.T) {
			waitForReaperPod(t, ctx)
		})
		t.Run("Check Reaper has registered the required elements", func(t *testing.T) {
			checkReaperRegistered(t, ctx)
		})
		repairStarted := t.Run("Start repair on the reaper_db keyspace", func(t *testing.T) {
			triggerRepair(t, ctx)
		})

		if repairStarted {
			repairSuccess := t.Run("Wait for at least one segment to be processed and cancel the repair", func(t *testing.T) {
				waitForSegmentDoneAndCancel(t, ctx)
			})

			if repairSuccess {
				t.Run("Tear down cluster", func(t *testing.T) {
					cleanupCluster(t, ctx)
				})
			}
		}
	}
}
```
This allows us to run a single subtest even if it's conditioned to a subtests that won't be executed:
```
% go test  -test.timeout=30m ./tests/integration -v -run=TestReaperDeploymentScenario/"Start repair on the reaper_db keyspace"
=== RUN   TestReaperDeploymentScenario
=== RUN   TestReaperDeploymentScenario/Start_repair_on_the_reaper_db_keyspace
...
2021/03/18 11:12:13 Reaper response:
2021/03/18 11:12:13 Reaper status code: 200
--- PASS: TestReaperDeploymentScenario (0.69s)
    --- PASS: TestReaperDeploymentScenario/Start_repair_on_the_reaper_db_keyspace (0.69s)
PASS
ok  	github.com/k8ssandra/k8ssandra/tests/integration	1.138s
```
We can check how it behaves when triggering a Fatal failure on a subtest that is not conditioning subsequent subtests and then one one that is:
```
func TestReaperDeploymentScenario(t *testing.T) {
	ctx := initContext()
	t.Run("Initialize cluster", func(t *testing.T) {
		t.Fatal("Not really fatal failure")
		initializeCluster(t, ctx)
	})

	deploySucceeded := t.Run("Deploy cluster", func(t *testing.T) {
		t.Fatal("Really fatal failure")
		deployClusterForReaper(t, ctx)
		checkResourcePresenceForReaper(t, ctx)
	})

	if deploySucceeded {
		t.Run("Wait for Reaper pod to be ready", func(t *testing.T) {
			waitForReaperPod(t, ctx)
		})
...

```
Gives:
```
% grc go test  -test.timeout=30m ./tests/integration -v -run=TestReaperDeploymentScenario
=== RUN   TestReaperDeploymentScenario
=== RUN   TestReaperDeploymentScenario/Initialize_cluster
    integration_suite_test.go:72: Non fatal failure
=== RUN   TestReaperDeploymentScenario/Deploy_cluster
    integration_suite_test.go:77: Really fatal failure
--- FAIL: TestReaperDeploymentScenario (0.00s)
    --- FAIL: TestReaperDeploymentScenario/Initialize_cluster (0.00s)
    --- FAIL: TestReaperDeploymentScenario/Deploy_cluster (0.00s)
FAIL
FAIL	github.com/k8ssandra/k8ssandra/tests/integration	0.333s
FAIL
```

Would that better match your requirements?

tests/integration/steps/integration_steps.go
adejanovski: Sure, but we have control over the deployment parameters and they are using a fixed value.
It can definitely be parameterized in the future if tests require it.

tests/integration/integration_suite_test.go
burmanm: Testing that Cassandra can write to a table is more like a setup phase, or at most a unit test. It's not an integration test since there's nothing that was integrated - not Medusa and Cassandra. Nor a e2e test since we didn't actually accomplish anything.

I don't find anything useful in writing to a Cassandra table, why would I want to test that? And if I do, I'd like to see it verify the structure of the Cassandra files in the memtable, disk etc. What functionality would it test in k8ssandra? And if there's nothing useful in an assertion, then it's not a test - it's just a step inside a test. Same with some Medusa step, I don't want to test that its s3 integration can list a directory - that's a unit test. I want to test that it can create a backup from my Cassandra installation and send it to a S3, that's 3 components that are integrated and work together - that's what I want to test. In Medusa's unit tests you can verify the internal behavior only.

In the Reaper example:

```go
	if deploySucceeded {
		t.Run("Wait for Reaper pod to be ready", func(t *testing.T) {
			waitForReaperPod(t, ctx)
		})
		t.Run("Check Reaper has registered the required elements", func(t *testing.T) {
			checkReaperRegistered(t, ctx)
		})
		repairStarted := t.Run("Start repair on the reaper_db keyspace", func(t *testing.T) {
			triggerRepair(t, ctx)
		})
```

This should rather be:

```go
t.Log("Wait for Reaper pod to be ready")
waitForReaperPod(t, ctx)
t.Log("Check Reaper has registered the required elements")
checkReaperRegistered(t, ctx)
t.Logf(""Start repair on the %s keyspace", reaperKeyspace)
triggerRepair(t, ctx)
```

Although I would probably bundle the first two functions together, unless you want to test something that hasn't deployed reaper. And if that's the case, then the ``checkReaperRegistered`` should take a parameter what reaper and what cluster are registered. At the moment it just magically checks something that's hardcoded (at least for the reader) - nothing reusable and should not in that sense be a separate function. Making it a function for a struct makes it again not-reusable, so I don't think that's the correct choice - it should just take the input parameters. A massive parameter struct doesn't look nice either, do each of the functions really need all that information? I'd rather see it from my IDE which input parameters it needs and use the return values like they're meant to be used - it's more like Go in that case (it's in the effective go to avoid mutable input parameter style as the output). 

Think of it like:

```go
t.Log("Deploying a reaper instance")
reaper := deployReaperInstance(t, namespace)
t.Log("Wait for Reaper pod to be ready")
waitForReaperPod(t, reaper)
t.Log("Check Reaper has registered the required elements")
checkReaperRegistered(t, reaper, cassdc)
t.Logf(""Start repair on the %s keyspace", reaperKeyspace)
triggerRepair(t, reaper, reaperKeyspace)
```

Or even removing the whole t from those and doing the assertion here (I'll write with gomega, but that's not relevant here):

```go
t.Log("Deploying a reaper instance")
reaper, err := deployReaperInstance(namespace)
Expect(err).To(BeNil())

t.Log("Wait for Reaper pod to be ready")
Expect(waitForReaperPod(reaper)).To(Succeed())

t.Log("Check Reaper has registered the required elements")
err = checkReaperRegistered(reaper, cassdc)
Expect(err).To(BeNil()) // check for k8sClient failures

// Or, if the checks are simple - keep them inline since too small functions are not worthy of being functions
Expect(reaper.Clusters).To(HaveKey(cassdc))

t.Logf("Start repair on the %s keyspace", reaperKeyspace)
Expect(triggerRepair(reaper, reaperKeyspace)).To(Succeed())
```

Now, with a single glance I can see the steps the test took and I can see all the verifications it did. I simplified my methods and I made sure my methods do not have to be test-only, lets say some of them called a client library such as reaper-client-go. I've reused existing code or I've allowed some other functions in the software to reuse what I've written. And if my code and test use the same methods to deploy the reaper - even better, then I'm really testing the actual implementation. Rule of thumb, if the test can fit a single screen, it's not too complex and should not be split.

adejanovski: > Testing that Cassandra can write to a table is more like a setup phase

I still disagree and think this is semantics. Even then do you really see this as a blocker for this PR?
Maybe the code currently splits up in too many steps (depending on your sensitivity) that could be bundled together. But it brings some clarity which will be helpful to understand what these tests do instead of having to follow the trail of all that's being done by following the track of each sub function call.
We can always be more concise and wrap everything together, but that complicates understanding what the test does.

Same goes for the output: subtests allow us to have a very granular view of which tests passed and which failed in the summary:

```
--- FAIL: TestReaperDeploymentScenario (0.00s)
    --- FAIL: TestReaperDeploymentScenario/Initialize_cluster (0.00s)
    --- PASS: TestReaperDeploymentScenario/Test_that_will_pass_for_sure (0.00s)
    --- FAIL: TestReaperDeploymentScenario/Deploy_cluster (0.00s)
```
We won't get that if we don't use subtests.

Let's put aside the input parameters and hardcoded stuff you're referring to in order to focus on flat vs subtests. The context variable introduction is a wip to remove global variables as per one of your comments and based on a suggestion from John. This integration test suite will hopefully go through refactorings in the future and evolve in a direction or another depending on the needs we have.

Mixing concerns makes it harder to answer the initial questions, so I'd like to refocus on the initial request from @jsanda's: have a way for several checks to fail without preventing other specific checks from running, while still allowing to stop the test execution as other subsequent checks would consistently fail. So `-failfast` was too extreme and he wanted more granular control on what can run.
How would you solve this in your proposal?

Also you repeatedly dodge the question about being able to run only specific steps in a scenario: Is that a requirement or do you want to re-run the whole scenario each time one step fails?

burmanm: I have no interest in running each assertion, I can't do that in real life. Lets say I modified the reaper-operator and before pushing I want to test it.

What I need is to run the "setup phase" first to setup my cluster. What I then want is to run the reaper process and let it clean the mess, fix my errors, rerun the reaper process and fix the mess. I don't want to run smaller parts - because that testing should be part of reaper-operator. I also can't run smaller subtests, because I will probably need to cleanup the mess that might have been caused by the buggy operator for example. So I need the init and clean for that reaper data, but I don't need to clean and redo the CassandraDatacenter part - I might need to remove an annotation, but that was hopefully something the reaper part added and removed.

We need to keep in mind what tests are part of k8ssandra and which are part of the individual operators. Just like we're not copying Medusa's internal tests here, we only want to see it from the sky-high perspective, higher than what medusa-operator does. We're not even testing medusa-operator's single details. And whatever this higher view is - is a single test which might need running. If it's smaller test, then it should be done in the component itself and not here.

.github/workflows/integration-tests.yaml
adejanovski: done, thanks!

tests/integration/integration_suite_test.go
jsanda: >I don't think subtests are the main issue in the behavior you describe: it's the distinction between Fail and Fatal which can't be made when using Gomega. It defaults to Fatal while the testing package provides both behaviors (Fail marking the test as failed but letting the execution continue). 

Subtests as designed particularly with `-failfast` are my main concern. I am also pointing out the distinction between `Fail`/`Error` and `Fatal` which is an important one.

>Also both your snippets aren't equivalent as t.Error() is not performing an assertion, it's marking the running test case as failed without stopping its execution.

The snippets were not intended to be the same. They were intended to contrast the differences.

jsanda: >Assuming we don't use -failfast, the Verify Reaper works as expected will fail but Verify Prometheus is accessible will be able to run:
...
>But it will allow cleanupCluster() to run, which is not really something we want as it will prevent us from collecting artifacts or investigate directly on the k8s cluster.

We can easily track whether there are any subtest failures and check that in `cleanupCluster`.

jsanda: >The above code would provide, I think, the behavior that you want, which could be further extended by not using Gomega for non fatal checks (although I don't really get the point of such assertions):
>* Remove the use of -failfast
>* Allow to run subtests from the same parent subtest to run despite each other success
>* Prevent nested subtest to run if the parent test failed (giving us an explicit failfast behavior) and allow conditional behavior on running tests at the same level.

I think we are making this more complicated than it needs to be right now. Do we really need to decide how/when/if all subtests will run right now? 

If I write a test with subtests A, B, and C and I want to skip B if A fails but still run C, then I can add the logic to do so. Over time if we see common patterns then we refactor accordingly.

jsanda: Here is what I have had in mind for subtest organization:

```go
func TestInstallFullStack(t* testing.T) {
    t.Run("verify reaper stuff", testReaperInstall)
    t.Run("verify stargate install", testStargateInstall)
    t.Run("verify medusa install", testMedusaInstall)
    t.Run("verify prometheus install", testPrometheusInstall)
    t.Run("verify grafana install", testGrafanaInstall) 
}
```
I don't think we need subtests to be more fine-grained than this. Of course this is pretty subjective.

tests/integration/integration_suite_test.go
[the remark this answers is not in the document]
jeffbanks: @adejanovski @jsanda I wanted to add this note since it seems relevant.  We should consider adding use of something like this [fake-client](https://github.com/kubernetes/client-go/blob/master/examples/fake-client/main_test.go) in our integration testing to alleviate having all resources available every time.  This will greatly speed up our local runs where we aren't performing a specific verification with another resource that can be mocked.

When we discuss `e2e`, I see all the resources running and we inject known inputs (i.e. state changes, configs, etc.) expecting known outputs (state, configs, etc.).  With `integration` tests, specific components are involved and where needed, the others can be `mocked`.  If we take this approach early on, it will make things much more manageable for local integration testing.

tests/integration/steps/integration_steps.go
adejanovski: I've now removed all uses of `kubectl wait` throughout the integration test suite.

tests/integration/steps/integration_steps.go
adejanovski: Done 👍

tests/integration/steps/integration_steps.go
adejanovski: Done 👍

tests/integration/steps/integration_steps.go
adejanovski: I'd like to reduce the scope of this PR and the related issue and defer the use of the reaper go client to a later time. It requires to make some changes in the Reaper client to support the endpoints that are needed to trigger a repair, cancel it and check the segments state, which is a little out of scope.
I'll create a ticket to track this.

tests/integration/steps/integration_steps.go
adejanovski: I've removed all global variables and context objects.
We either have constants or variables that are local to each test scenario.

tests/integration/steps/integration_steps.go
adejanovski: Done

tests/integration/steps/integration_steps.go
adejanovski: Done 👍

adejanovski:
I've applied most of the changes that were requested/recommended.
There's a full stack scenario on top of the individual ones, but it currently won't run in CI due to limited disk space in the CI runners.
Test scenarios run individually and the matrix now includes different Cassandra versions (3.11.10 and 4.0.0).
Using the reaper go client is pushed back to another PR as a lot of time was spent on this PR already and I'd like to limit the scope of changes.
The dev docs introduced by @jeffbanks were updated as part of this PR to show how to run the integration tests.

tests/integration/integration_suite_test.go
```diff
@@ -1,13 +1,326 @@
...
+	"os"
 	"testing"
+)
 
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
+const (
+	medusaTestTable    = "medusa_test"
+	medusaTestKeyspace = "medusa"
+	TRAEFIK_NAMESPACE  = "traefik"
```
burmanm, line 17: This is JAVA_ISM.

tests/integration/integration_suite_test.go
```diff
@@ -1,13 +1,326 @@
...
+	deployClusterForMonitoring(t, namespace)
+	prometheusSuccess := t.Run("Test Prometheus", func(t *testing.T) {
+		testPrometheus(t, namespace)
+	})
+
+	grafanaSuccess := t.Run("Test Grafana", func(t *testing.T) {
+		testGrafana(t, namespace)
+	})
+
+	if prometheusSuccess && grafanaSuccess {
```
burmanm, line 261: So if a test fails, the cluster is left to a disarray and user must clean it up?

tests/integration/integration_suite_test.go
```diff
@@ -1,13 +1,326 @@
...
+// - Check that Grafana is reachable through http
+// - Terminate the namespace and delete the cluster
+func TestMonitoringDeploymentScenario(t *testing.T) {
+	// only runs when the -short flag is used
+	if !testing.Short() {
+		t.Skip("Skipped in non-short mode")
+	}
+	namespace := initializeCluster(t)
+	deployClusterForMonitoring(t, namespace)
+	prometheusSuccess := t.Run("Test Prometheus", func(t *testing.T) {
```
burmanm, line 253: What's the point of t.Run in this case? Why not just testPrometheus(t, namespace) ?

tests/integration/steps/stargate_steps.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func authEndpointIsReachable(t *testing.T) bool {
+	restClient := resty.New()
+	response, err := restClient.R().Get("http://stargate.127.0.0.1.nip.io:8081/v1/auth")
+	if err != nil {
+		log.Printf("Failed connecting to Stargate auth endpoint: %s", err.Error())
+		return false
+	}
+	return response.StatusCode() == 405 // This endpoint should be invoked with a Post, we expect a 405 with a Get
```
jeffbanks, line 62: What if response is nil?  There are some other cases of this in some of the other functions as well.  I just didn't want us to get misleading errors that we have to trace if we need to protect them inside our verification functions.

tests/integration/steps/stargate_steps.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func readStargateDocument(t *testing.T, token, docNamespace, documentId string) []byte {
+	restClient := resty.New()
+	response, err := restClient.R().
+		SetHeader("Content-Type", "application/json").
+		SetHeader("X-Cassandra-Token", token).
+		Get(fmt.Sprintf("http://stargate.127.0.0.1.nip.io:8082/v2/namespaces/%s/collections/movies/%s", docNamespace, documentId))
+	g(t).Expect(err).To(BeNil(), "Failed reading Stargate document")
+	return response.Body()
```
jeffbanks, line 132: Same possible issue with response being nil, throwing an exception here.

tests/integration/integration_suite_test.go
jeffbanks: My opinion, based on early integration tests constructed, is to consider a full `cluster cleanup` as `teardown` step and `cluster setup` as `setup` step.

If the `scope of the test` is to verify a specific delete/removal use case, then prior to full teardown step, we would verify that there wasn't an error and the resource is properly cleaned up.

tests/integration/integration_suite_test.go
adejanovski: ok, I'll use camel case then :+1:

tests/integration/integration_suite_test.go
adejanovski: yes, and that's necessary for investigation purposes (in and out of the CI context, but specifically for CI we need to keep the cluster running). When running the integration tests with `make kind-integ-test`, the kind cluster will get deleted before being recreated.

tests/integration/steps/stargate_steps.go
adejanovski: Response should only be `nil` if `err` was different than `nil` which means we've already exited at line 60. I think I've used pretty much the same pattern in different places indeed, but errors should be caught appropriately if needed.

tests/integration/steps/stargate_steps.go
adejanovski: the call to `Expect(err)...` on the previous line protects us here. If response is nil indeed, then something went totally wrong in the rest call and we shouldn't reach the `return` instruction.

tests/integration/integration_suite_test.go
adejanovski: For this specific test scenario, it will allow to test Grafana even if the Prometheus test fails.
I can remove the subtest for the Stargate and Reaper tests though since there's a single component that is being tested there.

tests/integration/integration_suite_test.go
adejanovski: @jeffbanks, I see a few issues there (which could be due to my Go skills):

- The `initializeCluster()` function (which is what would be in `setUp()`) generates a unique namespace name which is passed on to all the test functions and there were concerns raised previously with the use of global variables. Now if we can pass the generated value on from `TestMain()` to the individual tests, then that's a good thing to do as it would reduce code duplication.
- The `initializeCluster()` and `cleanupCluster()` functions use Terratest underneath which requires to use the `testing.T` object which doesn't seem to be available in the `TestMain()` function. It seems like the only thing we can get out of the individual tests is the return code.

Unless we have an easy solution to do this (which doesn't involve removing Terratest, which wouldn't be a small change) then I'd like to defer this as part of a future refactoring.

What do you think?

tests/integration/integration_suite_test.go
adejanovski: Fixed

tests/integration/integration_suite_test.go
jeffbanks: 👍 I agree, let's address this moving forward, I was able to use `Terratest`  with a combination of setup and teardowns by managing the scope of context and namespace.   I am good with iterating and refactoring some of this as we understand more of what we want.  IMO, this PR has become massive, with too many working parts.  :smile:

jeffbanks (approved):
I am good with moving forward with this PR as a foundation to get us started.  We can refactor and iterate as needed as part of future PRs with smaller scope.

Great job overall Alex!  This will be extremely valuable for maintaining the quality of K8ssandra.

tests/integration/steps/integration_steps.go
jsanda: The duration for namespace deletion is increased due to the time it takes for the CassandraDatacenter deletion (see #570 and #571 for details).

@adejanovski after uninstalling the chart and before deleting the namespace, wait for the termination of the CassandraDatacenter. See https://github.com/jsanda/medusa-operator/blob/e2e-in-place-restore/test/framework/framework.go#L123 for an example.

tests/integration/steps/integration_steps.go
jsanda: Considering that this is a helper function that can be called from different tests I think it would be best to parameterize it now.

tests/integration/integration_suite_test.go
```diff
@@ -1,13 +1,316 @@
...
+// - Create a keyspace and a table
+// - Load 10 rows and check that we can read 10 rows
+// - Perform a backup with Medusa
+// - Load 10 rows and check that we can read 20 rows
+// - Restore the backup
+// - Verify that we can read 10 rows
+// - Terminate the namespace and delete the cluster
+func TestMedusaDeploymentScenario(t *testing.T) {
+	// only runs when the -short flag is used
+	if !testing.Short() {
```
jsanda, line 171: Is this backwards? If tests are run with the `-short` flag, that indicates that the user wants to skip long running tests.

tests/integration/steps/cassdc_client.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	cassdcapi "github.com/datastax/cass-operator/operator/pkg/apis/cassandra/v1beta1"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Client returns a controller-runtime client with cass-operator API defined
+func CassDcClient() (client.Client, error) {
```
jsanda, line 11: Note that this controller-runtime client be used with other types. Because CassandraDatacenter is a custom resource, you have to tell controller-runtime about it, but it already knows about all the "built-in" types like pods, services, etc.

jsanda (changes requested):

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,94 @@
...
+        uses: actions/cache@v2
+        with:
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-k8ssandra-buildx
+          restore-keys: |
+            ${{ runner.os }}-k8ssandra-buildx
+      - name: Set up Go 1.14
+        uses: actions/setup-go@v1
+        with:
+          go-version: 1.14
```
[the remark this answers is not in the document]
adejanovski: 👍

tests/integration/integration_suite_test.go
```diff
@@ -1,13 +1,316 @@
...
+
+	grafanaSuccess := t.Run("Test Grafana", func(t *testing.T) {
+		testGrafana(t, namespace)
+	})
+
+	stargateSuccess := t.Run("Test Stargate", func(t *testing.T) {
+		testStargate(t, namespace)
+	})
+
+	if reaperSuccess && medusaSuccess && prometheusSuccess && grafanaSuccess && stargateSuccess {
```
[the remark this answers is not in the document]
adejanovski: We don't, otherwise we have nothing left for investigation.
When ran locally, we want to be able to check what went wrong manually. When running in CI, in case of failure there are two steps that will trigger:

- artifacts extraction (outputs of various kubectl commands)
- enable ssh access (for manual investigation)

I don't see any drawback of not cleaning up the cluster in case of failure.

tests/integration/integration_suite_test.go
adejanovski: The long test here is the full stack one, which will run only if the `-short` flag isn't used. If individual components tests scenarios are to run, then you'll need to specify the `-short` flag. 
The full stack test already checks all that the individual tests do, so it's redundant to run them all when invoking `go test` without specifying a test function.
This way, if you want to test everything (without running redundant tests), you can just run `go test -test.timeout=30m ./tests/integration -v` and if you want to run an individual scenario : `go test $(GO_FLAGS) -test.timeout=30m ./tests/integration -v -short -run=TestReaperDeploymentScenario`.

Do you know of any other trick that would allow us to just run `TestFullStackScenario` by default, while still allowing to run the other tests if needed?

tests/integration/integration_suite_test.go
```diff
@@ -1,13 +1,316 @@
...
+				testMedusa(t, namespace, backend, backupName)
+			})
+			if medusaSuccess {
+				cleanupCluster(t, namespace)
+			}
+		})
+	}
+}
+
+func testMedusa(t *testing.T, namespace, backend, backupName string) {
```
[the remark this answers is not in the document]
adejanovski: Well, we'll have leftover files from each test. But they'll be in different folders since I've enabled multi-tenancy and a new namespace name is generated for each run.
We just need to think of doing some cleanup from time to time (or create a GH action that will handle it for us, which would be even better).

tests/integration/integration_suite_test.go
adejanovski: Sounds good to you @burmanm ?

tests/integration/steps/cassdc_client.go
adejanovski: Gotcha. 
I assume this means the client could be used to replace some calls we have in the tests to list pods and services by labels for example?
Ok to defer this to a subsequent issue?

tests/integration/steps/integration_steps.go
adejanovski: Done

tests/integration/integration_suite_test.go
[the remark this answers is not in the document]
burmanm: It leaves the cluster with stuff behind it and I need to manually clean it after each test run. I shouldn't need the end results in most cases - I should be able to see the problem from the log lines and then investigate. It's a rare occurrence in my opinion that I would need to check the status that was left behind.

I'd much rather see a flag that would indicate if I want I don't want it cleaned up. I at least use the same cluster for weeks without tearing it down and I don't want to pollute it with stuff that eats memory.

tests/integration/integration_suite_test.go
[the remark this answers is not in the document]
burmanm: So if I run these minio tests against my local installation, I'll have garbage sitting in my minio installation? Who is going to clean that?

tests/integration/steps/integration_steps.go
```diff
@@ -0,0 +1,422 @@
...
+		return cassdc.Status.CassandraOperatorProgress == cassdcapi.ProgressReady &&
+			cassdc.GetConditionStatus(cassdcapi.DatacenterReady) == v1.ConditionTrue
+	}, retryTimeout, retryInterval).Should(BeTrue())
+}
+
+func resourceWithLabelIsPresent(t *testing.T, namespace, resourceType, label string) bool {
+	switch resourceType {
+	case "pod":
+		pods := getPodsWithLabel(t, namespace, label)
+		if len(pods.Items) == 1 {
```
burmanm, line 194: Why not ``return len(pods.Items) == 1`` ?

burmanm: Also, you already have ``CountPodsWithLabel`` method also, so simplifying this would mean ``return CountPodsWithLabel(t, namespace, label) == 1``

Makefile
```diff
@@ -31,7 +32,17 @@ pkg-test:
 	source ${ENVTEST_ASSETS_DIR}/setup-envtest.sh && fetch_envtest_tools $(ENVTEST_ASSETS_DIR) && setup_envtest_env $(ENVTEST_ASSETS_DIR) && go test $(GO_FLAGS) -test.timeout=3m ./pkg/... -coverprofile cover.out
 
 integ-test:
-	go test $(GO_FLAGS) -test.timeout=5m ./tests/integration/... -coverprofile cover.out
+ifeq ($(TESTS), all)
+	go test $(GO_FLAGS) -test.timeout=30m ./tests/integration -v
```
burmanm, line 36: Missed this, -v is part of the GO_FLAGS, so it shouldn't be hardcoded.

Makefile
```diff
@@ -31,7 +32,17 @@ pkg-test:
...
+else
+	go test $(GO_FLAGS) -test.timeout=30m ./tests/integration -v -short -run=$(TESTS)
+endif
+
+kind-integ-test: create-kind-cluster integ-test
+
+create-kind-cluster:
+	kind delete cluster
```
burmanm, line 44: Why do we have to delete some other clusters (which might not have anything to do with this integ-test?)

tests/integration/integration_suite_test.go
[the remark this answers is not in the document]
jsanda: Let's take discussion of cleaning up backup storage outside of this PR. It is already. It is already hard enough to follow the all the discussions here, and we can address the cleanup separately in future work.

tests/integration/integration_suite_test.go
jsanda: Thanks for the explanation. Rather than have every individual component test check `testing.Short()`, I would switch the behavior and have `TestFullStackScenario` do the check. This would mean all tests run by default which I think is good and I can opt out of longer running tests with`-short`.

tests/integration/steps/cassdc_client.go
jsanda: >Ok to defer this to a subsequent issue?

Yes, please, thank you 😀

tests/integration/steps/integration_steps.go
adejanovski: Done: https://github.com/k8ssandra/k8ssandra/pull/544/files#diff-06af14725d9bc4e9eef72fdfd0d188ffbc8a3e657d3999dc6d1346a1eb37b9fdR53

tests/integration/steps/integration_steps.go
adejanovski: Done. The datacenter name (although being a constant) is parameterized everywhere (and if changed will be forcefully set during helm install).
Same goes for the release name.

tests/integration/integration_suite_test.go
adejanovski: Since the cleanup phase is now configurable and depends on both config and test success, I've had to wrap all tests within subtests.
I bet there are other ways to achieve this, but such changes would have to be done in a follow up PR.

tests/integration/integration_suite_test.go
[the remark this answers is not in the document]
adejanovski: Cleanup is now done by default in all cases, unless configured otherwise. CI will not perform cleanup to allow extracting artifacts and provide ssh access for investigation.

tests/integration/integration_suite_test.go
adejanovski: I've removed the use of `-short`. The Makefile handles running the appropriate tests only. Running `go test ...` would run all tests though, the full stack one and the individual ones as well.

tests/integration/steps/integration_steps.go
adejanovski: Good point, I introduced `CountPodsWithLabel` later and didn't remember it could be used here.

Makefile
adejanovski: done

Makefile
adejanovski: It's now creating/deleting a specific cluster named `k8ssandra-it`.

jsanda (approved):
@adejanovski Thanks for all the hard work, effort, and patience!

PR #587, opened 2021-03-25
Automate helm chart version management on releases

### LINKED ISSUE #557: Automated next/stable release versioning strategy
N/A

With a versioning and branching strategy defined via #549 we'll next need to implement automation of that release strategy that can produce chart releases to the next and stable chart repositories.

adejanovski:
**What this PR does**:
Automates Helm chart version number management for the next channel.
We're moving to using a fixed `x.x.x-SNAPSHOT` Helm chart version as long as it is unreleased. Pushes to main trigger a helm release on the next channel, and this PR implements a hot replacement of the version with `x.x.x-<datetime>-<git sha>`.

A check was added to the stable release workflow which verifies that the helm chart version (which should be manually updated before release) and the tag name match.

A problem was spotted with the cache that's being used for the `helm` and `yq` binaries: since the install uses a script, we need to perform a checkout in case we get a cache miss. It wasn't the case in the next workflow, which led to some failures.
The install being very fast (checkout + install helm + install yq takes about 3 seconds), I removed the cache everywhere in favor of performing the installs each time. This will also help us spot bugs that could be introduced in the install scripts.

I also bumped the yq and helm version numbers to the latest.

Obviously, I could only partially check that these changes were working as expected: https://github.com/adejanovski/k8ssandra/runs/2191981991?check_suite_focus=true#step:8:12

Fixes #557 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -11,7 +11,6 @@ Fixes #<issue number>
 
 **Checklist**
 - [ ] Changes manually tested
-- [ ] Chart versions updated (if necessary)
 - [ ] Automated Tests added/updated
 - [ ] Documentation added/updated
 - [ ] CHANGELOG.md updated (not required for documentation PRs)
--- a/.github/workflows/release-next.yaml
+++ b/.github/workflows/release-next.yaml
@@ -8,24 +8,18 @@ jobs:
   setup:
     runs-on: ubuntu-latest
     env:
-      HELM_VERSION: 3.4.0
-      YQ_VERSION: 3.4.1
+      HELM_VERSION: 3.5.3
+      YQ_VERSION: v4.6.3
     outputs:
       VERSION: ${{ steps.get_version.outputs.VERSION }}
       SHORT_VERSION: ${{ steps.get_short_version.outputs.SHORT_VERSION }}
       HELM_VERSION: ${{ env.HELM_VERSION}}
       YQ_VERSION: ${{ env.YQ_VERSION }}
     steps:
-      - name: Cache bin path
-        id: cache
-        uses: actions/cache@v2
-        with:
-          path: bin
-          key: ${{ runner.os }}-${{ env.HELM_VERSION }}-${{ env.YQ_VERSION }}
+      - name: Checkout
+        uses: actions/checkout@v2
       - run: scripts/install-helm.sh $HELM_VERSION
-        if: steps.cache.outputs.cache-hit != 'true'
       - run: scripts/install-yq.sh $YQ_VERSION
-        if: steps.cache.outputs.cache-hit != 'true'
       - name: Update PATH
         run: |
           echo "$GITHUB_WORKSPACE/bin" >> $GITHUB_PATH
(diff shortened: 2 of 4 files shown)


.github/workflows/release-next.yaml
```diff
@@ -88,19 +87,10 @@ jobs:
...
-          path: ~/bin
-          key: ${{ runner.os }}-${{ needs.setup.outputs.HELM_VERSION }}-${{ needs.setup.outputs.YQ_VERSION }}
       - name: Checkout
         uses: actions/checkout@v2
-        if: steps.cache.outputs.cache-hit != 'true'
-      - run: scripts/install-helm.sh $HELM_VERSION
-        if: steps.cache.outputs.cache-hit != 'true'
-      - run: scripts/install-yq.sh $YQ_VERSION
-        if: steps.cache.outputs.cache-hit != 'true'
+      - run: scripts/install-helm.sh ${{ needs.setup.outputs.HELM_VERSION }}
```
burmanm, line 92: Was there a reason to remove the caches?

adejanovski: Yes, I explained it in the PR description:

>A problem was spotted with the cache that's being used for the helm and yq binaries: since the install uses a script, we need to perform a checkout in case we get a cache miss. It wasn't the case in the next workflow, which led to some failures.
The install being very fast (checkout + install helm + install yq takes about 3 seconds), I removed the cache everywhere in favor of performing the installs each time. This will also help us spot bugs that could be introduced in the install scripts.

This is all debatable of course, but the cache didn't seem to bring any significant improvement and was hiding some errors in the workflows.

.github/pull_request_template.md
```diff
@@ -11,7 +11,6 @@ Fixes #<issue number>
 
 **Checklist**
 - [ ] Changes manually tested
-- [ ] Chart versions updated (if necessary)
```
jdonenine, line 14: Ah, thanks for thinking to update this!

jdonenine (approved):
I took a look through the action logs on the PR, and it looks like the versions are created as expected, so I guess that's about the extent that we can test things out?

Assuming we can't think of another way to test things out I'd like to see us get this moved in sooner rather than later so we've had a chance to see it in action and work out the kinks if need be ahead of trying to release 1.1. 

Although, even that begs the question of how do we observe the stable flow before that?

Also, reading yaml changes is the worst 🤪

Thanks!

.github/workflows/release-next.yaml
```diff
@@ -47,24 +41,29 @@ jobs:
...
         run: |
           echo "$GITHUB_WORKSPACE/bin" >> $GITHUB_PATH
       - name: Update chart dependencies
         run: |
           scripts/update-helm-deps.sh
+      - name: Compute release chart version
+        id: compute_next_version
+        run: |
+          DATE_TIME=$(date '+%Y%m%d%H%M%S')
+          NEXT_VERSION=$(echo $(yq e '.version' charts/k8ssandra/Chart.yaml) | sed "s/-.*/-${DATE_TIME}-${GITHUB_SHA::8}/")
```
jsanda, line 58: @adejanovski and I talked about version number format previously and my biggest concern was that the format we use plays nicely with Semver. We're good to go here 👍

.github/workflows/release-stable.yaml
```diff
@@ -49,24 +41,23 @@ jobs:
...
+      - run: scripts/install-yq.sh ${{ needs.setup.outputs.YQ_VERSION }}
       - name: Update PATH
         run: |
           echo "$GITHUB_WORKSPACE/bin" >> $GITHUB_PATH
       - name: Update chart dependencies
         run: |
           scripts/update-helm-deps.sh
+      - name: Check chart version matches tag
+        run: |
+          RELEASE_VERSION=$(yq e '.version' charts/k8ssandra/Chart.yaml)
```
jsanda, line 56: Nice addition!

jsanda (commented):
The only reason I didn't approve is because @adejanovski is out for an extended weekend. I would prefer him to be available when we pull the trigger in case there any hiccups.

jsanda (approved):

.github/workflows/release-stable.yaml
adejanovski: Thanks

PR #591, opened 2021-03-25
Create webhook secret to ensure cass-operator deployment isn't blocked

### LINKED ISSUE #590: Prepopulate cass-operator-certs-volume for cass-operator
## Bug Report

**Describe the bug**
cass-operator has a validating webhook that runs in process and uses certificates. It gets the certificates from a secret mounted in a volume named `cass-operator-certs-volume`. The cass-operator chart does not currently support deploying the webhook. The cass-operator chart declares the volume but does not create the secret.

Secret and ConfigMap volumes can be marked optional such that if the secret or ConfigMap does not exist, an empty directory is created rather than block pod startup. 

A user running in an EKS 1.16 cluster reported that this was a problem and prevented cass-operator from starting. 

Initially I suggested that we remove the volume from the deployment since we are neither using the validating webhook nor creating the secret. This is not an option though. cass-operator fails at startup with this error message:

```
Secrets volume not found, unable to start webhook
```

When the admission webhook is disabled we should create an empty secret.

**To Reproduce**
Note that this probably needs to be tested in EKS. I wasn't able to produce in my environment.

1. Install k8ssandra, e.g., `helm install volume-test k8ssandra/k8ssandra`
2. Wait for the cass-operator pod to initialize
3. Do `kubectl describe pod <cass-operator pod>`
4. There should be an event with a reason of `FailedMount` and the message should mention `cass-operator-certs-volume`.

**Expected behavior**
cass-operator should deploy. We should create an empty secret when the admission webhook is disabled.

**Screenshots**

* Helm charts version info
1.0.0

1.16

* Kubernetes cluster kind:
EKS

**Additional context**

jsanda:
**What this PR does**:

Fixes #590

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/charts/cass-operator/templates/deployment.yaml
+++ b/charts/cass-operator/templates/deployment.yaml
@@ -30,11 +30,9 @@ spec:
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           volumeMounts:
-            {{- if .Values.admissionWebhooks.enabled }}
             - mountPath: /tmp/k8s-webhook-server/serving-certs
               name: cass-operator-certs-volume
               readOnly: false
-            {{- end }}
             - mountPath: /tmp/
               name: tmpconfig-volume
               readOnly: false
@@ -96,4 +94,4 @@ spec:
             medium: "Memory"
         - name: cass-operator-certs-volume
           secret:
-            secretName: cass-operator-webhook-config
+            secretName: {{ .Release.Name }}-webhook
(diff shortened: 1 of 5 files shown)


charts/cass-operator/templates/deployment.yaml
```diff
@@ -30,11 +30,9 @@ spec:
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           volumeMounts:
-            {{- if .Values.admissionWebhooks.enabled }}
```
jeffbanks, line 33: Will we be using this for other admissionWebhook enablement?
`.Values.admissionWebhooks.enabled`

jeffbanks: Thinking if not, we can remove it from values.yaml, looks like this may be only place we use it.

https://github.com/k8ssandra/k8ssandra/blob/0ff888bcfa27a66be8876f7214f207ed9d4a4a75/charts/cass-operator/values.yaml

jsanda: There are no other webhooks at this time. cass-operator expects the volume mount to exist even if the webhook is disabled.

jeffbanks (approved):
LGTM

emerkle826 (approved):
I don't have EKS setup for testing, but the changes look good to me.

PR #624, opened 2021-04-01
[#623] wait for CassandraDatacenters to be gone before exiting

### LINKED ISSUE #623: Uninstalling can leave CassandraDatacenter behind that cannot be deleted
## Bug Report

**Describe the bug**
This is a follow up to #60. In short, when you run `helm uninstall` the deletion of objects is unordered. We have a problem if cass-operator is deleted before the CassandraDatacenter because it adds a finalizer to the CassandraDatacenter. Kubernetes will block the deletion of an object until all of its finalizers are cleared.

With #60 we added a pre-delete hook that runs job to delete the CassandraDatacenter. This runs before Helm deletes anything. There is a race condition though. The job makes the call to delete the CassandraDatacenter and then exits. When that deletion request is successful, the state of the CassandraDatacenter is updated but it is not yet fully deleted. At this point Helm is free to start deleting objects that were installed with the release. Sometimes the cass-operator Deployment gets deleted before it clears the finalizer.

The cleaner job needs to wait until the CassandraDatacenter is terminated and gone before exiting.

**To Reproduce**
1. Install k8ssandra, preferably with multiple C* nodes
2. Wait for all pods to reach the ready state
3. Run `helm uninstall`
4. Check to see if the Cassnadra pods' statuses change to `TERMINATING`. If they do not, you hit the bug. Due to the race condition this error does not happen consistently.

**Expected behavior**
`helm uninstall` should always properly clean up and remove the CassandraDatacenter.

**Screenshots**

* Helm charts version info
1.0.0

* Kubernetes cluster kind:

**Additional context**

jsanda:
**What this PR does**:
Updates the cleaner job to block until CassandraDatacenters installed in the release are gone.

Fixes #623

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ When cutting a new release of the parent `k8ssandra` chart update the `main / un
 * [BUGFIX] #475 Fix Cassandra config clobbering when enabling Medusa
 * [BUGIFX] #590 Create cass-operator webhook secret
 * [BUGFIX] #602 Fix indentation error in example backup-restore-values.yaml
+* [BUGFIX] #623 `helm uninstall` can leave CassandraDatacenter behind
 * [ENHANCEMENT] #547 Add support for additionalSeeds in the CassandraDatacenter
 * [ENHANCEMENT] #606 Support installation of operators only, disabling the Cassandra cluster creation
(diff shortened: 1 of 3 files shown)


jeffbanks (approved):
Look good!

PR #634, opened 2021-04-07
shutdown cluster by default

### LINKED ISSUE #611: In-place restores should shutdown the cluster by default
In-place restores are done via rolling restart by default. If the cluster remains available and continues accepting writes while the restore is in process, we cannot guarantee that the cluster will restored to the backup state. 

Furthermore, there is a possibility that writes could be lost. During the restore process Medusa deletes the contents of the data subdirectory of the table being restored. This could easily lead to data loss.

Shut down the entire cluster by default to be proactive in preventing data loss or inconsistent state during a restore.

It is worth noting that shutting down the cluster won't completely prevent these problems. As nodes come back online, clients could still perform queries before the restore is complete. Completely preventing client access is beyond the scope of this issue and might be something we can better address when we do service mesh integration.

**Additional context**
Restores are currently are really, really slow. Part of that is due to #570 and #571. 

The other reason is that the restore operation is done serially one Cassandra pod at a time. The restore operation is performed by the `medusa-restore` init container. The bulk of the time of the restore operation will be from downloading the backup files. Doing it serially one node at a time will make it painfully slow and could result in result in long periods of downtime for the cluster. 

The reason it is done serially is because the restore happens by way of a StatefulSet update. The StatefulSet controller does a rolling update of pods, one at a time. I think we can work around this relatively easily. The StatefulSet's `.spec.podManagementPolicy` property controls how pods are created. The default value is `OrderedReady` which means one pod at a time. cass-operator uses `Parallel`. If we simply delete the StatefulSet(s) and let them get recreated, the `medusa-restore` init containers will run in parallel. Note that deleting a Statefulset does not cascade to the PersistentVolumeClaims and PersistentVolumes that it uses.

jsanda:
**What this PR does**:

Fixes #611
**Checklist**
- [x] Changes manually tested

### DIFF
--- a/charts/restore/values.yaml
+++ b/charts/restore/values.yaml
@@ -11,13 +11,20 @@ cassandraDatacenter:
   # restored
   name: dc1
 
-# -- In-place restore will restore the backup to the source cluster. Note that
-# this will trigger a rolling restart of the cluster.
+# -- In-place restore will restore the backup to the source cluster.
 inPlace: true
 
 # -- When true will shutdown the entire Cassandra cluster. The underlying
 # StatefulSets are scaled down to zero. Persistent volumes remain intact. If the
-# backup includes schema changes like dropping a table, then set this to true;
-# otherwise, the changes will be lost via gossip from nodes that have not yet
-# been restored.
-shutdown: false
+# backup includes schema changes like dropping a table, then must be set to
+# true; otherwise, the changes will be lost via gossip from nodes that have
+# not yet been restored. It is recommended in general to shutdown the cluster
+# prior to a restore to avoid data inconsistencies and/or data loss that could
+# happen with clients writing to the cluster while the restore operation is in
+# progress.
+#
+# When set the cluster is shutdown the restore operations happen in parallel
+# across all Cassandra pods. If `shutdown` is `false` the restore operation is
+# done via a rolling restart where the restore operation runs on each pod
+# serially.
+shutdown: true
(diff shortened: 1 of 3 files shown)


charts/restore/values.yaml
```diff
@@ -11,13 +11,20 @@ cassandraDatacenter:
...
-# been restored.
-shutdown: false
+# backup includes schema changes like dropping a table, then must be set to
+# true; otherwise, the changes will be lost via gossip from nodes that have
+# not yet been restored. It is recommended in general to shutdown the cluster
+# prior to a restore to avoid data inconsistencies and/or data loss that could
+# happen with clients writing to the cluster while the restore operation is in
+# progress.
+#
+# When set the cluster is shutdown the restore operations happen in parallel
```
emerkle826, line 26: nit: add the word `and`

_suggested change_:
```
When set the cluster is shutdown and the restore operations happen in parallel
```

emerkle826 (approved):
small nitpick if you want to address it.

charts/restore/values.yaml
jsanda: Yup, I want to address it :)

burmanm:
Is this missing a file in the commit? There's no changes to the backup.yaml template

jsanda:
> Is this missing a file in the commit? There's no changes to the backup.yaml template

It's there now. The only change is in charts/restore/values.yaml.

PR #642, opened 2021-04-08
Make the stable release workflow dependent on the integration tests one

### LINKED ISSUE #632: Make the GHA release workflow to require integration tests to pass
Currently the integration tests won't run on the release workflow. If any issue was introduced late in the process we could cut out a release that is broken.
We need to add the integration tests to the release workflow and have the release steps require them to pass.

adejanovski:
**What this PR does**:
Makes the stable release workflow depend upon the successful completion of the "Integration Tests" one.

The release workflow is now triggered by a `workflow_run` event on tags, and the integration tests workflow also triggers on tags.
Using workflow runs as trigger events is described in [this documentation](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run) and its usage is discussed [in this thread](https://github.community/t/workflow-run-completed-event-triggered-by-failed-workflow/128001/2).

Fixes #632

**Checklist**
- [x] Changes manually tested


.github/workflows/integration-tests.yaml
```diff
@@ -16,14 +16,17 @@ on:
...
+      - 'v*'
 jobs:
   integration-tests:
     runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
       matrix:
         cassandra_version: ["3.11.10", "4.0.0"]
-        scenario: ["TestMedusaDeploymentScenario/\"S3\"", "TestMedusaDeploymentScenario/\"Minio\"", "TestReaperDeploymentScenario", "TestMonitoringDeploymentScenario", "TestStargateDeploymentScenario"]
+        #scenario: ["TestMedusaDeploymentScenario/\"S3\"", "TestMedusaDeploymentScenario/\"Minio\"", "TestReaperDeploymentScenario", "TestMonitoringDeploymentScenario", "TestStargateDeploymentScenario"]
```
burmanm, line 28: Is this to ensure that the flakey tests (Stargate and Reaper) do not fail the release?

adejanovski: No, I'm just trying to get the "Integration Tests" workflow pass as fast as possible to see if the stable release workflow triggers as expected.
So far it's not, but I'll add back all test scenarios before the PR is ready for review.

adejanovski:
@k8ssandra/maintainers, the PR is ready for review.
I've merged the integration-tests and release-stable/next workflows into a single one because chaining workflows is clunky in GHA and this gives better control.
It also allowed to reduce code duplication between workflows.

One thing that's changed here also is that the GH release isn't created by the workflow (which created quite a lot of garbage) but the GH release has to be published for the workflow to run. It protects us from tags pushed by mistake when cutting out releases.

I think it would be good to go as far as integrating the pr-checks (Helm linting + unit tests) to this workflow and condition the integration tests on the pr-checks success. Thoughts?

.github/workflows/integration-tests-release.yaml
```diff
@@ -0,0 +1,219 @@
...
+      - 'pkg/**'
+      - 'cmd/**'
+      - 'charts/**'
+      - 'tests/integration/**'
+      - '.github/workflows/**'
+  push:
+    branches:
+      - 'main'
+      - '**-run-ci'
+      - '*.*'
```
burmanm, line 16: Have we decided that tag-branch naming convention?

.github/workflows/integration-tests-release.yaml
```diff
@@ -0,0 +1,219 @@
...
+    paths:
+      - 'pkg/**'
+      - 'cmd/**'
+      - 'charts/**'
+      - 'tests/integration/**'
+      - '.github/workflows/**'
+  push:
+    branches:
+      - 'main'
+      - '**-run-ci'
```
burmanm, line 15: Didn't we opt to not creating any non-fork branches for features? What are these for?

.github/workflows/integration-tests-release.yaml
```diff
@@ -0,0 +1,219 @@
...
+      - '*.*'
+    paths:
+      - 'pkg/**'
+      - 'cmd/**'
+      - 'charts/**'
+      - 'tests/integration/**'
+      - '.github/workflows/**'
+  release:
```
burmanm, line 23: Are we using release publishing from Github instead of just pushing a tag?

.github/workflows/integration-tests-release.yaml
```diff
@@ -0,0 +1,219 @@
...
+      - name: Install Dependencies
+        run: |
+          set -x
+          sudo apt-get update || echo "something went slightly wrong with apt-get update"
+
+          # install kubectl
+          curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+          sudo install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
+
+          # install kind
```
burmanm, line 58: Other repos use kind script + helm script to install those dependencies.

.github/workflows/integration-tests-release.yaml
```diff
@@ -0,0 +1,219 @@
...
+      - name: Setup tmate session
+        if: "failure() && contains(github.event.head_commit.message, '[SSH]')"
+        uses: mxschmitt/action-tmate@v3
+        timeout-minutes: 10
+  
+  release:
+    if: |
+      ${{ (( github.event_name == 'release' && github.event.action == 'published' ) 
+        || ( github.event_name == 'push' && github.ref == 'refs/heads/main' )) }}
+    runs-on: ubuntu-latest
```
burmanm, line 120: This one uses -latest, the first one used -20.04. I think both should be -latest.

.github/workflows/integration-tests-release.yaml
```diff
@@ -0,0 +1,219 @@
...
+      - name: Checkout
+        uses: actions/checkout@v2
+      
+      - name: Update PATH
+        run: |
+          echo "$GITHUB_WORKSPACE/bin" >> $GITHUB_PATH
+
+      - name: Install and validate tool versions
+        run: |
+          scripts/install-helm.sh $HELM_VERSION
```
burmanm, line 136: Yet another different way of installing helm? Why not the same as integration-tests? And same for yq. Why the definition of version here, but not previously? It could in theory cause different behaviour.

.github/workflows/integration-tests-release.yaml
```diff
@@ -0,0 +1,219 @@
...
+      - name: Generate package archives
+        run: |
+          ls | while read c; do
+            if [[ -d $c ]]; then
+              helm package $c
+            fi
+          done
+        working-directory: build
+
+      - name: Check if Helm release is possible
```
burmanm, line 173: Shouldn't this just be in the original if clause? That the whole job is skipped if this isn't true?

burmanm (changes requested):
Can't modify single comments to review parts, so empty comment here.

.github/workflows/integration-tests-release.yaml
adejanovski: yes, we cut out releases from release branches named in the form: `major.minor`

.github/workflows/integration-tests-release.yaml
adejanovski: This allows the opt-in approach to running CI when pushing commits without a PR. That's what we agreed upon.

.github/workflows/integration-tests-release.yaml
adejanovski: yes, I've explained this in the PR description. Otherwise we tend to forget to publish the GH releases and update the description there.

.github/workflows/integration-tests-release.yaml
adejanovski: Which scripts are these? Helm is installed using `./scripts/install-helm.sh` but I don't see an equivalent for kind.

.github/workflows/integration-tests-release.yaml
adejanovski: agreed 👍

.github/workflows/integration-tests-release.yaml
adejanovski: I kept it the way it was done in the release workflow, but agree helm should be installed with the same version (we're already using the same script).

Here we're installing `yq`, but the integration tests are installing `jq` 😉  Actually, this could be a leftover because I'm not sure what we need `jq` for 🤔  I'll remove it and see if something breaks.

.github/workflows/integration-tests-release.yaml
adejanovski: I'm not sure I can put this in an `if:` clause as is actually. I'll check how I can make it so that the whole job is skipped, it's cleaner indeed.

.github/workflows/integration-tests-release.yaml
burmanm:
```yaml
      - name: Install Helm
        uses: azure/setup-helm@v1
        with:
          version: v3.5.3
      - name: Create Kind Cluster
        uses: helm/kind-action@v1.1.0
        with:
          version: v0.10.0
          node_image: kindest/node:v1.20.2
          cluster_name: kind
          config: tests/testdata/kind/kind_config_6_workers.yaml
```

.github/workflows/integration-tests-release.yaml
burmanm: But that was for forks only, not k8ssandra-repo?

adejanovski: I'm not sure how to make that distinction

.github/workflows/integration-tests-release.yaml
adejanovski: fixed

.github/workflows/integration-tests-release.yaml
adejanovski: I've made the changes to use your recommended actions instead of using bash to install helm and create the kind clusters.

.github/workflows/integration-tests-release.yaml
adejanovski: I made it so that the release job only runs if we're not on a fork.

adejanovski:
@burmanm, I made the requested changes, you can now take another pass on the PR.
Thanks

burmanm (approved):

PR #638, opened 2021-04-07
Update Management API image locations

emerkle826:
**What this PR does**:
Updates the Management API image locations.

**Which issue(s) this PR fixes**:
Fixes #637

**Checklist**
- [x] Changes manually tested
- [x] Automated Tests added/updated
- [ ] Documentation added/updated
- [x] CHANGELOG.md updated (not required for documentation PRs)
- [x] CLA Signed:  [DataStax CLA](https://cla.datastax.com/)

PR #647, opened 2021-04-08
Upgrade to Medusa 0.10.0 and add IT for scale up after restore issue

### LINKED ISSUE #616: Scale up fails if a restore was performed in the past
## Bug Report

**Describe the bug**
Scaling up the Cassandra nodes after having restored a backup will make the new pod loop crash as it tries to perform the restore as well.

Even after having deleted the cassandrabackup/cassandrarestore CRDs, it will still start the init container with a restore key and a backup name for the medusa-restore init container:

```
  medusa-restore:
    Container ID:   containerd://8e2d71cdeb931b43568bd2bd714f700b3c8f397cab3f3f713720a9518d38bd65
    Image:          docker.io/jsanda/medusa:grpc-sigterm-13d70bb2
    Image ID:       docker.io/jsanda/medusa@sha256:2095ea34de678178ff0798b8045390947e1963d6dd376b2669f255d74c822786
    Port:           <none>
    Host Port:      <none>
    State:          Terminated
      Reason:       Error
      Exit Code:    1
      Started:      Thu, 01 Apr 2021 11:14:03 +0200
      Finished:     Thu, 01 Apr 2021 11:14:04 +0200
    Last State:     Terminated
      Reason:       Error
      Exit Code:    1
      Started:      Thu, 01 Apr 2021 11:13:45 +0200
      Finished:     Thu, 01 Apr 2021 11:13:47 +0200
    Ready:          False
    Restart Count:  2
    Environment:
      MEDUSA_MODE:   RESTORE
      CQL_USERNAME:  <set to the key 'username' in secret 'k8ssandra-medusa'>  Optional: false
      CQL_PASSWORD:  <set to the key 'password' in secret 'k8ssandra-medusa'>  Optional: false
      BACKUP_NAME:   backup1
      RESTORE_KEY:   4c37944c-82bd-4342-a016-2b2544884afe
```

**To Reproduce**
1. Create a cluster with a single Cassandra node
2. Create a backup
3. Restore the backup
4. After the restore is finished, scale up to 2 Cassandra nodes
5. The newly created pod will go into a crash loop and will fail to initialize

**Expected behavior**
Once a restore is finished, it should not be scheduled on nodes that were not part of the initial backup.
Scaling up should be possible without a restore being attempted.

* Helm charts version info

`1.1.0-SNAPSHOT` 

* Kubernetes cluster kind:

kind

adejanovski:
**What this PR does**:

Fixes #616 

**Checklist**
- [x] Changes manually tested


tests/integration/steps/integration_steps.go
```diff
@@ -133,43 +132,70 @@ func DeployClusterWithValues(t *testing.T, namespace, options, customValues stri
...
 }
 
-func DeployClusterWithValuesAndHeapSettings(t *testing.T, namespace, options, cassandraHeap, stargateHeap, customValues string) {
-	log.Printf("I can deploy a cluster with %s options, %s Cassandra Heap and %s Stargate heap using the %s values", options, cassandraHeap, stargateHeap, customValues)
-
-	splitOptions := strings.Split(options, "-")
-	medusaEnabled := "true"
-	reaperEnabled := "true"
-	monitoringEnabled := "true"
+func ScaleClusterWithValues(t *testing.T, namespace, options, customValues string, nodes int) {
```
jsanda, line 140: `ScaleClusterWithValues` looks identical to `DeployClusterWithValues` except for the last line where it calls `scaleCluster` instead of `scaleCluster`. Maybe we could refactor the duplicate code?

tests/integration/steps/integration_steps.go
```diff
@@ -133,43 +132,70 @@ func DeployClusterWithValues(t *testing.T, namespace, options, customValues stri
...
+		}
 	}
+	helmValues["cassandra.datacenters[0].size"] = strconv.Itoa(nodes)
+	helmValues["cassandra.datacenters[0].name"] = datacenterName
+	scaleCluster(t, namespace, customValues, helmValues)
+}
 
-	if Find(splitOptions, "nomonitoring") {
-		monitoringEnabled = "false"
+func scaleCluster(t *testing.T, namespace, customValues string, helmValues map[string]string) {
```
jsanda, line 160: I find the name `scaleCluster` misleading. I say that because I looked over the function and it looks like it is doing a `helm upgrade` and then blocking until the CassandraDatacenter is ready. If my observation is correct, could we rename it to make it more clear?

tests/integration/integration_suite_test.go
```diff
@@ -191,8 +191,9 @@ func TestMedusaDeploymentScenario(t *testing.T) {
 			namespace := initializeCluster(t)
 			medusaSuccess := t.Run("Test backup and restore", func(t *testing.T) {
 				createMedusaSecretAndInstallDeps(t, namespace, backend)
-				deployClusterForMedusa(t, namespace, backend)
+				deployClusterForMedusa(t, namespace, backend, 1)
 				testMedusa(t, namespace, backend, backupName)
+				scaleUpCassandra(t, namespace, backend, 2)
```
jsanda, line 196: Do we need to test the scale up for each backend? The bug that prevented scale up is not impacted by the backend used. On that basis I wonder if it makes sense to have a separate test that focuses on scale up/down with backup/restore.

jsanda (changes requested):

tests/integration/steps/integration_steps.go
adejanovski: What would you name it instead?

tests/integration/steps/integration_steps.go
adejanovski: I can probably remove some code duplication indeed.

tests/integration/integration_suite_test.go
adejanovski: I can make a different test suite with just MinIO, but I'm not sure it's worth the pain of creating yet another test suite. It just takes a few more minutes to complete with both backends (although S3 is mostly not running due to the use of a secret).
As long as we test this behavior, I think it's fine to run it each time we have a backup test.

burmanm: I think John meant that we shouldn't think "backup / restore" as the main thing in that scaling test, but the actual scale up / scale down. The backup restore would just be a minor detail / single subtest in that process. 

The problem with "few minutes more" is that they accumulate and soon it's hours to see if a PR passes at which point no one will give a damn if it passes or not, especially if it even once fails without the PR being the reason.

jsanda: >I think John meant that we shouldn't think "backup / restore" as the main thing in that scaling test, but the actual scale up / scale down. The backup restore would just be a minor detail / single subtest in that process.

I do think we want to implement separate scaling tests and backup/restore should be covered in those tests, but I was also thinking about execution times.

>I can make a different test suite with just MinIO, but I'm not sure it's worth the pain of creating yet another test suite. It just takes a few more minutes to complete with both backends (although S3 is mostly not running due to the use of a secret).
As long as we test this behavior, I think it's fine to run it each time we have a backup test.

I looked at the results for [TestMedusaDeploymentScenario/"Minio"] (https://github.com/k8ssandra/k8ssandra/pull/647/checks?check_run_id=2298520333) in this PR. The test took 721 seconds. 

Then I looked at [TestMedusaDeploymentScenario/"Minio"](https://github.com/k8ssandra/k8ssandra/pull/638/checks?check_run_id=2300598089) for #638. It took 342 seconds.

That is a difference of over 6 minutes. I am sure that there will be a decent bit of variation if we were to consider more samples, but I am confident it is still going to be a very appreciable difference. Because scaling up is slow, that difference is enough for me to disagree that it is fine to run it across each backup/restore test. Also consider this. Right now we are only testing again MinIO and S3, but we will be adding GCS soon and then hopefully Azure after that.

Here is my thinking. We have the additional test coverage which is great. Thanks for that @adejanovski. We want to get a release out asap. The test changes would be non-trivial, and I don't see any reason to rush them. I propose we merge the PR and address test refactoring post 1.1. 

@burmanm and @adejanovski does that sound reasonable?

jdonenine: Just a couple of thoughts on things here:

> Here is my thinking. We have the additional test coverage which is great. Thanks for that @adejanovski. We want to get a release out asap. The test changes would be non-trivial, and I don't see any reason to rush them. I propose we merge the PR and address test refactoring post 1.1.

Totally agree with this, the big benefit is the additional testing, and at the moment, that's a really big benefit because it gets us to a higher level of confidence to release 1.1.

Sounds like the refactoring is much about the speed of the tests - wouldn't we, in theory, be capable of running the different backend (s3, minio, etc) tests in parallel - so even if each did include the scale up/down we could maybe reduce the overall impact that way?  Perhaps we're limited there by GH's runners though?

adejanovski: > I think John meant that we shouldn't think "backup / restore" as the main thing in that scaling test, but the actual scale up / scale down. The backup restore would just be a minor detail / single subtest in that process.

> I do think we want to implement separate scaling tests and backup/restore should be covered in those tests, but I was also thinking about execution times.

I understand what you're both saying but for this particular case I just wanted to test that scaling up is still possible after a restore.
We'll need separate, proper scaling tests to be implemented, which was not the intent in this PR.

We could restrict the scaling test to just a specific backend, but overall it's going to take the same amount of time as our checks will take as long as the longest of all of them (they're running in parallel, so having a single one running the scale up test won't save time overall). 
We also have other leads to reduce the overall duration, including tuning the probes. I've seen Cassandra pods being restarted multiple times during restores instead of just once, which make the overall operation last way longer.

> Here is my thinking. We have the additional test coverage which is great. Thanks for that @adejanovski. We want to get a release out asap. The test changes would be non-trivial, and I don't see any reason to rush them. I propose we merge the PR and address test refactoring post 1.1.

I agree 👍 

> Sounds like the refactoring is much about the speed of the tests - wouldn't we, in theory, be capable of running the different backend (s3, minio, etc) tests in parallel - so even if each did include the scale up/down we could maybe reduce the overall impact that way? Perhaps we're limited there by GH's runners though?

It's the case already, we're running the tests in parallel, and have enough GH runners so that they all run at the same time.

I agree with Michael and John on trying to avoid making tests last longer, and I hope we can focus on speeding things up in the coming sprints while still allowing to add test cases. We'll need to do some profiling on K8ssandra deployments to see where time is spent and tune appropriately (which John already started doing).

tests/integration/steps/integration_steps.go
adejanovski: Since most of the code is exactly the same, I deleted the `ScaleCluster***` functions and added an `upgrade` boolean as argument for `DeployCluster***` to distinguish running `helm install` and `helm upgrade`.

Good recommendation 👍

tests/integration/steps/integration_steps.go
adejanovski: It's removed now and fully integrated into the existing `DeployCluster***` functions.

adejanovski:
@jsanda, the PR is ready for another review.

tests/integration/integration_suite_test.go
jsanda: >Sounds like the refactoring is much about the speed of the tests - wouldn't we, in theory, be capable of running the different backend (s3, minio, etc) tests in parallel - so even if each did include the scale up/down we could maybe reduce the overall impact that way? Perhaps we're limited there by GH's runners though?

@jdonenine We can run the tests in parallel which does mitigate the impact; however, it means those runners are tied up longer before they can move onto the next tests. There is another aspect of execution that seems to be getting lost in the discussion and that is running tests locally. As a developer I want to be able to also run the e2e tests locally. For the local development scenario I likely wouldn't run those tests in parallel.

jsanda (approved):
@adejanovski I think we're good to go. Thanks!

PR #731, opened 2021-05-05
Update CRDs on pre-upgrade phase

### LINKED ISSUE #110: Upgrading from versions 0.12.0 and earlier of the k8ssandra and k8ssandra-cluster charts fails
## Bug Report

**Describe the bug**

**To Reproduce**

1. Make sure you have the latest charts:

```
$ helm repo update
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "datastax" chart repository
...Successfully got an update from the "k8ssandra" chart repository
...Successfully got an update from the "incubator" chart repository
...Successfully got an update from the "prometheus-community" chart repository
...Successfully got an update from the "traefik" chart repository
...Successfully got an update from the "bitnami" chart repository
...Successfully got an update from the "stable" chart repository
Update Complete. ⎈ Happy Helming!⎈

$ helm search repo k8ssandra
NAME                       	CHART VERSION	APP VERSION	DESCRIPTION
k8ssandra/k8ssandra        	0.14.0       	3.11.7     	Configures and provisions the full k8ssandra stack
k8ssandra/k8ssandra-cluster	0.14.0       	3.11.7     	Configures and creates a k8ssandra cluster
k8ssandra/backup           	0.14.0       	0.1.0      	Creates a CassandraBackup
k8ssandra/restore          	0.14.0       	0.1.0      	Creates a CassandraRestore
```

1. `helm install k8ssandra k8ssandra/k8ssandra --version 0.12.0`
1. `helm install upgrade-test k8ssandra/k8ssandra-cluster --version 0.12.0`
1. `helm upgrade k8ssandra k8ssandra/k8ssandra`
1. Upgrading `upgrade-test` fails:

```
$ helm upgrade upgrade-test k8ssandra/k8ssandra-cluster
Error: UPGRADE FAILED: cannot patch "upgrade-test-reaper-k8ssandra" with kind Reaper: Reaper.reaper.cassandra-reaper.io "upgrade-test-reaper-k8ssandra" is invalid: [spec.serverConfig.cassandraBackend.cassandraService: Required value, spec.serverConfig.cassandraBackend.clusterName: Required value]
```

**Expected behavior**
`helm upgrade` should work without error.

`$ kubectl version
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.9", GitCommit:"4fb7ed12476d57b8437ada90b4f93b17ffaeed99", GitTreeState:"clean", BuildDate:"2020-07-15T16:18:16Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"17+", GitVersion:"v1.17.13-gke.600", GitCommit:"3415f17fc893455f55dd732b92b7036a59609135", GitTreeState:"clean", BuildDate:"2020-10-20T23:58:27Z", GoVersion:"go1.13.15b4", Compiler:"gc", Platform:"linux/amd64"}`

**Additional context**
Reaper CRD changes were made for #45 which were introduced in version 0.13.0 or 0.14.0 of the charts. The errr above says that the `spec.serverConfig.cassandraBackend.clusterName` field is required; however, it was removed from the CRD. 

The real problem is that Helm does not add/update CRDs with `helm upgrade`. See this [ticket](https://github.com/helm/helm/issues/6581) for details.

We need to implement a [hook](https://helm.sh/docs/topics/charts_hooks/) to ensure any CRD changes are applied.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-76) by [Unito](https://www.unito.io)

burmanm:
**What this PR does**:
When release is upgraded, update the CRDs to a new version before updating other resources. The test does not work with SNAPSHOT versions at this point however..

Fixes #110

**Checklist**
- [x] Changes manually tested


charts/k8ssandra/values.yaml
```diff
@@ -588,6 +588,11 @@ monitoring:
 cleaner:
   image: k8ssandra/k8ssandra-cleaner:e6c3702701ca
 
+# k8ssandra-client provides CLI utilities, but also certain functions such as 
+# upgradecrds that allow modifying the running instances
+client:
+  image: burmanm/k8ssandra-client:8772293f
```
jsanda, line 594: I am not sure how I feel about pulling in k8ssandra-client in this PR. I have only looked at it briefly. It looks like a lot of really good stuff. With that said, I have a couple concerns. First, it is scope creep. Secondly and more importantly, I think there ought to be some discussion, design docs, etc. on what we expect for a CLI client before we go ahead and introduce one into the project. Clearly you are in the best position to drive that effort 😄

burmanm:
Freaking git.

jsanda (commented):
While I may not agree with everything in this [doc](https://github.com/helm/community/blob/f9e06c16d89ccea1bea77c01a6a96ae3b309f823/architecture/crds.md) it is worth a read. It does make me wonder whether we should be handling CRD updates implicitly. 

Would it be better/safer to put CRDs into a separate chart (storing them in the templates directory)? Doing so would make upgrading them an explicit action

charts/k8ssandra/templates/crd/role.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ .Release.Name }}-crd-upgrader-k8ssandra
+  labels: {{ include "k8ssandra.labels" . | indent 4 }}
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+rules:
```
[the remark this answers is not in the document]
burmanm: I only added those which I've seen modifications in lately. But I guess we could add all possible ones.

burmanm:
I don't think storing CRDs in a separate template would make the issue any easier. In the end, my goal would be that this is user-friendly, thus they wouldn't require any knowledge if they should apply some flag or not. If we advertise a new feature, that feature should work once you upgrade, not "if you upgrade and you do this and that and maybe some unicorns also". 

But we indeed should be extra careful that we don't (when using the same v1beta1) introduce any backwards incompatible changes. If we do that, we should update the version of the CRD in that case (and then consider the consequences).

charts/k8ssandra/templates/crd/batch_job.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ .Release.Name }}-crd-upgrader-job-k8ssandra
+  labels: {{ include "k8ssandra.labels" . | indent 4 }}
+  annotations:
+    "helm.sh/hook": pre-upgrade
+    "helm.sh/hook-delete-policy": hook-succeeded
```
jsanda, line 8: It might be good to make the deletion policy configurable and consider making the default `before-hook-creation`  so that if things do go sideways the user will have logs.

charts/k8ssandra/templates/crd/batch_job.yaml
```diff
@@ -0,0 +1,30 @@
...
+          imagePullPolicy: IfNotPresent
+          env:
+            - name: POD_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+          args:
+            - -upgradecrds
+            - --targetVersion
+            - {{ .Chart.Version }}
```
jsanda, line 30: We have a bit of a chicken and egg problem here. I checked out your branch to test and upgraded from my local chart on the file system. That chart version is not installed. The latest version in the Helm repo is `1.2.0-20210514022645-da7547a5`. I temporarily changed the version to that so that the CRDs would be downloaded.

pkg/crds/upgrade.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	return &Upgrader{
+		client:    c,
+		namespace: namespace,
+	}, nil
+}
+
+// Upgrade installs the missing CRDs or updates them if they exists already
+func (u *Upgrader) Upgrade(targetVersion string) error {
```
jsanda, line 56: It would be nice to have some debug logging in here. I mentioned in my previous comment how I hard coded the value of `--targetversion`. I did the upgrade and as a quick test did `kubectl get crd cassandradatacenters.cassandra.datastax.com -o yaml | grep configSecret`. I did not get any results back. Then I uninstalled, deleted CRDs, and reinstalled with `helm install test k8ssandra/k8ssandra --version 1.2.0-20210514022645-da7547a5`. This time when I the preceding grep I get back results as expected.

jsanda (changes requested):

pkg/crds/upgrade_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	// +kubebuilder:scaffold:imports
+)
+
+const (
+	UpgradeTestNamespace = "k8ssandra-upgrade-test"
+)
+
+func TestAFunctionality(t *testing.T) {
```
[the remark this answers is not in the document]
burmanm: envtests never deploy RBACs, the rest.Config probably runs as cluster admin.

charts/k8ssandra/templates/crd/role.yaml
```diff
@@ -0,0 +1,32 @@
...
+  - apiGroups:
```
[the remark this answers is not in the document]
burmanm: Kubernetes RBAC docs say: 

```
User-facing ClusterRoles use ClusterRole aggregation to allow admins to include rules for custom resources on these ClusterRoles. To add rules to the admin, edit, or view roles, create a ClusterRole with one or more of the following labels:
```

For those labels.

pkg/crds/upgrade.go
```diff
@@ -0,0 +1,190 @@
...
+		if gvk.Kind != "CustomResourceDefinition" {
+			return nil
+		}
+
+		*crds = append(*crds, crd)
+		// }
+
+		// reader := k8syaml.NewYAMLReader(bufio.NewReader(bytes.NewReader(b)))
+
+		_ = k8syaml.NewYAMLOrJSONDecoder(bytes.NewReader(b), 4096)
```
jsanda, line 161: What is the purpose of this line?

pkg/crds/upgrade_test.go
```diff
@@ -0,0 +1,107 @@
...
+	objs = []runtime.Object{}
+	for _, crd := range crds {
+		objs = append(objs, &crd)
+	}
+
+	envtest.WaitForCRDs(cfg, objs, testOptions)
+
+	By("Upgrading to 1.2.0-20210514022645-da7547a5")
+	crds, err = u.Upgrade("1.2.0-20210514022645-da7547a5")
+	g.Expect(err).Should(Succeed())
```
jsanda, line 92: It would be good to add an additional verification step that fetches the CassandraDatacenter CRD and check for one of the new properties.

tests/integration/steps/integration_steps.go
```diff
@@ -460,6 +459,38 @@ func DeleteNamespace(t *testing.T, namespace string) {
...
+	err = helm.InstallE(t, helmOptions, "k8ssandra/k8ssandra", releaseName)
+	g(t).Expect(err).To(BeNil())
+
+	// Wait for cass-operator pod to be ready
+	g(t).Eventually(func() bool {
+		return PodWithLabelsIsReady(t, namespace, map[string]string{"app.kubernetes.io/name": "cass-operator"})
+	}, retryTimeout, retryInterval).Should(BeTrue())
+
+	// Wait for CassandraDatacenter to be udpating..
+	WaitForCassDcToBeUpdating(t, namespace)
```
jsanda, line 488: Do we need this? Isn't it sufficient to wait for the datacenter to become ready?

jsanda (changes requested):

.github/workflows/publish-tools.yaml
```diff
@@ -0,0 +1,45 @@
+---
```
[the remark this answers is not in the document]
burmanm: None, I thought I did git mv to rename it, but it seems it wasn't deleted for some reason. Deleted now.

pkg/crds/upgrade.go
burmanm: I still need to fix one thing, that is some CRDs could be built with "---" as starting and that messes up.

pkg/crds/upgrade_test.go
burmanm: Done.

tests/integration/steps/integration_steps.go
burmanm: I copied this pattern from cass-operator, I guess it validates that it actually starts the update, not just becomes ready.

charts/k8ssandra/templates/crd/cluster_role.yaml
```diff
@@ -0,0 +1,24 @@
...
+      -  apiextensions.k8s.io
+    resources:
+      - customresourcedefinitions
+    verbs:
+      - create
+      - get
+      - watch
+      - list
+      - update
+      - patch
```
jsanda, line 24: We don't need `patch`. I confirmed in my latest round of testing.

jsanda (changes requested):

jsanda (approved):
Nice work!

PR #736, opened 2021-05-07
Reduce initial delay of Stargate readiness probe

### LINKED ISSUE #654: Reduce initial delay of Stargate readiness probe
The Stargate deployment readiness probe has an initial delay of 4 minutes. This means that the kubelet will wait 4 minutes before checking that the pods are ready. This really slows down startup down. It also means that clients from outside the cluster will not be able to access Stargate during this time.

The Stargate deployment already uses an initContainer to wait for at least one Cassandra pod to become ready. It needs to be updated to wait until the whole C* cluster is ready (see #612 for details). Given the delay from the initContainer we should be able to safely reduce the initial delay substantially, maybe to 10 or 15 seconds. When #612 is fixed, we should be able to further reduce it. 

We are currently **not** setting the failure threshold. We can set that in conjunction with the reduced initial delay.

The initial delay of the liveness probe is also set to 4 minutes. That can probably be safely reduced quite a bit as well, but I don't think it will impact or delay when the pods become ready.

As part of the work for this it would be nice to see some analysis on what the impact is for using different different settings for the readiness probe as well as the liveness probe.

**Additional context**
I want to point out that when we implement controllers for k8ssandra and Stargate, we won't need the initContainer any more. The controller can periodically check on the CassandraDatacenter. When it is ready, it can then deploy Stargate. We already do this in reaper-operator for Reaper.

adutra:
**What this PR does**:

* Modifies Stargate init container to wait not for one, but for all nodes in datacenter 0.
* Modifies the initial delay for Stargate probes

Fixes #654

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/CHANGELOG-1.2.md
+++ b/CHANGELOG-1.2.md
@@ -1,6 +1,7 @@
 # Changelog
 
-Changelog for K8ssandra, new PRs should update the ` unreleased` section with entries in the order:
+Changelog for K8ssandra, new PRs should update the `unreleased` section below with entries sorted by type, in the 
+following order:
 
 ```markdown
 * [CHANGE]
@@ -9,12 +10,17 @@ Changelog for K8ssandra, new PRs should update the ` unreleased` section with en
 * [BUGFIX]
 ```
 
-When cutting a new release of the parent `k8ssandra` chart update the `unreleased` heading to the tag being generated and date `## vX.Y.Z - YYYY-MM-DD` and create a new placeholder section for  `unreleased` entries.
+If two entries have the same type, they should be sorted from newest to oldest (the newest comes first, the oldest comes 
+last).
+
+When cutting a new release of the parent `k8ssandra` chart update the `unreleased` heading to the tag being generated 
+and date `## vX.Y.Z - YYYY-MM-DD` and create a new placeholder section for  `unreleased` entries.
 
 ## unreleased
 
-* [BUGFIX] #678 Upgrade to Medusa 0.10.1 fixing failed backups after a restore
 * [CHANGE] #726 Upgrade to Cassandra 4.0-rc1
+* [ENHANCEMENT] #654 Reduce initial delay of Stargate readiness probe
 * [ENHANCEMENT] #560 Add the ability to attach additional PVs for medusa backups
 * [ENHANCEMENT] #693 Update cass-operator to v1.7.0
 * [ENHANCEMENT] #732 Make allocate_tokens_for_replication_factor configurable
+* [BUGFIX] #678 Upgrade to Medusa 0.10.1 fixing failed backups after a restore
--- a/charts/k8ssandra/templates/stargate/stargate.yaml
+++ b/charts/k8ssandra/templates/stargate/stargate.yaml
@@ -111,13 +111,13 @@ spec:
               path: /checker/liveness
               port: health
             timeoutSeconds: 10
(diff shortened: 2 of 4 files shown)


charts/k8ssandra/templates/stargate/stargate.yaml
```diff
@@ -111,11 +111,13 @@ spec:
               path: /checker/liveness
               port: health
             timeoutSeconds: 10
-            initialDelaySeconds: 240
+            initialDelaySeconds: 30
```
jsanda, line 114: These sound good for defaults, but we should make the probes configurable.

jsanda (changes requested):

CHANGELOG-1.2.md
```diff
@@ -13,6 +13,7 @@ When cutting a new release of the parent `k8ssandra` chart update the `unrelease
 
 ## unreleased
 
+* [ENHANCEMENT] #654 Reduce initial delay of Stargate readiness probe
```
emerkle826, line 16: nit: This should be moved down and grouped with the other `[ENHANCEMENT]` updates in the list.

charts/k8ssandra/templates/stargate/stargate.yaml
adutra: Good idea, implemented in eac44a5.

charts/k8ssandra/templates/stargate/stargate.yaml
```diff
@@ -111,11 +111,13 @@ spec:
               path: /checker/liveness
               port: health
             timeoutSeconds: 10
-            initialDelaySeconds: 240
+            initialDelaySeconds: {{ .Values.stargate.livenessInitialDelaySeconds | default 30 }}
```
jsanda, line 114: Since you also define the default in `values.yaml` the `default` function is unnecessary here. `values.yaml` is part of the chart just like template files. If someone or another charts uses the k8ssandra charts and supplies their own values either by way of command line args with `--set` and/or with a chart properties file, they will be merged with `k8ssandra/values.yaml`. This means that `livenessInitialDelaySeconds` and `readinessInitialDelaySeconds` will not be unset for example by using a new chart properties file.

Does it make sense to make the other probe properties configurable as well? I think it is a question worth asking, but I haven't done enough research to see what is done elsewhere to formulate an opinion. 

I will say this though. It would be nice if we could expose probes in a consistent way across components, although that could be difficult due to differences within components.

jsanda (changes requested):

charts/k8ssandra/templates/stargate/stargate.yaml
adutra: > Since you also define the default in values.yaml the default function is unnecessary here. values.yaml is part of the chart just like template files.

Gotcha, will change that, thanks!

> Does it make sense to make the other probe properties configurable as well?

We'd need to weigh the pros (increased configurability) and cons (increased configuration complexity). My personal position would be to wait until more configurability is actually requested by users.

jsanda: >We'd need to weigh the pros (increased configurability) and cons (increased configuration complexity). My personal position would be to wait until more configurability is actually requested by users.

👍

jsanda (approved):

adutra:
Rebased on main and squashed some commits. Waiting for CI before merging.

PR #818, opened 2021-05-21
K8ssandra Operator design

### LINKED ISSUE #719: Operator Design Discussions
┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-24) by [Unito](https://www.unito.io)
┆Epic: K8ssandra Operator

jdonenine:
Migrating the K8ssandra Operator design document from Google Docs to an in-tree file for review and discussion.

**What this PR does**:
An initial design approach for the build-out of a future K8ssandra Operator to replace a bulk of functionality currently provided by helm.

Fixes #719 
Fixes https://k8ssandra.atlassian.net/browse/K8SSAND-24

**Checklist**

* [ ] Changes manually tested
* [ ] Automated Tests added/updated
* [ ] Documentation added/updated
* [ ] CHANGELOG.md updated (not required for documentation PRs)
* [x] CLA Signed:  [DataStax CLA](https://cla.datastax.com/)

┆Issue is synchronized with this [Jira Bug](https://k8ssandra.atlassian.net/browse/K8SSAND-550) by [Unito](https://www.unito.io)

### DIFF
--- a/technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
+++ b/technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
@@ -1,0 +1,410 @@
+# K8ssandra Operator Design
+
+Author: @jsanda
+
+## Overview
+
+GH issue: https://github.com/k8ssandra/k8ssandra/issues/683
+
+This is a design document for an operator for k8ssandra. For background see 
+https://github.com/k8ssandra/k8ssandra/issues/485.
+
+Much of the behavior that is implemented in the Helm charts will be reimplemented in various controllers. The role of the Helm charts changes to focus primarily on installation and configuration of operators.
+
+While we will continue to use and support Helm, it will not be required for installing k8ssandra from a purely technical perspective. Users can install manifests with kubectl apply for example. We can also provide support for kustomize.
+
+## CustomResourceDefinitions
+
+We will utilize a number of CRDs, some of which are developed and maintained by other projects.
+
+There will be a separate controller for each CRD.
+
+The goal of this document is not to be entirely prescriptive for each CRD. Rather, the intent is to give a general idea of how things will look and fit together. 
+
+### K8ssandraCluster
+
+This is the primary object which users will manage. The controller will be responsible for creating other constituent objects.
+
+It will look something like this:
+
+```
+apiVersion: k8ssandra.io/v1alpha1
+kind: K8ssandraCluster
+metadata:
+  name: k8ssandra
+spec:
+  cassandra:
+    clusterName: k8ssandra
+    auth:
+      enabled: true
+    datacenters:
+    - name: dc1
+      size: 3
+  reaper:
+    enabled: true
+    # optionally set and override defaults for Reaper    
+  stargate: 
+    enabled: true  
+    # optionally set and override defaults for Stargate 
+  monitoring:
+    prometheus:
+      enabled: true
+      # optionally set and override defaults for Prometheus
+    grafana: 
+      enabled: true
+      # optionally set and override defaults for Grafana
+```
+
+Similar to the k8ssandra Helm chart, there is an enabled flag for each component. The Go type for this will be a *bool. In general as it is with the Helm chart, setting enabled: true should be all that is necessary. The controller will take care of configuring the respective objects. 
+
+The user will be able to override defaults as necessary as illustrated in the next example:
+
+```
+apiVersion: k8ssandra.io/v1alpha1
+kind: K8ssandraCluster
+metadata:
+  name: k8ssandra
+spec:
+  cassandra:
+    clusterName: k8ssandra
+    auth:
+      enabled: true
+    datacenters:
+    - name: dc1
+      size: 3
+  reaper:
+    enabled: true
+    serverConfig:
+      jmxUserSecretName: my-custom-secret
+      cassandraBackend:
+        replication:
+          networkTopology:
+            dc1: 5         
+  stargate: 
+    enabled: true  
+    replicas: 3
+```
+
+### CassandraCluster
+
+```
+apiVersion: k8ssandra.io/v1alpha1
+kind: CassandraCluster
+metadata:
+  name: k8ssandra
+spec:
+  clusterName: k8ssandra
+  auth:
+    enabled: true
+  datacenters:
+  - name: dc1
+    size: 3
+```
+
+The controller will create a CassandraDatacenter object for each entry in the datacenters array. Similar to what we have done in the Helm charts, you will be able to specify things at cluster level and override them at the DC level. You will also be able to specify properties at the DC level without specifying them at the cluster level.
+
+Multi-cluster is out of scope, but I want to point out that I think this is where we introduce it. Consider the following example:
+
+```
+apiVersion: k8ssandra.io/v1alpha1
+kind: CassandraCluster
+metadata:
+  name: k8ssandra
+spec:
+  clusterName: k8ssandra
+  auth:
+    enabled: true
+  datacenters:
+  - name: dc1
+    size: 3
+  - name: dc2
+    k8s:
+      kubeconfigSecret: my-other-cluster
+      namespace: dev
+    size: 3
+```
+
+`kubeconfigSecret` specifies the name of a secret that contains the kubeconfig file needed to access a remote cluster. 
+
+`namespace` indicates in which namespace the DC should be created.
+
+### CassandraDatacenter
+
+This is defined in the [cass-operator](https://github.com/k8ssandra/cass-operator) repo. The controller runs in its own Deployment.
+
+### Reaper
+
+```
+apiVersion k8ssandra.io/v1alpha1
+kind: Reaper
+metadata:
+  name: k8ssandra
+spec:
+  serverConfig:
+    storageType: cassandra
+    jmxUserSecretName: k8ssandra
+    cassandraBackend:
+      cassandraDatacenter
+        name: dc1
+      replication:
+        networkTopology:
+          dc1: 3
+```
+
+This is defined in the [reaper-operator](https://github.com/k8ssandra/reaper-operator) repo. It is possible today to use reaper-operator on its own without k8ssandra. In that regard, living in its own repo makes sense. There is a good bit of overhead for each repo that we need to maintain. In practice, I do not think reaper-operator will be used a whole lot outside of k8ssandra. For these reasons I am in favor of consolidating with the k8ssandra-operator repo.
+
+From a technical standpoint, consolidation means that the reaper-operator controllers would run in the k8ssandra-operator binary. It would reduce overhead in that it would be able to use the same cache as other controllers.
+
+### Stargate
+
+This does not yet exist. Similar to reaper-operator, one could certainly make the case that a separate stargate-operator project makes sense. I am more on the fence about where this should live than I am with reaper-operator.
+
+```
+apiVersion: k8ssandra.io/v1alpha1
+kind: Stargate
+metadata:
+  name: k8ssandra
+spec:
+  replicas: 3
+  cassandraDatacenter:
+    name: dc1
+```
+
+### CassandraBackup / CassandraRestore
+
+These are defined in the [medusa-operator](https://github.com/k8ssandra/medusa-operator) repo. I propose consolidating this into the k8ssandra-operator repo.
+
+See this [GitHub discussion](https://github.com/k8ssandra/k8ssandra/discussions/273) more for ideas around backup/restore and CRDs.
+
+### MonitoringConfiguration
+
+```
+apiVersion: k8ssandra.io/v1alpha1
+kind: MonitoringConfiguration
+metadata:
+  name: k8ssandra
+spec:
+  prometheus:
+    # prometheus properties...
+  grafana:
+  # grafana properties...
+```
+
+With the Grafana license changes and discussion around Victoria Metrics, I think it is a good idea to encapsulate Prometheus and Grafana.
+
+### Grafana, GrafanaDashboard, GrafanaDatasource
+
+Defined in [grafana-operator](https://github.com/integr8ly/grafana-operator).
+
+### Prometheus
+
+Defined in [prometheus-operator](https://github.com/integr8ly/grafana-operator).
+
+## Scenarios
+
+Each scenario consists of a series of events. Each event lists the steps performed by controllers. Controllers run concurrently. When you see multiple controllers listed for an event assume that the steps are performed concurrently and potentially in parallel.
+
+### Create K8ssandraCluster
+
+Let's first consider creating a single DC cluster. User creates a K8ssandraCluster object via kubectl apply:
+
+```
+apiVersion: k8ssandra.io/v1alpha1
+kind: K8ssandraCluster
+metadata:
+  name: test
+spec:
+  cassandra:
+    clusterName: "test"
+    version: "4.0.0" 
+    auth:
+      enabled: true
+    datacenters:
+    - name: dc1
+      size: 3
+  reaper:
+    # reaper properties...
+  stargate:
+    enabled: true
+  medusa:
+    enabled: true
+  monitoring:
+    prometheus:
+      enabled: true
+    grafana:
+      enabled: true
+```
+
+#### Event: K8ssandraCluster created
+
+##### k8ssandra controller
+
+* Start reconciliation loop when K8ssandraCluster object is created
+* Creates a CassandraCluster
+  * Update status
+* Creates a Reaper
+  * Update status
+* Creates a Stargate
+  * Update status
+* Creates a MonitoringConfiguration
+  * Update status
+* End reconciliation loop
+
+#### Event: CassandraCluster created
+
+##### cassandra cluster controller
+
+* Start reconciliation loop when CassandraCluster object is created
+* Create CassandraDatacenter
+  * Update status
+* End reconciliation loop
+
+#### Event: Reaper created
+
+##### reaper controller
+
+* Start reconciliation loop when Reaper object is created
+* Creates reaper Service
+* Requeue request until CassandraDatacenter is ready
+
+#### Event: Stargate created
+
+##### stargate controller
+
+* Start reconciliation loop when Stargate object is created
+* Creates stargate Service
+* Creates ServiceMonitor
+* Creates a GrafanaDashboard
+* Requeue request until CassandraDatacenter is ready
+
+#### Event: MonitoringConfiguration created
+
+##### monitoring controller
+
+* Start reconciliation loop when MonitoringConfiguration created
+* Create Grafana object
+* Create GrafanaDataSource object
+* Create Prometheus object
+* End reconciliation loop
+
+#### Event: CassandraDatacenter ready
+
+##### cassandra cluster controller
+
+* Update status to reflect state of CassandraCluster
+* End reconciliation loop
+
+##### reaper controller
+
+* Start reconciliation loop after requeue
+* Create Job to initialize reaper keyspace
+* Requeue request until job finishes
+
+##### stargate controller
+
+* Start reconciliation loop after requeue
+* Create stargate Deployment
+* Requeue request until Deployment ready
+
+#### Event: CassandraCluster status updated
+
+##### k8ssandra controller
+
+* Start reconciliation loop
+  * Update status to reflect state of CassandraCluster
+* End reconciliation loop
+
+#### Event: Reaper schema job completed
+
+##### reaper controller
+
+* Create Deployment
+* Requeue request until Deployment ready
+* End reconciliation loop
+
+#### Event: Stargate ready
+
+##### k8ssandra controller
+
+* Update status to reflect state of Stargate
+
+#### Event: Reaper ready
+
+##### k8ssandra controller
+
+* Update status to reflect state of Reaper
+
+#### Event: Grafana and Prometheus ready
+
+##### monitoring controller
+
+* Update status of MonitoringConfiguration to reflect state of Grafana and Prometheus
+
+#### Event: MonitoringConfiguration status updated
+
+##### k8ssandra controller
+
+* Update status to reflect state of MonitoringConfiguration
+
+#### Notes
+
+The k8ssandra controller creates several objects:
+
+* CassandraCluster
+* Reaper
+* Stargate
+* MonitoringConfiguration
+
+Whenever any of these objects is updated, it should trigger a reconciliation of the k8ssandra controller. In the reconciliation loop, the status of each object will essentially be pushed up into the K8ssandraCluster object.
+
+### Change Cassandra heap settings
+
+The user updates the heap settings in the K8ssandraCluster manifest and applies the changes with kubectl apply.
+
+#### Event: K8ssandraCluster updated
+
+##### k8ssandra controller
+
+* Update CassandraCluster 
+* End reconciliation loop
+
+#### Event: CassandraCluster updated
+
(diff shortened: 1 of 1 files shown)


technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
```diff
@@ -0,0 +1,410 @@
...
+          dc1: 3
+```
+
+This is defined in the [reaper-operator](https://github.com/k8ssandra/reaper-operator) repo. It is possible today to use reaper-operator on its own without k8ssandra. In that regard, living in its own repo makes sense. There is a good bit of overhead for each repo that we need to maintain. In practice, I do not think reaper-operator will be used a whole lot outside of k8ssandra. For these reasons I am in favor of consolidating with the k8ssandra-operator repo.
+
+From a technical standpoint, consolidation means that the reaper-operator controllers would run in the k8ssandra-operator binary. It would reduce overhead in that it would be able to use the same cache as other controllers.
+
+### Stargate
+
+This does not yet exist. Similar to reaper-operator, one could certainly make the case that a separate stargate-operator project makes sense. I am more on the fence about where this should live than I am with reaper-operator.
```
jdonenine, line 160: It's probably a fair question to ask if an operator for Stargate should live in K8ssandra at all, or if there is external value in the Stargate operator residing with Stargate in its org.  Certainly complicates things, and perhaps should start here, but we should definitely have that conversation.

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
```diff
@@ -0,0 +1,410 @@
...
+      cassandraDatacenter
+        name: dc1
+      replication:
+        networkTopology:
+          dc1: 3
+```
+
+This is defined in the [reaper-operator](https://github.com/k8ssandra/reaper-operator) repo. It is possible today to use reaper-operator on its own without k8ssandra. In that regard, living in its own repo makes sense. There is a good bit of overhead for each repo that we need to maintain. In practice, I do not think reaper-operator will be used a whole lot outside of k8ssandra. For these reasons I am in favor of consolidating with the k8ssandra-operator repo.
+
+From a technical standpoint, consolidation means that the reaper-operator controllers would run in the k8ssandra-operator binary. It would reduce overhead in that it would be able to use the same cache as other controllers.
```
jdonenine, line 156: So to be clear, you're proposing more than just a consolidation of the code-base into a single repo, but more-so a consolidation of the handling into a single operator that would handle:  k8ssandra, reaper, and medusa deployment.  Right?

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
```diff
@@ -0,0 +1,410 @@
...
+  auth:
+    enabled: true
+  datacenters:
+  - name: dc1
+    size: 3
+```
+
+The controller will create a CassandraDatacenter object for each entry in the datacenters array. Similar to what we have done in the Helm charts, you will be able to specify things at cluster level and override them at the DC level. You will also be able to specify properties at the DC level without specifying them at the cluster level.
+
+Multi-cluster is out of scope, but I want to point out that I think this is where we introduce it. Consider the following example:
```
jdonenine, line 106: With mulit-cluster support now sort of squarely in our sights for K8ssandra, we should definitely consider this aspect as we start making all of these decisions.

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
```diff
@@ -0,0 +1,410 @@
...
+metadata:
+  name: k8ssandra
+spec:
+  prometheus:
+    # prometheus properties...
+  grafana:
+  # grafana properties...
+```
+
+With the Grafana license changes and discussion around Victoria Metrics, I think it is a good idea to encapsulate Prometheus and Grafana.
```
jdonenine, line 193: +1 here, we'll also need to provide the ability for the control of the version of Grafana being deployed, so that people have the choice of pre-AGPL licensed versions.

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
```diff
@@ -0,0 +1,410 @@
...
+
+### Grafana, GrafanaDashboard, GrafanaDatasource
```
jdonenine, line 195: Here and for Prometheus as well, we would begin deploying directly the grafana-operator and prometheus-operator?

How would that change the configuration capabilities that users have become accustomed to in the k8ssandra chart passing through the kube-prometheus-stack values?

What do we imagine that would now look like at the top-level?

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
```diff
@@ -0,0 +1,410 @@
...
+#### Notes
+
+The k8ssandra controller creates several objects:
+
+* CassandraCluster
+* Reaper
+* Stargate
+* MonitoringConfiguration
+
+Whenever any of these objects is updated, it should trigger a reconciliation of the k8ssandra controller. In the reconciliation loop, the status of each object will essentially be pushed up into the K8ssandraCluster object.
```
jdonenine, line 358: With the push up of the status you mention, we would then be able to inspect just the top-level object to get status information for the whole deployment from the top-down?

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
```diff
@@ -0,0 +1,410 @@
...
+
+With the Grafana license changes and discussion around Victoria Metrics, I think it is a good idea to encapsulate Prometheus and Grafana.
+
+### Grafana, GrafanaDashboard, GrafanaDatasource
+
+Defined in [grafana-operator](https://github.com/integr8ly/grafana-operator).
+
+### Prometheus
+
+Defined in [prometheus-operator](https://github.com/integr8ly/grafana-operator).
```
jdonenine, line 201: Just a minor note: You've got the wrong link here

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
```diff
@@ -0,0 +1,410 @@
...
+
+## Being Prescriptive
+
+With the Helm charts we are prescriptive with what properties and settings of the CassandraDatacenter we expose. I have been thinking that each parent type would have a TemplateSpec property for each of its child types. For example, K8ssandraCluster would have a CassandraClusterTemplate property that is of type CassandraClusterSpec. CassandraCluster would have something like a []CassandraDatacenterTemplate where CassandraDatacenterTemplate is of type CassandraDatacenterSpec. This provides flexibility and is a pretty common design pattern. 
+
+Aside from deviating away from the current, prescriptive approach there is something else that we have to consider. Suppose the user decides to add a volume or container that clases with one that K8ssandra adds. cass-operator already deals with this with the PodTemplateSpec property. I am not suggesting that we do one thing or another at this time. I just want to raise awareness.
```
jdonenine, line 406: Can you give an example of what you're talking about here?

jdonenine (commented):
Thanks for sharing this @jsanda there's a lot of good content here.  The description of the event handling gives a good idea of what the operator is responsible for doing for someone who hasn't built an operator before.

There's a few areas I think we'll need to explore in some more depth, probably separately:

* Multi-cluster
* Upgrade Scenarios

You touched on them here, but I think they probably are worthy of their own independent design and discussion.

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
jsanda: Short answer is yes.

I struggle to see the value of maintaining separate repos for these various operators/controllers. If we package them in the same binary we still maintain loose coupling and separation of concerns. We also have better, easier code reuse.

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
jsanda: I have been thinking a lot about multi-cluster over the past few days, particularly around how we do orchestration. I will try to get something written up within the next couple days.

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
jsanda: >we would begin deploying directly the grafana-operator and prometheus-operator?

yes

>How would that change the configuration capabilities that users have become accustomed to in the k8ssandra chart passing through the kube-prometheus-stack values?

We can still provide the ability to declare monitoring configuration through k8ssandra.

>What do we imagine that would now look like at the top-level?

It helps to look at this in terms of the underlying Go types. For prometheus-operator we have [PrometheusSpec](https://github.com/prometheus-operator/prometheus-operator/blob/master/pkg/apis/monitoring/v1/types.go#L86) that defines all of the properties you can declare in a Prometheus object. We could have something like this:

```go
type K8ssandraClusterSpec struct {
    CassandraCluster       CassandraCluster `json:"cassandra,omitempty"`

    PrometheusTemplate PrometheusTemplateSpec `json:"prometheus,omitempty"`

    // Other properties....
}

type PrometheusTemplateSpec struct {
    metav1.ObjectMeta `json:"metadata,omitempty"`

    // This is the same type defined in prometheus-operatoroperator/blob/master/pkg/apis/monitoring/v1/types.go#L86)
    Spec PrometheusSpec `json:"spec,omitempty"`
}
```
This is a familiar pattern that we see with other parent/child types like StatefulSet and Deployment. Both the StatefulSetSpec and DeploymentSpec types have a PodTemplateSpec property.

On the one hand this provides the flexibility to declare and configure things as you see fit. On the other hand, it potentially means are less prescriptive as we are today. @jdonenine this is an area that I think merits some additional discussion.

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
jsanda: Yup, that's the idea. Then you can inspect child objects for additional details. I know that I said multi-cluster will be covered in a separate, follow-up doc, but I want to point out that this would apply in a multi-cluster deployment as well. That top-level status would provide information across all k8s clusters.

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
jsanda: I wound up swerving into an example in my [previous comment](https://github.com/k8ssandra/k8ssandra/pull/818#discussion_r643960190) but it is worth going over again. Take a look at a StatefulSet or Deployment manifest. Consider this example from the kubernetes [docs](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/):

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  labels:
    app: nginx
spec:
  replicas: 3
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80
```
The `template` property declare the pod's metadata and spec. Anything that can be declared in a pod manifest (at least the spec), can be declared here. Let's take a look at the underlying Go types to see how this happens.

```go
type DeploymentSpec struct {
     ...

    // Template describes the pods that will be created.
    Template v1.PodTemplateSpec `json:"template" protobuf:"bytes,3,opt,name=template"`

    ...
}

// PodTemplateSpec describes the data a pod should have when created from a template
type PodTemplateSpec struct {
	// Standard object's metadata.
	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
	// +optional
	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`

	// Specification of the desired behavior of the pod.
	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
	// +optional
	Spec PodSpec `json:"spec,omitempty" protobuf:"bytes,2,opt,name=spec"`
}
```

Here is the Go definition for `CassandraDatacenter`:

```go
type CassandraDatacenter struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	Spec   CassandraDatacenterSpec   `json:"spec,omitempty"`
	Status CassandraDatacenterStatus `json:"status,omitempty"`
}
```

The `CassandraCluster`  type definition might look like this:

```go
type CassandraCluster struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	Spec   CassandraClusterSpec   `json:"spec,omitempty"`
	Status CassandraClusterStatus `json:"status,omitempty"`

}

type CassandraClusterSpec struct {
        DatacenterTemplates []CassandraDatacenterTemplateSpec `json:"datacenters"`       
}

type CassandraDatacenterTemplateSpec struct {
        metav1.ObjectMeta `json:"metadata,omitempty"`

        Spec CassandraDatacenterSpec `json:"spec,omitempty"`
}
```

This is not prescriptive like we are today. It provides the flexibility to declare and configure underlying objects as you see fit.

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
```diff
@@ -0,0 +1,410 @@
...
+
+#### Event: CassandraCluster status updated
+
+##### k8ssandra controller
+
+* Update status to reflect state of CassandraCluster
+
+#### Notes
+
+Notice that the user does not directly update the CassandraDatacenter. The CassandraDatacenter is owned and managed by the CassandraCluster controller. Updates made directly to the CassandraDatacenter would potentially be lost. Changes need to be propagated through the K8ssandraCluster object.
```
jeffbanks, line 400: Does this still allow for template change detection for the deployment for the new updates?  Or, does that end up impacting ALL of the K8ssandraCluster "children" objects?

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
jeffbanks: I like that idea, a rolled-up status that includes all k8s clusters would be awesome.

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
jeffbanks: Maybe this is where some upfront validations will assist with that flexibility.

jsanda (approved):

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
jsanda: Not sure I understand the question but maybe some background will help. Each controller watches for events involving a particular custom resource. Those events include create, update, delete. The controller's job is to make the desired state of the object (as declare in its spec) match the actual state of the object.

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
```diff
@@ -0,0 +1,410 @@
...
+## Overview
+
+GH issue: https://github.com/k8ssandra/k8ssandra/issues/683
+
+This is a design document for an operator for k8ssandra. For background see 
+https://github.com/k8ssandra/k8ssandra/issues/485.
+
+Much of the behavior that is implemented in the Helm charts will be reimplemented in various controllers. The role of the Helm charts changes to focus primarily on installation and configuration of operators.
+
+While we will continue to use and support Helm, it will not be required for installing k8ssandra from a purely technical perspective. Users can install manifests with kubectl apply for example. We can also provide support for kustomize.
```
burmanm, line 14: Given that kubectl ships with kustomize these days (kubectl apply -k), it's kinda given we should design our CRDs in a way that allows to use kustomize generators / transformers.

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
```diff
@@ -0,0 +1,410 @@
...
+
+## CustomResourceDefinitions
+
+We will utilize a number of CRDs, some of which are developed and maintained by other projects.
```
burmanm, line 18: I guess instead of shipping with pre-generated yamls from other projects, we should look at including only the api modules through go dependencies and do go generate to get the yaml files. That means we should probably have proper annotations in all those apis (such as //+kubebuilder:rbac styles)

burmanm: Of course, if we include reaper & medusa operator to this single operator, then that's automated process - but for cass-operator resources we can do the same. After that, only thing required would be to update the go.mod

burmanm: If - and only if - we don't do kustomize templates in cass-operator that modify the generated CRD. Which we actually might want to do for some kube-x- properties, so keep the generated yaml for cass-operator.

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
```diff
@@ -0,0 +1,410 @@
...
+    datacenters:
+    - name: dc1
+      size: 3
+  reaper:
+    enabled: true
+    # optionally set and override defaults for Reaper    
+  stargate: 
+    enabled: true  
+    # optionally set and override defaults for Stargate 
+  monitoring:
```
burmanm, line 49: Newest operator ships with kube-auth-proxy also (as optional dependency) that allows auth to restrict metrics endpoints. Do we wish to support those features also? Also, operator-sdk generates the ServiceMonitors and metrics configuration automatically these days. So we might want to relook at using kube-prometheus-stack (including the fact that Grafana is problematic)

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
```diff
@@ -0,0 +1,410 @@
...
+spec:
+  clusterName: k8ssandra
+  auth:
+    enabled: true
+  datacenters:
+  - name: dc1
+    size: 3
+```
+
+The controller will create a CassandraDatacenter object for each entry in the datacenters array. Similar to what we have done in the Helm charts, you will be able to specify things at cluster level and override them at the DC level. You will also be able to specify properties at the DC level without specifying them at the cluster level.
```
burmanm, line 104: Which one is user intended to update? If I update CassandraCluster for dc1, will the controller update K8ssandraCluster's dc1 entry? Or will I lose the updates? 

I'm slightly worried too many CRDs that are just "nested" might confuse users. This already creates a path: K8ssandraCluster -> CassandraCluster -> CassandraDatacenter to deploy a single Cassandra. Will we restrict through webhooks the modification of certain ones (if installed with K8ssandraManager/Operator) ? I don't think just giving users the "don't do that" doc is going to be user-friendly.

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
burmanm: If Stargate is separate then it makes medusa / reaper ingestion quite odd. I'd say either all of them, or none of them.

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
```diff
@@ -0,0 +1,410 @@
...
+#### Event: K8ssandraCluster created
+
+##### k8ssandra controller
+
+* Start reconciliation loop when K8ssandraCluster object is created
+* Creates a CassandraCluster
+  * Update status
+* Creates a Reaper
+  * Update status
+* Creates a Stargate
```
burmanm, line 247: If both Reaper & Stargate depend on the created CassandraCluster, wouldn't it make more sense to create them on the next reconcile loop once the CassandraCluster has been created? So that we don't create them to poll some non-existant resource.

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
```diff
@@ -0,0 +1,410 @@
...
+    name: dc1
+```
+
+### CassandraBackup / CassandraRestore
+
+These are defined in the [medusa-operator](https://github.com/k8ssandra/medusa-operator) repo. I propose consolidating this into the k8ssandra-operator repo.
+
+See this [GitHub discussion](https://github.com/k8ssandra/k8ssandra/discussions/273) more for ideas around backup/restore and CRDs.
+
+### MonitoringConfiguration
```
burmanm, line 179: I don't know what the product target is, but perhaps we shouldn't at all have this? What if we just provided ways of getting metrics? And not because someone might want Victoria Metrics. What if we want to create a single pane of glass, an UI where we could use our own graphs? It feels like less hassle soon than to maintain Grafana. At the same time, some people might want to use their global bought monitoring system.

And in this case, I don't mean necessarily that metrics couldn't be configured in a single place, but that we wouldn't think about external products to deploy here. Just basically rather:

* serviceMonitors: true
* auth-metrics-endpoint: true

And then give instructions perhaps in a separate build how to deploy our current grafana boards etc if they wish to use them (with caveats). We might need to include other parts as well, like "cert-manager" templates etc

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
```diff
@@ -0,0 +1,410 @@
...
+
+#### Event: Stargate created
+
+##### stargate controller
+
+* Start reconciliation loop when Stargate object is created
+* Creates stargate Service
+* Creates ServiceMonitor
+* Creates a GrafanaDashboard
+* Requeue request until CassandraDatacenter is ready
```
burmanm, line 278: I feel like these "requeue" until something is ready are unnecessary and cause "ignore-this errors", when we could just create them after we know the resource is ready. It reduces noise in case of deployment error.

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
burmanm: Is this push two-way? So when K8ssandraCluster object is modified, those changes too are pushed down?

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
```diff
@@ -0,0 +1,410 @@
...
+  name: k8ssandra
+spec:
+  cassandra:
+    clusterName: k8ssandra
+    auth:
+      enabled: true
+    datacenters:
+    - name: dc1
+      size: 3
+  reaper:
```
burmanm, line 43: One more discussion point.. do we still want to maintain that all these are separate products? Or just separate features? Could I just configure my repairs here? My backups here? I know we've gone back and forth with these, but does the word "reaper" give anything to users? Or medusa? Or stargate?

burmanm: Or just maintenance-tools ?

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
jsanda: To start everything with the exception of cass-operator should live in-tree.

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
jsanda: I think we should provide kustomize support for cass-operator as well. As for some of the kube-x properties, we can apply patches with kustomize to handle them. We already do this with medusa-operator.

technical-docs/K8SSANDRA-OPERATOR-DESIGN.md
jsanda: Am I correct in thinking these ServiceMonitors are for the operators themselves?

PR #826, opened 2021-05-25
Add design doc for remote restores using Medusa

### LINKED ISSUE #720: Backup/Restore Design Discussions
┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-23) by [Unito](https://www.unito.io)

adejanovski:
Fixes #720

┆Issue is synchronized with this [Jira Bug](https://k8ssandra.atlassian.net/browse/K8SSAND-551) by [Unito](https://www.unito.io)


technical-docs/2021-05-remote-restore-redesign.md
```diff
@@ -0,0 +1,110 @@
...
+The sync should allow:
+
+- Creating local missing resources from existing backups
+- Deleting local resources for missing backups
+- Deleting remote backups if a local resource was marked as deleted
+
+A single control loop should be able to handle all 3 operations.
+The sync should be required explicitly, meaning that a new CRD should be created to request synchronization to the medusa-operator, and that CRD should contain a prefix so that a cluster can list backups stored for another cluster in a multi-tenant storage bucket.
+
+The CassandraBackup resource should be updated, adding a delete field which would allow marking backups for deletion.
```
jsanda, line 31: The way to mark an object for deletion is with finalizers. See this [post](https://kubernetes.io/blog/2021/05/14/using-finalizers-to-control-deletion/) for background.

Here is brief summary of how it works. When a CassandraBackup is created the backup controller adds a finalizer to the object. Kubernetes will not delete an object that has finalizers. When the CassandraBackup is deleted, Kubernetes sets a deletion timestamp on the object and reconciliation is triggered. The backup controller sees that the deletion timestamp is set and then proceeds with whatever clean up after is needed. When all of the clean up is done, the finalizer is removed. At this point Kubernetes can continue with the actual deletion process.

technical-docs/2021-05-remote-restore-redesign.md
```diff
@@ -0,0 +1,110 @@
...
+The sync operation would:
+
+- List the backups in the storage bucket, through a gRPC call
+- Compare that list to the existing CassandraBackup resources
+- Create a CassandraBackup object for each missing backup
+- Delete local resources that don’t exist in the list of backups
+- Delete remote backups through gRPC for CassandraBackup objects marked for deletion, and upon success delete the resource.
+
+## CassandraBackup CRD changes
+CassandraBackup resources shouldn’t store the CassandraDatacenter spec anymore but rather the topology object generated by Medusa.
```
jsanda, line 42: Can you provide an example of what the topology object looks like?

technical-docs/2021-05-remote-restore-redesign.md
```diff
@@ -0,0 +1,110 @@
...
+
+## CassandraBackup CRD changes
+CassandraBackup resources shouldn’t store the CassandraDatacenter spec anymore but rather the topology object generated by Medusa.  
+This object will contain the list of live nodes at the time of the backup along with their token assignments.  
+The CRD should also contain information that should be overridable such as the prefix, which is controlled by the namespace name. 
+
+## Medusa changes
+The tokenmap generated by Medusa will need to be augmented to provide the placement of nodes in the cluster: datacenter and rack.
+Currently Medusa is not dc/rack aware and it can probably be fairly surprising to users. It was built to backup a single DC per cluster and didn’t account for racks either.  
+Restoring from a DC/rack aware cluster to another one imposes some constraints on the source/target nodes mapping, which are currently impossible to deal with safely in a remote restore.
```
jsanda, line 49: What are those constraints?

technical-docs/2021-05-remote-restore-redesign.md
```diff
@@ -0,0 +1,110 @@
...
+
+![](img/K8ssandra-new-restore-design-5.jpg)  
+
+  Once the cassandradatacenter is in Ready state, the restore will be marked as finished and the cluster will be fully restored and available.
+
+## Other required changes
+
+**The cassandradatacenter object will not be recreated during restore.**
+Currently the remote restore procedure assumes that the Cassandra cluster is not running and that the cassandradatacenter object hasn’t been created.
+This is a problem because creating the cassandradatacenter shouldn’t predate the helm install which includes other components (Reaper, monitoring, …) that will not be restored by Medusa.
```
jsanda, line 103: We need to think about this with an operator-based implementation as opposed to the current Helm based solution.

technical-docs/2021-05-remote-restore-redesign.md
```diff
@@ -0,0 +1,110 @@
...
+The new procedure will assume that a cassandradatacenter has been created already using the same number of nodes (token assignments can be different and will be enforced, datacenter name can differ).
```
jsanda, line 104: If the new procedure assumes that the CassandraDatacenter already exists then would the restore procedure be the same or similar to what is done for in-place restores with respect to the restore controller?

technical-docs/2021-05-remote-restore-redesign.md
```diff
@@ -0,0 +1,110 @@
...
+- Restoring a medusa-operator based backup to another Cassandra install in a different k8s cluster
+![](img/medusa_k8s-new-restore-design-2.jpg)
+
+- Restoring a Medusa backup (non k8s) to a k8s Cassandra install
+![](img/medusa_k8s-new-restore-design-3.jpg)
+
+
+
+
+## Sync CassandraBackup resources with storage
```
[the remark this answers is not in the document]
adejanovski: Yes, but they would use different prefixes assuming multitenancy is turned on. That's why I'd want to be able to override the prefix at sync/restore time.
It would be up to the ops to change the storage bucket settings after a restore if needed.

technical-docs/2021-05-remote-restore-redesign.md
adejanovski: Sounds good, I'll update the doc appropriately to use finalizers instead.

technical-docs/2021-05-remote-restore-redesign.md
adejanovski: Sure thing, it currently looks like this:
```
{"ip-179-31-16-38.us-west-2.compute.internal": {"tokens": [-1051417974279770378, -112095142485141927, ..., 9217379683620055011, 939597040143840721], "is_up": true}, 
 "ip-179-31-23-54.us-west-2.compute.internal": {"tokens": [-1080182033668782299, -1156652990105327691, ..., 9134535641516355167, 9216775765860996507], "is_up": true}, 
 "ip-179-31-30-139.us-west-2.compute.internal": {"tokens": [-1018393144648419592, -1342532530495729638, ..., 8846603596658721956, 8849342512812637165], "is_up": true}}
```

It should be augmented with the dc and rack information.

technical-docs/2021-05-remote-restore-redesign.md
adejanovski: When using NetworkTopologyStrategy, the placement of replicas is dependent on the DC and rack layout.
In this context, the replica for a token range will be placed on the node that owns the next token range in the next rack.
We can't restore any node to any other node as we may end up putting several replicas in the same rack, and the nodes we'd restore to would possibly end up not containing data they're supposed be replica for.
We can be rigid about the mapping and expect the same dc and rack names exclusively, or do a 1:1 mapping between DCs/Racks despite having different names (but expect the same number of both).

technical-docs/2021-05-remote-restore-redesign.md
adejanovski: Does it change the fact that we should expect the rest of the components to be installed as they wouldn't be recreated during the restore operation?

technical-docs/2021-05-remote-restore-redesign.md
adejanovski: Not really, because remote restores do two things differently compared to in place restores:

- `system.local` and `system.peers` are not restored
- tokens are enforced and the nodes need to bootstrap

technical-docs/2021-05-remote-restore-redesign.md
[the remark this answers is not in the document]
jsanda: Ok cool, so they wouldn't be storing backups in the same location. 👍

technical-docs/2021-05-remote-restore-redesign.md
jsanda: Thanks for the detailed explanation!

technical-docs/2021-05-remote-restore-redesign.md
adejanovski: I've updated the doc to use finalizers. Let me know how it looks.

technical-docs/2021-05-remote-restore-redesign.md
adejanovski: Added the sample to the design doc.

technical-docs/2021-05-remote-restore-redesign.md
adejanovski: Thanks, I've added the explanation to the doc itself.

technical-docs/2021-05-remote-restore-redesign.md
jsanda: Looks good!

technical-docs/2021-05-remote-restore-redesign.md
```diff
@@ -0,0 +1,122 @@
+# Medusa Operator - Restore redesign document
+
+## Background
+
+Medusa operator doesn’t yet support remote restores: restoring a backup to another cluster.
+As we implement this feature, we would like to have it support multiple use cases that operators could expect to have working:
+
+- Restoring a medusa-operator based backup to another Cassandra install in the same k8s cluster
+![](img/medusa_k8s-new-restore-design-1.jpg)
```
adutra, line 9: It seems that the images do not correspond to the descriptions.

adutra: Also the images read "K8ssandra" where I think we mean "Kubernetes cluster".

technical-docs/2021-05-remote-restore-redesign.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Medusa changes
+The tokenmap generated by Medusa will need to be augmented to provide the placement of nodes in the cluster: datacenter and rack.
+Currently Medusa is not dc/rack aware and it can probably be fairly surprising to users. It was built to backup a single DC per cluster and didn’t account for racks either.  
+Restoring from a DC/rack aware cluster to another one imposes some constraints on the source/target nodes mapping, which are currently impossible to deal with safely in a remote restore: 
+
+When using NetworkTopologyStrategy, the placement of replicas is dependent on the DC and rack layout.
+In this context, the replica for a token range will be placed on the node that owns the next token range in the next rack.
+We can't restore any node to any other node as we may end up putting several replicas in the same rack, and the nodes we'd restore to would possibly end up not containing data they're supposed be replica for.
+We can be rigid about the mapping and expect the same dc and rack names exclusively, or do a 1:1 mapping between DCs/Racks despite having different names (but expect the same number of both).
```
adutra, line 61: I'm clearly a newbie in Medusa but dealing with heterogenous topologies sounds to me as a recipe for trouble. Shouldn't we put some constraints on what the origin and target clusters should look like? E.g. same number of nodes, 1:1 dc-rack pairing, same topology strategy... Another idea: require the user to provide the full origin-to-target mapping somehow?

technical-docs/2021-05-remote-restore-redesign.md
```diff
@@ -0,0 +1,122 @@
...
+
+2. Set the env variables on the restore containers
+![](img/K8ssandra-new-restore-design-2.jpg)  
+The backup definition will be used to map the backup hosts with the restore hosts and perform checks on whether or not the restore is possible:
+
+  - Same number of DCs
+  - Same number of nodes per DC
+  - Same rack distribution per DC
+  
+  Nodes will be mapped 1:1 between the source and destination topologies (sorted alphabetically). If the mapping fails to associate nodes 1:1, the restore operation must fail.
```
adutra, line 79: Ok that answers my question above :-)

adutra: Also, do we need to check for same token ranges? Or this doesn't matter? If origin and target nodes have different ranges, I wonder if target could end up containing data that don't belong to it.

technical-docs/2021-05-remote-restore-redesign.md
```diff
@@ -0,0 +1,122 @@
...
+  - Bucket prefix (new)
+  - In Place/Remote (new)
+
+  These values will then be used to invoke medusa.restore_node in the medusa-restore init container.
+The init-container needs to deal with mapping between source and dest topologies to self determine which backup it needs to restore. Pre-checks in the operator will guarantee that a 1:1 mapping is possible and topologies are compatible.
+
+3. Shutdown Cassandra to trigger the restore
+![](img/K8ssandra-new-restore-design-3.jpg)  
+
+  medusa-operator will patch the cassandradatacenter spec to provoke a shutdown of the Cassandra nodes through cass-operator. The pods will be restarted and the medusa-restore init container will trigger the restore.
```
adutra, line 94: Is this going to happen all nodes at once, or this is going to be rolling restarts?

technical-docs/2021-05-remote-restore-redesign.md
```diff
@@ -0,0 +1,122 @@
...
+
+4. Restore the backups
+
+![](img/K8ssandra-new-restore-design-4.jpg)  
+
+  The medusa-restore container will perform a remote restore which involves the following:
+system.local and system.peers will not be restored
+auto_bootstrap will be set to false and seed nodes will start first
+tokens will be enforced by modifying cassandra.yaml
```
adutra, line 103: Interesting, so we are going to override the current token distribution and apply the distribution from the origin cluster. Indeed, this would require all nodes to be stopped at once. Couldn't we image a softer approach: same token distribution in origin and target => no need to do a full shutdown. For people with strong SLAs, avoiding a downtime during restore could be a good thing, unless this procedure is being designed primarily for setting up new clusters from scratch (as opposed to restoring a backup to an existing, running cluster).

technical-docs/2021-05-remote-restore-redesign.md
adejanovski: Good catch 😅 
I'll fix this.

technical-docs/2021-05-remote-restore-redesign.md
adejanovski: Supporting restore in cluster that have a different topology is non trivial indeed but that's really what you expect from a backup/restore solution.

As you found out in your next comment, we'll limit the supported cases here as I'm not crazy about using the sstableloader...

technical-docs/2021-05-remote-restore-redesign.md
adejanovski: Medusa will enforce the tokens to match the original topology. We need to make sure the backup node and restore node are placed the same in the NetworkTopologyStrategy context, otherwise we'll have some replicated tokens that won't be accessible indeed.
So given we have the same number of DCs and racks (which is a prereq), a backup from a node in the first DC and the second rack must be restored on a node in the first DC and the second rack in the target cluster.

technical-docs/2021-05-remote-restore-redesign.md
adejanovski: All at once, so that we can guarantee a consistent restore from the apps perspective. In the case of in place restores, we're studying how we can perform live restores while staying consistent, but it'll be part of another task.

technical-docs/2021-05-remote-restore-redesign.md
adejanovski: We're working on live restores for in place restores only (although we'd still prevent apps from accessing the cluster while the restore is running). Here we're talking about remote restores which will most likely require to enforce the tokens.
Also Medusa has an "all or nothing" philosophy. While people could expect to be able to restore a single table from a cluster and a single keyspace from another one to the same target cluster, that comes with too many challenges to be reliable in the Cassandra world. Stop, wipe and restore is how we do stuff in Medusa, for better or worse :)

jsanda (approved):

PR #834, opened 2021-05-27
Adds upgrade integration tests and upgrades cass-operator to v1.7.1

adejanovski:
**What this PR does**:
Adds upgrade integration tests and upgrades cass-operator to v1.7.1

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/tests/integration/charts/cluster_with_medusa_local.yaml
+++ b/tests/integration/charts/cluster_with_medusa_local.yaml
@@ -7,8 +7,6 @@ cassandra:
     size: 1
   ingress:
     enabled: false
-  loggingSidecar:
-    enabled: true
 stargate:
   enabled: false
 reaper:
--- a/tests/integration/charts/cluster_with_medusa_local.yaml
+++ b/tests/integration/charts/cluster_with_medusa_local.yaml
@@ -7,6 +7,8 @@ cassandra:
     size: 1
   ingress:
     enabled: false
+  loggingSidecar:
+    enabled: true
 stargate:
   enabled: false
 reaper:
--- a/tests/integration/integration_suite_test.go
+++ b/tests/integration/integration_suite_test.go
@@ -19,7 +19,7 @@ const (
 )
 
 // "" means latest stable version in the Helm repo
-var upgradeStartVersions = []string{"v1.0.0", ""}
+var upgradeStartVersions = []string{"v1.0.0", "latest"}
 
 func TestMain(m *testing.M) {
 	err := InitTestClient()
@@ -67,6 +67,8 @@ func cleanupCluster(t *testing.T, namespace string, success bool) {
 		DeleteNamespace(t, traefikNamespace)
 		DeleteNamespace(t, minioNamespace)
 		CheckNamespaceIsAbsent(t, namespace)
+		CheckNamespaceIsAbsent(t, traefikNamespace)
+		CheckNamespaceIsAbsent(t, minioNamespace)
 	} else {
 		log.Println(Info("Not cleaning up cluster as requested"))
 	}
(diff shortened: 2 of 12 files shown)


burmanm:
Could you rebase and we could get those modifications merged?

tests/integration/charts/cluster_with_medusa_local.yaml
```diff
@@ -7,6 +7,8 @@ cassandra:
     size: 1
   ingress:
     enabled: false
+  loggingSidecar:
```
burmanm, line 10: enabled is the default, so why bother readding them?

tests/integration/integration_suite_test.go
```diff
@@ -347,15 +352,40 @@ func testStargate(t *testing.T, namespace string) {
 func TestUpgradeScenario(t *testing.T) {
 	namespace := initializeCluster(t)
 	// Install first production version
-	InstallK8ssandraFromRepo(t, namespace, "1.0.0")
+	DeployClusterWithValues(t, namespace, "default", "cluster_with_reaper.yaml", 1, false, false)
```
burmanm, line 355: But this does not actually deploy an old version? Or where is the version defined?

burmanm: I think we might want two tests that is, one from older version, one from the previous stable. Or even one -> stable -> current.

tests/integration/charts/cluster_with_medusa_local.yaml
adejanovski: You're right, I'll remove these.

tests/integration/integration_suite_test.go
adejanovski: Agreed, let me make it so that we can specify a version.

tests/integration/integration_suite_test.go
```diff
@@ -345,17 +353,46 @@ func testStargate(t *testing.T, namespace string) {
...
+		checkResourcePresenceForReaper(t, namespace)
+		waitForReaperPod(t, namespace)
+
+		// Upgrade to current version
+		DeployClusterWithValues(t, namespace, "default", "cluster_with_reaper.yaml", 1, true, true, "")
+		checkResourcePresenceForReaper(t, namespace)
+		waitForReaperPod(t, namespace)
+		checkReaperRegistered(t, namespace)
+
+		cleanupCluster(t, namespace, t.Failed())
```
burmanm, line 369: If this fails, we shouldn't ever continue (as it will fail to same name etc issue). Should probably just do if t.Failed() .. { t.FailNow() } or equal.

burmanm:
Upgrade tests fail to "failed to install the minio helm chart: error while running command: exit status 1; Error: cannot re-use a name that is still in use" now. Probably those DeployClusterWithValues ones should recognize that they're running upgrade instead of install.

tests/integration/steps/integration_steps.go
```diff
@@ -86,16 +86,31 @@ func getKubectlOptions(namespace string) *k8s.KubectlOptions {
...
+		clusterChartPath = "k8ssandra/k8ssandra"
+	}
+
 	customChartPath, err := filepath.Abs("charts/" + customValues)
 	g(t).Expect(err).To(BeNil())
 
 	if os.Getenv("K8SSANDRA_CASSANDRA_VERSION") != "" {
-		log.Println(Info(fmt.Sprintf("Using Cassandra version %s", os.Getenv("K8SSANDRA_CASSANDRA_VERSION"))))
-		helmValues["cassandra.version"] = os.Getenv("K8SSANDRA_CASSANDRA_VERSION")
+		if !useLocalCharts && strings.HasPrefix(os.Getenv("K8SSANDRA_CASSANDRA_VERSION"), "3.11") {
```
burmanm, line 107: We can fix this later, but probably we shouldn't do it for 4.0 either. Although, also - this is a scenario which can fail in the real update.

burmanm (approved):

PR #845, opened 2021-06-01
Remove k8ssandra from the path of CRDs and instead check all subpaths

burmanm:
**What this PR does**:
Removes k8ssandra from the CRD path

Fixes #<issue number>

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/cmd/k8ssandra-client/Dockerfile.it
+++ b/cmd/k8ssandra-client/Dockerfile.it
@@ -22,6 +22,6 @@ FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY --from=builder /workspace/k8ssandra-client .
 USER nonroot:nonroot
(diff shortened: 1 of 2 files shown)


jsanda (approved):

jsanda:
I confirmed that these changes along with the changes in main do not work, particularly the Makefile target for building the tools image, do not work for downloading an image. 

I also confirmed that the changes do work if you upgrade against the chart in the local git repo. 

I can upgrade against a snapshot version in the Helm repo as follows:

```
helm upgrade k8ssandra  k8ssandra-stable/k8ssandra --devel --version 1.2.0-20210531063246-34e3c4b2
```
A simple solution is to make using the local charts for the tools image configurable as I did here https://github.com/jsanda/k8ssandra/commit/20089a115a25040e64956756c5c0ebed56815b46. This way, if we intend to upgrade to a version in the remote Helm repo we do not copy any directories and avoid the file system permission issues.

PR #883, opened 2021-06-14
Detect IEC formatted c* heap.size and heap.newGenSize

### LINKED ISSUE #701: Add validation check for Cassandra heap size properties
Users (including myself) often run into the problem of specifying invalid values for the JVM heap size. This prevents Cassandra from starting and can be difficult to debug and track down. This happens because k8s resources limits expect `1Gi` for example, while the JVM expects `1G`.

Here is an example of invalid values that will prevent Cassandra from starting:

```yaml
#invalid-values.yaml

cassandra:
  heap:
    size: 1Gi
    newGenSize: 512Mi
```

And here is an example of valid values:

```yaml
#valid-values.yaml

cassandra:
  heap:
    size: 1G
    newGenSize: 512M

```

I would like validation checks for the `heap.size` and `heap.newGenSize` properties at both the cluster and DC levels. The error message should be descriptive and list the format of acceptable values.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-29) by [Unito](https://www.unito.io)
┆Fix Versions: k8ssandra-1.3.0

jeffbanks:
**What this PR does**:

Detection of IEC formatted (base 2) values used instead of SI (base 10) for C* `heap.size` and `heap.newGenSize` at both the c* cluster and dc levels.  E.g. no longer support 1MiB, but would need to be 1M.

Fixes #701 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/charts/k8ssandra/templates/_helpers.tpl
+++ b/charts/k8ssandra/templates/_helpers.tpl
@@ -117,24 +117,41 @@ Return the ingress host that should be used for Stargate's Cassandra/CQL interfa
 
 {{/*
 Create the jvm options based on heap properties specified.
+Expecting that c* heap.size and heap.newGenSize are NOT in IEC format.
 */}}
 {{- define "k8ssandra.configureJvmHeap" -}}
 {{- $datacenter := (index .Values.cassandra.datacenters 0) -}}
 {{- if $datacenter.heap }}
   {{- if $datacenter.heap.size }}
-      initial_heap_size: {{ $datacenter.heap.size }}
-      max_heap_size: {{ $datacenter.heap.size }}
+      {{- if (regexMatch "^(([0]\\.\\d*)+|(^[1-9]\\d*)+(\\.\\d+)?)(?i)(k|m|g|e|p|t){1}$" (print $datacenter.heap.size) ) }}
+        {{- nindent 6 (print "initial_heap_size: " $datacenter.heap.size) }}
+        {{- nindent 6 (print "max_heap_size: " $datacenter.heap.size) }}
+      {{- else }}
+        {{- fail "Specify datacenter.heap.size using one of these suffixes: E, P, T, G, M, K. Format: <NUMBER>[.<NUMBER>]<SUFFIX>" }}
+      {{- end }}
   {{- end }}
   {{- if $datacenter.heap.newGenSize }}
-      heap_size_young_generation: {{ $datacenter.heap.newGenSize }}
+      {{- if (regexMatch "^(([0]\\.\\d*)+|(^[1-9]\\d*)+(\\.\\d+)?)(?i)(k|m|g|e|p|t){1}$" (print $datacenter.heap.newGenSize) ) }}
+        {{- nindent 6 (print "heap_size_young_generation: " $datacenter.heap.newGenSize) }}
+      {{- else }}
+        {{- fail "Specify datacenter.heap.newGenSize using one of these suffixes: E, P, T, G, M, K. Format: <NUMBER>[.<NUMBER>]<SUFFIX>" }}
+      {{- end }}
   {{- end }}
 {{- else if .Values.cassandra.heap }}
-  {{- if .Values.cassandra.heap.size  }}
-      initial_heap_size: {{ .Values.cassandra.heap.size }}
-      max_heap_size: {{ .Values.cassandra.heap.size }}
+  {{- if .Values.cassandra.heap.size }}
+    {{- if (regexMatch "^(([0]\\.\\d*)+|(^[1-9]\\d*)+(\\.\\d+)?)(?i)(k|m|g|e|p|t){1}$" (print .Values.cassandra.heap.size)) }}
+      {{- nindent 6 (print "initial_heap_size: " .Values.cassandra.heap.size) }}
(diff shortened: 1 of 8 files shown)


charts/k8ssandra/templates/_helpers.tpl
```diff
@@ -117,24 +117,41 @@ Return the ingress host that should be used for Stargate's Cassandra/CQL interfa
...
 Create the jvm options based on heap properties specified.
+Expecting that c* heap.size and heap.newGenSize are NOT in IEC format.
 */}}
 {{- define "k8ssandra.configureJvmHeap" -}}
 {{- $datacenter := (index .Values.cassandra.datacenters 0) -}}
 {{- if $datacenter.heap }}
   {{- if $datacenter.heap.size }}
-      initial_heap_size: {{ $datacenter.heap.size }}
-      max_heap_size: {{ $datacenter.heap.size }}
+      {{- if (regexMatch ".+(i)" $datacenter.heap.size) }}
```
jsanda, line 126: I did a quick test and I think you might want to use the `print` function to make sure the values are cast to strings. I used the following chart properties:

```yaml
cassandra:
  version: 4.0.0
  heap:
    size: 1024
  datacenters:
  - name: dc1
```

The `helm install` did fail, but not as expected. Here is the error I got:

`
Error: template: k8ssandra/templates/cassandra/cassdc.yaml:148:4: executing "k8ssandra/templates/cassandra/cassdc.yaml" at <include "k8ssandra.configureJvmHeap" .>: error calling include: template: k8ssandra/templates/_helpers.tpl:142:38: executing "k8ssandra.configureJvmHeap" at <.Values.cassandra.heap.size>: wrong type for value; expected string; got float64
`

jsanda (changes requested):

charts/k8ssandra/templates/_helpers.tpl
jeffbanks: Right, good find!  It looks through the tests we have at current are using a [string]string for value inputs, so that is how it fell through the cracks.  I'll add that print function to the regex functions for the cassdc values passed in.

jeffbanks: Updated w/ fix for this one and tested locally with latest.

jsanda (commented):
Now my same test fails with:

```
Error: CassandraDatacenter.cassandra.datastax.com "dc1" is invalid: spec.size: Invalid value: "null": spec.size in body must be of type integer: "null"
```

jeffbanks:
@jsanda can you share your test inputs?

I am using the following for my testing as reference:
```yaml
heap:
    size: 1000M
    newGenSize: 1G
  resources:
    requests:
      cpu: 1000m
      memory: 2Gi
    limits:
      cpu: 1000m
      memory: 2Gi
  datacenters:
  - name: dc1
    size: 1
    heap:
      size: 512M
      newGenSize: 256M
    racks:
    - name: default 
```

jsanda:
Here is what I used:

```yaml
cassandra:
  version: 4.0.0
  heap:
    size: 1024foo
  datacenters:
  - name: dc1
```

charts/k8ssandra/templates/_helpers.tpl
```diff
@@ -117,24 +117,41 @@ Return the ingress host that should be used for Stargate's Cassandra/CQL interfa
...
+Expecting that c* heap.size and heap.newGenSize are NOT in IEC format.
 */}}
 {{- define "k8ssandra.configureJvmHeap" -}}
 {{- $datacenter := (index .Values.cassandra.datacenters 0) -}}
 {{- if $datacenter.heap }}
   {{- if $datacenter.heap.size }}
-      initial_heap_size: {{ $datacenter.heap.size }}
-      max_heap_size: {{ $datacenter.heap.size }}
+      {{- if (regexMatch "^(([0]\\.\\d*)+|(^[1-9]\\d*)+(\\.\\d+)?)(?i)(k|m|g|e|p|t){1}$" (print $datacenter.heap.size) ) }}
+        {{- nindent 8 (print "initial_heap_size: " $datacenter.heap.size) }}
```
jsanda, line 127: Indentation is off by 2 and the same for the other `nindent` calls.

jsanda (changes requested):

charts/k8ssandra/templates/_helpers.tpl
jeffbanks: Good catch!

To assist with any future changes in `_helpers.tpl`, I've included a test utility in the upcoming push.  It provides support for an optional flag to capture the rendered output in a file.  The option takes a destination `dir` and `name` of the rendered output file.

```json
--rendertmpl={ "dir":"/tmp/foobar","name":"cluster_heap_size_valid.yaml"}
```

Usage:
```go
    options := &helm.Options{
      SetValues: map[string]string{
        "cassandra.heap.size":           "4.5G",
	"cassandra.heap.newGenSize":     "1.5G",
	"cassandra.datacenters[0].name": dcName,
	helmUtils.OPT_RENDER_TEMPLATE:   `{ "dir":"/tmp/foobar","name":"cluster_heap_size_valid.yaml"}`,
    },
    KubectlOptions: defaultKubeCtlOptions,
}
```

jsanda (approved):
@jeffbanks Needs a rebase and then good to go!

PR #888, opened 2021-06-15
Add workflow for creating dependency upgrade PRs

### LINKED ISSUE #886: Automate PR creation for dependency upgrades
It would be nice to have PRs sent automatically when a dependency (Stargate, Reaper and Medusa) gets a new release. CI will run on the PRs and verify that new versions don't break K8ssandra.

┆Issue is synchronized with this [Jira Bug](https://k8ssandra.atlassian.net/browse/K8SSAND-592) by [Unito](https://www.unito.io)

adejanovski:
**What this PR does**:
Runs a schedule GHA workflow to send PRs for dependency upgrades (Stargate, Reaper, Medusa).

Fixes #886 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/.github/workflows/dependency-upgrade.yaml
+++ b/.github/workflows/dependency-upgrade.yaml
@@ -1,0 +1,76 @@
+---
+name: Dependency Upgrade Check
+
+on:
+  schedule:
+    # Runs daily at noon UTC (I guess)
+    - cron:  '0 12 * * *'
+
+jobs:
+  check-upgrades:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        dependency: ["Stargate", "Medusa", "Reaper"]
+        include:
+          - dependency: Stargate
+            version_path: ".stargate.version"
+            docker_hub_url: "https://registry.hub.docker.com/v2/repositories/stargateio/stargate-3_11/tags?page_size=1"
+          - dependency: Medusa
+            version_path: ".medusa.image.tag"
+            docker_hub_url: "https://registry.hub.docker.com/v2/repositories/k8ssandra/medusa/tags?page_size=25"
+          - dependency: Reaper
+            version_path: ".reaper.image.tag"
+            docker_hub_url: "https://registry.hub.docker.com/v2/repositories/thelastpickle/cassandra-reaper/tags?page_size=25"
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Create env variables
+        run: ./.github/scripts/env_variables.sh
+      - name: Update PATH
+        run: |
+          echo "$GITHUB_WORKSPACE/bin" >> $GITHUB_PATH
+      - name: Install YQ and JQ
+        run: |
+          scripts/install-yq.sh ${{ env.YQ_VERSION }}
+          yq -V
+          sudo apt-get update
+          sudo apt-get install jq -y
+          jq --version
+      - name: Get current dependencies versions
+        run: |
+          CURRENT=$(yq e '${{ matrix.version_path }}' charts/k8ssandra/values.yaml)
+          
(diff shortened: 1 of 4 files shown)


.github/workflows/dependency-upgrade.yaml
```diff
@@ -0,0 +1,72 @@
...
+          
+          echo "CURRENT_VERSION=${CURRENT}" >> $GITHUB_ENV
+          
+      - name: Get latest released versions
+        run: |
+          LATEST=$(curl -L -s '${{ matrix.docker_hub_url }}'|jq -r '.results[]["name"]'|grep -Ei "^[v]{0,1}[0-9]{1,2}\.[0-9]{1,2}\.[0-9]{1,2}$"|head -1|sed s/v//)
+
+          echo "LATEST_VERSION=${LATEST}" >> $GITHUB_ENV
+          
+      - name: Check if an open PR exists for the latest versions
```
burmanm, line 51: Won't the create-pull-request action do this? At least it says:

``If a pull request already exists and there are no further changes (i.e. no diff with the current pull request branch) then the action exits silently.``

adejanovski: I've seen that, yes, but I was afraid it may reopen a PR that we close if we don't want to perform the upgrade 🤔 
Let me try this actually and we'll see if it doesn't try to reopen closed PRs.

.github/workflows/dependency-upgrade.yaml
```diff
@@ -0,0 +1,72 @@
...
+            docker_hub_url: "https://registry.hub.docker.com/v2/repositories/thelastpickle/cassandra-reaper/tags?page_size=25"
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Update PATH
+        run: |
+          echo "$GITHUB_WORKSPACE/bin" >> $GITHUB_PATH
+      - name: Install YQ and JQ
+        run: |
+          scripts/install-yq.sh v4.6.3
```
jeffbanks, line 34: I'm wondering since the version references and other value constants are increasing across the .yaml files under the `workflows` directory, if it would be beneficial to have them defined in one place?  Something similar to the approach we are thinking about with `image` repositories.

.github/workflows/dependency-upgrade.yaml
```diff
@@ -0,0 +1,72 @@
...
+      - name: Check if a closed PR exists for the same version
+        run: |
+          CLOSED_UPGRADE_PR=$(curl -L -s "https://api.github.com/repos/k8ssandra/k8ssandra/pulls?state=closed"|jq -c '.[] | select( .title == "Upgrade ${{ matrix.dependency }} dependency ${{ env.CURRENT_VERSION }} -> ${{ env.LATEST_VERSION }}" )'|jq -r '.title')
+          
+          if [ -z "$CLOSED_UPGRADE_PR" ] && [ "${{ env.CURRENT_VERSION }}" != "${{ env.LATEST_VERSION }}" ]; then
+            echo "PR_EXISTS=no" >> $GITHUB_ENV
+            yq eval "${{ matrix.version_path }} |= \"${{ env.LATEST_VERSION }}\"" charts/k8ssandra/values.yaml -i
+            cat charts/k8ssandra/values.yaml
+          fi
+
```
jeffbanks, line 60: Consider echo out that the PR exists to be consistent with the other run operations.

jeffbanks (approved):
LGTM, a couple of comments added for consideration.

.github/workflows/dependency-upgrade.yaml
adejanovski: Done 👍

.github/workflows/dependency-upgrade.yaml
adejanovski: Very good point!
I created a script that sets the versions in env variables and can be called from the different workflows.

PR #894, opened 2021-06-16
Use personal access token to trigger CI

adejanovski:
Using the repository default token will not trigger CI as intended by GH.
Switching to a Personal Access Token will trigger CI on PR creation by the dependency upgrade workflow.


adutra (approved):
LGTM I see that you created the CREATE_PR_TOKEN token accordingly.

PR #895, opened 2021-06-16
Upgrade Reaper dependency 2.2.2 -> 2.2.5

adejanovski:

### DIFF
--- a/CHANGELOG-1.3.md
+++ b/CHANGELOG-1.3.md
@@ -18,6 +18,7 @@ and date `## vX.Y.Z - YYYY-MM-DD` and create a new placeholder section for  `unr
 
 ## unreleased
 
+* [CHANGE] Upgrade from Reaper 2.2.2 to 2.2.5
 * [CHANGE] #812 Integrate Fossa component/license scanning
 * [BUGFIX] #853 Fix property name in scaling docs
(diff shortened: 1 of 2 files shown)


jsanda (commented):
How come it introduced all the formatting changes?

Do you think we should also change the default version in reaper-operator?

adejanovski:
> How come it introduced all the formatting changes?

The values files update is done using yq, which seems to reformat the file according to its own standard. I guess once we merge one of these PRs, we'll be good.
 
> Do you think we should also change the default version in reaper-operator?

I'll answer with more questions: do you think we should include the operators in the list of upgraded dependencies so that we get upgrade PRs for them as well? Should we upgrade both Reaper and reaper-operator at the same time then or separately? Same question for medusa/medusa-operator.

jsanda:
Thinking out loud... 

I would like to say we do Reaper --> reaper-operator --> k8ssandra and same for Medusa. The reality is that we have end to end test coverage in k8ssandra whereas it is lacking in reaper-operator and in medusa-operator. And even if we had more thorough test coverage in the operator repos, the tests in k8ssandra are authoritative. 

I am in favor of Reaper -->k8ssandra --> reaper-operator.

adejanovski:
> I am in favor of Reaper -->k8ssandra --> reaper-operator.

Sounds good

PR #897, opened 2021-06-17
Improve post-upgrade checks

adejanovski:
**What this PR does**:
Fixes failing upgrade test by giving some time for cass-operator to update the cassdc object in case of an upgrade. Since this state change is optional (sometimes the source/target versions are too close for them to trigger an update), the condition is not failing the whole test if it's not fulfilled.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/tests/integration/steps/integration_steps.go
+++ b/tests/integration/steps/integration_steps.go
@@ -22,6 +22,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	wait "k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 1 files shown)


tests/integration/steps/integration_steps.go
```diff
@@ -249,6 +250,13 @@ func waitForCassDcUpgrade(t *testing.T, namespace, initialResourceVersion string
 		Namespace: namespace,
 	}
 
+	// Wait for at most 1 minute, waiting for cassdc to move to Upgrading state.
+	// This may not happen if the upgrade doesn't trigger a rolling restart.
+
+	wait.Poll(10*time.Second, 1*time.Minute, func() (bool, error) {
```
adutra, line 256: So I understand you can't use `g(t).Eventually()` because you would like to resume test execution.

But taking a step back: shouldn't the caller of this function (or the scenario being run) provide the missing information here? That is, whether we should block and wait for the state to change?

adejanovski: Well, the thing is we don't know if the state will change or not. It depends on whether or not the upgrade from an older K8ssandra version to the current one will change the spec of the cassandra pods.
That's why we wait a little bit for the status change to happen and continue if it doesn't.

adutra: OK-ish. I still find it a bit odd but don't want to hold you here. Approving.

adutra (approved):

PR #903, opened 2021-06-22
Make the service monitors namespace configurable

### LINKED ISSUE #844: Allow configuration of ServiceMonitor namespace
I already have a prometheus stack installed (In the `monitoring` namespace) so I disabled K8ssandra's kube-prometheus-stack. However it doesn't pick up the `ServiceMonitor` as it's deployed in the wrong namespace.    Would it be possible to define the namespace the `ServiceMonitor` is defined in?

Usually charts expose a `metrics.serviceMonitor.namespace` option to specify where the chart's servicemonitor is installed. See e.g. https://github.com/bitnami/charts/blob/395934bfc46804f5aa1b96d2fca416c0d5bfa02e/bitnami/rabbitmq/templates/servicemonitor.yaml#L6

None

**Additional context**

┆Issue is synchronized with this [Jira Feature](https://k8ssandra.atlassian.net/browse/K8SSAND-499) by [Unito](https://www.unito.io)
┆Fix Versions: k8ssandra-1.3.0

adejanovski:
**What this PR does**:
Allows to configure the namespace in which the service monitors are deployed.

Fixes #844

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/CHANGELOG-1.3.md
+++ b/CHANGELOG-1.3.md
@@ -23,6 +23,7 @@ and date `## vX.Y.Z - YYYY-MM-DD` and create a new placeholder section for  `unr
 * [CHANGE] #905 Upgrade medusa-operator to v0.3.3
 * [FEATURE] #617 Make affinity configurable for Stargate
 * [FEATURE] #847 Make affinity configurable for Reaper
+* [ENHANCEMENT] #844 Allow configuring the namespace of service monitors
 * [ENHANCEMENT] #29 Detect IEC formatted c* heap.size and heap.newGenSize; return error identifying issue  
 * [ENHANCEMENT] #420 Add support for private registries
 * [BUGFIX] #853 Fix property name in scaling docs
--- a/charts/k8ssandra/templates/prometheus/service_monitor.yaml
+++ b/charts/k8ssandra/templates/prometheus/service_monitor.yaml
@@ -6,6 +6,9 @@ metadata:
   labels:
     release: {{ .Release.Name }}
 {{ include "k8ssandra.labels" . | indent 4 }}
+{{- if .Values.monitoring.serviceMonitors }}
+  namespace: {{ default .Release.Namespace .Values.monitoring.serviceMonitors.namespace }}
+{{- end }}
 spec:
   selector:
     matchLabels:
@@ -14,6 +17,9 @@ spec:
     matchExpressions:
       - {key: cassandra.datastax.com/cluster, operator: In, values: [{{ include "k8ssandra.clusterName" . }}]}
       - {key: cassandra.datastax.com/datacenter, operator: In, values: [{{ include "k8ssandra.datacenterName" . }}]}
+  namespaceSelector:
+    matchNames:
+      - {{ .Release.Namespace }}
   targetLabels:
   - cassandra.datastax.com/cluster
   - cassandra.datastax.com/datacenter
--- a/charts/k8ssandra/templates/stargate/service_monitor.yaml
+++ b/charts/k8ssandra/templates/stargate/service_monitor.yaml
@@ -7,10 +7,16 @@ metadata:
     release: {{ .Release.Name }}
     app: {{ .Release.Name }}-{{ include "k8ssandra.datacenterName" . }}-stargate
 {{ include "k8ssandra.labels" . | indent 4 }}
+{{- if .Values.monitoring.serviceMonitors }}
+  namespace: {{ default .Release.Namespace .Values.monitoring.serviceMonitors.namespace }}
+{{- end }}
 spec:
   selector:
     matchLabels:
       app: {{ .Release.Name }}-{{ include "k8ssandra.datacenterName" . }}-stargate
+  namespaceSelector:
+    matchNames:
+      - {{ .Release.Namespace }}
   endpoints:
   - port: health
     interval: 15s
--- a/charts/k8ssandra/values.yaml
+++ b/charts/k8ssandra/values.yaml
@@ -618,6 +618,10 @@ monitoring:
     # not have the ServiceMonitor CRD installed on your cluster, set this value
     # to `false`.
     provision_service_monitors: true
+  # -- Deploy the service monitors on a different namespace than the release namespace.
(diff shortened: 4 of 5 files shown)


CHANGELOG-1.3.md
```diff
@@ -18,6 +18,7 @@ and date `## vX.Y.Z - YYYY-MM-DD` and create a new placeholder section for  `unr
 
 ## unreleased
 
+* [ENHANCEMENT] Allow configuring the namespace of service monitors
```
adutra, line 21: Nit: enhancements should come after changes. Also, I wonder if we need to include the GH issue number and/or Jira ticket number.

charts/k8ssandra/templates/prometheus/service_monitor.yaml
```diff
@@ -6,6 +6,11 @@ metadata:
   labels:
     release: {{ .Release.Name }}
 {{ include "k8ssandra.labels" . | indent 4 }}
+  {{- if .Values.monitoring }}
```
jsanda, line 9: This can be simplified as:

```
  {{- if .Values.monitoring.serviceMonitor }}
  namespace: {{ default .Release.Namespace .Values.monitoring.serviceMonitor.namespace }}
  {{- end }}
```

charts/k8ssandra/values.yaml
```diff
@@ -537,6 +537,10 @@ monitoring:
     # not have the ServiceMonitor CRD installed on your cluster, set this value
     # to `false`.
     provision_service_monitors: true
+  # -- Deploy the service monitors on a different namespace than the release namespace.
+  # This should only be set if the kube-prometheus stack isn't installed in the release namespace
+  serviceMonitor: {}
```
adutra, line 542: Nit: shouldn't we use `serviceMonitors` in the plural here, since we can have up to 2 service monitors deployed?

charts/k8ssandra/templates/prometheus/service_monitor.yaml
```diff
@@ -14,6 +19,9 @@ spec:
     matchExpressions:
       - {key: cassandra.datastax.com/cluster, operator: In, values: [{{ include "k8ssandra.clusterName" . }}]}
       - {key: cassandra.datastax.com/datacenter, operator: In, values: [{{ include "k8ssandra.datacenterName" . }}]}
+  namespaceSelector:
+    matchNames:
+      - {{ .Release.Namespace | quote }}
```
jsanda, line 24: This can be simplified as `{{ .Release.Namespace }}`. No need for the `quote` function.

charts/k8ssandra/templates/stargate/service_monitor.yaml
```diff
@@ -7,10 +7,18 @@ metadata:
     release: {{ .Release.Name }}
     app: {{ .Release.Name }}-{{ include "k8ssandra.datacenterName" . }}-stargate
 {{ include "k8ssandra.labels" . | indent 4 }}
+  {{- if .Values.monitoring }}
```
jsanda, line 10: This can be simplified like the previous one.

charts/k8ssandra/templates/stargate/service_monitor.yaml
```diff
@@ -7,10 +7,18 @@ metadata:
...
+  namespace: {{ .Values.monitoring.serviceMonitor.namespace | default (.Release.Namespace | quote) }}
+  {{- end }}
+  {{- end }}
 spec:
   selector:
     matchLabels:
       app: {{ .Release.Name }}-{{ include "k8ssandra.datacenterName" . }}-stargate
+  namespaceSelector:
+    matchNames:
+      - {{ .Release.Namespace | quote }}
```
jsanda, line 21: Same as the previous one, no need for the `quote` function.

adutra (approved):
Looking good, I only have very minor remarks 👍

charts/k8ssandra/values.yaml
jsanda, line 542: An empty map here suggests that I can specify anything or a ServiceMonitor in its entirety. I would change this based on the properties that you are using - `namespace` and `namespaceSelector`.

jsanda (changes requested):

charts/k8ssandra/templates/prometheus/service_monitor.yaml
adejanovski: I was getting nil errors prior to adding `serviceMonitor: {}` in the default values file, because trying to test the presence of `.Values.monitoring.serviceMonitor.namespace` required to first make sure `.Values.monitoring.serviceMonitor` existed as it was not mandatory.
I'll apply your suggestion since it's the case now and a dual check is unnecessary.

charts/k8ssandra/templates/prometheus/service_monitor.yaml
adejanovski: gotcha 👍

CHANGELOG-1.3.md
adejanovski: I actually thought we were adding updates in reverse chronological order 🤔  aren't we?

charts/k8ssandra/values.yaml
adejanovski: yup, good point 👍

adejanovski: How can I make it so that the dict exists even if the namespace value isn't provided then?
I've replicated what's right above for `medusa.podStorage` or `medusa.storage_properties`.

CHANGELOG-1.3.md
adutra: My understanding is that we have 2 ordering keys: by type first, then by reverse chronological order.

adejanovski: Scratch that, I just read the header section for the first time with the right level of attention 😅

charts/k8ssandra/values.yaml
adejanovski: It should look better now. Let me know what you think.

jsanda (approved):

PR #950, opened 2021-07-08
Modify to use the cass-operator v1.8.0-rc.1

burmanm:
**What this PR does**:
Modifications for the cass-operator based on the next kubebuilder struct.

Fixes #<issue number>

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/charts/cass-operator/templates/deployment.yaml
+++ b/charts/cass-operator/templates/deployment.yaml
@@ -25,6 +25,10 @@ spec:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
       containers:
         - name: {{ .Chart.Name }}
+          args:
+          - --config=/configs/controller_manager_config.yaml
+          command:
+          - /manager
           securityContext:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: {{ include "k8ssandra-common.flattenedImage" .Values.image }}
--- a/charts/cass-operator/templates/deployment.yaml
+++ b/charts/cass-operator/templates/deployment.yaml
@@ -36,7 +36,7 @@ spec:
           {{- if .Values.admissionWebhooks.enabled }}
           ports:
             - containerPort: 9443
-              name: cass-operator-webhook-server
+              name: webhook-server
               protocol: TCP
           {{- end }}
           volumeMounts:
--- a/charts/cass-operator/templates/deployment.yaml
+++ b/charts/cass-operator/templates/deployment.yaml
@@ -29,31 +29,28 @@ spec:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: {{ include "k8ssandra-common.flattenedImage" .Values.image }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          {{- if .Values.admissionWebhooks.enabled }}
+          ports:
+            - containerPort: 9443
+              name: cass-operator-webhook-server
+              protocol: TCP
           volumeMounts:
             - mountPath: /tmp/k8s-webhook-server/serving-certs
               name: cass-operator-certs-volume
-              readOnly: false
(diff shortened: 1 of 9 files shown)


burmanm:
Don't merge yet.

charts/cass-operator/values.yaml
```diff
@@ -33,7 +33,7 @@ image:
   pullPolicy: IfNotPresent
 
   # -- Tag of the cass-operator image to pull from image.repository
-  tag: v1.7.1
+  tag: latest
```
jsanda, line 36: Would it be better to pin to a specific commit hash until 1.8.0 is released?

burmanm:
@jsanda

charts/cass-operator/templates/deployment.yaml
```diff
@@ -25,48 +25,47 @@ spec:
...
+          command:
+          - /manager
           securityContext:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: {{ include "k8ssandra-common.flattenedImage" .Values.image }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          {{- if .Values.admissionWebhooks.enabled }}
+          ports:
+            - containerPort: 9443
+              name: cass-operator-webhook-server
```
jsanda, line 39: Trying to install with the webhook enabled fails:

`helm install cass-operator ./charts/cass-operator --set admissionWebhooks.enabled=true
Error: Deployment.apps "cass-operator" is invalid: spec.template.spec.containers[0].ports[0].name: Invalid value: "cass-operator-webhook-server": must be no more than 15 characters`

jsanda: I changed the port name locally to `webhook-server` so I could get further with the install and see additional errors. cert-manager is a dependency now, at least when the webhook is enabled. The operator is a crash loop due to this error in the log:

```
2021-09-01T02:07:06.358Z	INFO	controller-runtime.webhook.webhooks	starting webhook server
2021-09-01T02:07:06.359Z	ERROR	controller-runtime.manager	error received after stop sequence was engaged	{"error": "context canceled"}
2021-09-01T02:07:06.360Z	ERROR	controller-runtime.manager	error received after stop sequence was engaged	{"error": "context canceled"}
E0901 02:07:06.360592       1 leaderelection.go:325] error retrieving resource lock default/b569adb7.cassandra.datastax.com: Get "https://10.96.0.1:443/api/v1/namespaces/default/configmaps/b569adb7.cassandra.datastax.com": context canceled
2021-09-01T02:07:06.360Z	ERROR	controller-runtime.manager	error received after stop sequence was engaged	{"error": "leader election lost"}
2021-09-01T02:07:06.360Z	ERROR	setup	problem running manager	{"error": "tls: failed to find any PEM data in certificate input"}
main.main
```
If we are going to support running the webhook, then I think we should have some additional test coverage for when it is enabled.

jsanda (changes requested):

charts/cass-operator/templates/deployment.yaml
burmanm: So the webhook enabled has never worked (or for a long time) ? The ports name shouldn't have changed ..

burmanm: Actually, now that I think about it.. we know it hasn't worked in k8ssandra (and it's not exposed in k8ssandra/k8ssandra chart) since it required cluster scope rights to be usable.

jsanda (changes requested):
The 4.0.1 image name in k8ssandra/values.yaml is wrong. It should be `k8ssandra/cass-mgmtapi:4.0.1-v0.1.29` not `k8ssandra/cass-management-api:4.0.1-v0.1.29`.

jsanda (approved):

Issue #985, opened 2021-07-16
The service account for the crd-upgrader pre-upgrade hook should be configurable


jsanda:
This is needed for private registry support and is a follow up to #901.

┆Issue is synchronized with this [Jira Bug](https://k8ssandra.atlassian.net/browse/K8SSAND-727) by [Unito](https://www.unito.io)

Issue #984, opened 2021-07-16
The service account for the cleaner pre-delete hook should be configurable


jsanda:
The service account for the cleaner pre-delete chart hook should be configurable. This should have been done as part of #901 but I missed it.

PR #901, opened 2021-06-21
add support for private registries 

jsanda:
**What this PR does**:
Adds support for private registries by making image registry, service accounts, and image pull secrets configurable.

Summary of changes:
* **All images** deployed are configurable with the following properties:
    * registry
    * repository
    * tag
    * pullPolicy 
* Add `cassandra.serviceAccount` to make the service account for Cassandra pods configurable
* Make service account for Stargate pods configurable
* Make image pull secrets for service accounts in operator charts configurable
* Image pull secrets can be configured via exposed service accounts and with the default service account for places where service account is not explicitly set

**Which issue(s) this PR fixes**:
Fixes #420, #839, #840

**Checklist**
- [x] Changes manually tested
- [x] Automated Tests added/updated
- [ ] Documentation added/updated
- [x] CHANGELOG.md updated (not required for documentation PRs)
- [x] CLA Signed:  [DataStax CLA](https://cla.datastax.com/)

PR #986, opened 2021-07-16
make service accounts configurable for reaper, cleaner, and crd upgrader

### LINKED ISSUE #983: Reaper service account is not configurable
The Reaper CRD allows you to specify the name of the service account to use for the Reaper pods, but it is not configurable in k8ssandra.

This should have been done in PR #901 but I missed it.

┆Issue is synchronized with this [Jira Bug](https://k8ssandra.atlassian.net/browse/K8SSAND-725) by [Unito](https://www.unito.io)

jsanda:
**What this PR does**:
Makes the service account to use configurable. It does not however allow you to configure the service account. This just give you the ability to point to the one you want to use.

I did not update the changelog for this since this PR really should have been part of #901.

Ideally, for the cleaner and crd upgrader, I would have liked to use the `k8ssandra-common.serviceAccount` template. It makes the service account name as well as image pull secrets available. It would require some refactoring to be able to safely use it for them though. The template assumes the scope of `.Values`. This would need to change and certainly beyond the scope of 1.3.0 at this point.

Fixes #983, #984, #985

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/charts/k8ssandra/values.yaml
+++ b/charts/k8ssandra/values.yaml
@@ -468,6 +468,8 @@ reaper:
   # will be configured in Cassandra with credentials only created for Reaper
   # in order to limit access.
   enabled: true
+  # -- The name of the service account to use for Reaper pods.
+  serviceAccount: default
   # Configures the Reaper container image to use. Exercise care when changing
   # the Reaper image. Reaper is deployed and managed by reaper-operator. You
   # will need to make sure that the image is compatible with reaper-operator.
@@ -556,7 +558,7 @@ medusa:
   multiTenant: false
   # -- API interface used by the object store. Supported values include `s3`, 's3_compatible',
   # `google_storage` and 'azure_blobs'. For file system storage, i.e., a pod volume
-  # mount, use 'local' and set the podStorage properties. Note that 'local' does not necessarily 
+  # mount, use 'local' and set the podStorage properties. Note that 'local' does not necessarily
   # imply a local volume. It could also be network attached storage. It is simply accessed
   # through the file system.
   storage: s3
@@ -621,6 +623,9 @@ monitoring:
 # deployment could be deleted before the CassandraDatacenter. The cleaner
 # ensures that the CassandraDatacenter is deleted before cass-operator.
 cleaner:
+  # -- Uncomment to specify the name of the service account to use for the
+  # cleaner. Defaults to <release-name>-cleaner-k8ssandra
+  # serviceAccount:
   image:
     # -- Image registry for the cleaner
     registry: docker.io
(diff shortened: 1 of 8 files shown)


charts/k8ssandra/values.yaml
```diff
@@ -468,6 +468,8 @@ reaper:
   # will be configured in Cassandra with credentials only created for Reaper
   # in order to limit access.
   enabled: true
+  # -- The name of the service account to use for Reaper pods.
+  serviceAccount: default
```
jeffbanks, line 472: serviceAccount is commented out for cleaner and client and is set to empty string for c*.  Double checking that you want Reaper to have the default value or should this be commented out as well?

jsanda: Good observations. To make things a little more consistent, I can default them all to the empty string. In the case of Reaper, Stargate, and Cassandra, this will result in using the `default` service account. In the case of the cleaner and client this would result in the creation of their respective service accounts. Unfortunately there are some inconsistencies that would be nice to address, but I think that would need to be post 1.3.0.

There is another problem with the cleaner and client service accounts. I explained in the description why I didn't use the `k8ssandra-common.serviceAccount` template. If the user specifies a different service account I need to update the RoleBinding template to use the specified ones.

jsanda: I thought about this some more, and here things I think we need to do post 1.3.0:

* Refactor the `k8ssandra-common.serviceAccount` template so it doesn't assume a scope of `.Values`. This will make it more flexible and easier to use elsewhere.
    * Service accounts created with the `k8ssandra-common.serviceAccount` can have both the service account name and image pull secrets configured.
* Create a service account for Cassandra pods using the `k8ssandra-common.serviceAccount` template
    * Allow the user to specify a different service account name to use
* Create a service account for Stargate pods using the `k8ssandra-common.serviceAccount` template
    * Allow the user to specify a different service account name to use
* Create a service account for Reaper pod using the `k8ssandra-common.serviceAccount` template
    * Allow the user to specify a different service account name to use
* Create a service account for cleaner pod using the `k8ssandra-common.serviceAccount` template
    * Allow the user to specify a different service account name to use 
* Create a service account for crd upgrader (i.e., client) pod using the `k8ssandra-common.serviceAccount` template
    * Allow the user to specify a different service account name to use 

In summary, we should be creating a service account for every pod that we are deploying and it should be done using the `k8ssandra-common.serviceAccount` template.

@jdonenine I definitely think we need to get these ticketed. I don't know how much bandwidth we have to address them though with everything else that is in flight.

jdonenine: @jsanda got it... I'll take what you laid out here and get it into some issues to consider post-1.3

jeffbanks (approved):
@jsanda the initial set of changes LGTM for this scope of work.

PR #1007, opened 2021-07-27
Security enhancements

jeffbanks:
**What this PR does**:
Security context additions for K8ssandra containers and pods.

Fixes k8ssand-635 and k8ssand-732 as part of the [k8ssandra security epic](https://k8ssandra.atlassian.net/browse/K8SSAND-705)

**Checklist**

* [x] Changes manually tested
* [x] Automated Tests added/updated
* [ ] Documentation added/updated
* [x] CHANGELOG.md updated (not required for documentation PRs)
* [x] CLA Signed:  [DataStax CLA](https://cla.datastax.com/)

┆Issue is synchronized with this [Jiraserver Task](https://k8ssandra.atlassian.net/browse/K8SSAND-762) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-762
┆Priority: Medium

### DIFF
--- a/charts/k8ssandra/templates/_helpers.tpl
+++ b/charts/k8ssandra/templates/_helpers.tpl
@@ -369,3 +369,21 @@ Cassandra image (Management API image and version tag).
 {{- (printf (get .Values.cassandra.versionImageMap .Values.cassandra.version)) }}
 {{- end }}
 {{- end }}
+
+{{/**
+SecurityContext helpers for pod and container scope.
+*/}}
+
+{{/**
+k8ssandra container securityContext defaults
+**/}}
+{{- define "k8ssandra.container.security-ctx-defaults" -}}
+readOnlyRootFilesystem: true
+{{- end }}
+
+{{/**
+k8ssandra Cassandra pod's podSecurityContext defaults
+**/}}
+{{- define "k8ssandra.pod.pod-security-ctx-defaults" -}}
+securityContext: {}
+{{- end }}
--- a/charts/k8ssandra/templates/cassandra/cassdc.yaml
+++ b/charts/k8ssandra/templates/cassandra/cassdc.yaml
@@ -1,5 +1,6 @@
 {{- $datacenter := (index .Values.cassandra.datacenters 0) -}}
 {{- $medusaImage := (printf "%s:%s" .Values.medusa.image.repository .Values.medusa.image.tag) -}}
+{{- $defaultSecurityCtx := (include "k8ssandra.container.security-ctx-defaults" .) -}}
 
 {{- if and (not .Values.cassandra.image)  (not (hasKey .Values.cassandra.versionImageMap .Values.cassandra.version)) -}}
   {{- fail (print .Values.cassandra.version " is not a supported Cassandra version") }}
@@ -9,7 +10,6 @@
   {{- fail (print "If medusa storage type is set to local, the podStorage storageClass and size must be set") }}
 {{- end -}}
 {{- if .Values.cassandra.enabled -}}
-
 
 apiVersion: cassandra.datastax.com/v1beta1
 kind: CassandraDatacenter
@@ -161,10 +161,12 @@ spec:
 {{- end }}
   podTemplateSpec:
     spec:
+      securityContext: {{- ternary ($defaultSecurityCtx) (.Values.cassandra.podSecurityContext | toYaml) (empty .Values.cassandra.podSecurityContext) | nindent 10 }}
       initContainers:
       - name: base-config-init
         image: {{ include "k8ssandra.cassandraImage" . }}
         imagePullPolicy: IfNotPresent
+        securityContext: {{- ternary ($defaultSecurityCtx) (.Values.cassandra.baseConfig.securityContext | toYaml) (empty .Values.cassandra.baseConfig.securityContext) | nindent 10 }}
         command:
           - /bin/sh
         args:
@@ -178,10 +180,12 @@ spec:
           - name: cassandra-tmp
             mountPath: /tmp/
       - name: server-config-init
+        securityContext: {{- ternary ($defaultSecurityCtx) (.Values.cassandra.configBuilder.securityContext | toYaml) (empty .Values.cassandra.configBuilder.securityContext) | nindent 10 }}
      {{- if (or .Values.cassandra.auth.enabled .Values.reaper.enabled) }}
       - name: jmx-credentials
         image: {{ include "k8ssandra-common.flattenedImage" .Values.cassandra.jmxCredentialsConfig.image }}
         imagePullPolicy: {{ .Values.cassandra.jmxCredentialsConfig.image.pullPolicy }}
+        securityContext: {{- ternary ($defaultSecurityCtx) (.Values.cassandra.jmxCredentialsConfig.securityContext | toYaml) (empty .Values.cassandra.jmxCredentialsConfig.securityContext) | nindent 10 }}
         env:
         {{- if .Values.reaper.enabled }}
           - name: REAPER_JMX_USERNAME
@@ -220,6 +224,11 @@ spec:
       {{- if .Values.medusa.enabled }}
       - name: medusa-restore
         image: {{ include "k8ssandra-common.flattenedImage" .Values.medusa.image }}
+        {{- if (and .Values.medusa.restoreInitContainerConfig .Values.medusa.restoreInitContainerConfig.securityContext) }}
+        securityContext: {{ .Values.medusa.restoreInitContainerConfig.securityContext | toYaml | nindent 10 }}
+        {{- else }}
+        securityContext: {{ include "k8ssandra.container.security-ctx-defaults" . | nindent 10 }}
+        {{- end }}
         imagePullPolicy: {{ .Values.medusa.image.pullPolicy }}
         env:
           - name: MEDUSA_MODE
@@ -247,6 +256,7 @@ spec:
       {{- end }}
       containers:
       - name: cassandra
+        securityContext: {{- ternary ($defaultSecurityCtx) (.Values.cassandra.securityContext | toYaml) (empty .Values.cassandra.securityContext) | nindent 10 }}
         {{- if .Values.reaper.enabled }}
         env:
           - name: LOCAL_JMX
@@ -262,6 +272,7 @@ spec:
       {{- if .Values.medusa.enabled }}
       - name: medusa
         image: {{ include "k8ssandra-common.flattenedImage" .Values.medusa.image }}
+        securityContext: {{- ternary ($defaultSecurityCtx) (.Values.medusa.securityContext | toYaml) (empty .Values.medusa.securityContext) | nindent 10 }}
         imagePullPolicy: {{ .Values.medusa.image.pullPolicy }}
         ports:
           - containerPort: 50051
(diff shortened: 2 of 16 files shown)


jeffbanks:
Current PR includes:

Container-level and pod-level securityContext additions, defaulting to the following unless otherwise specified as overrides in values.yaml.

```yaml
 securityContext:
    readOnlyRootFilesystem: true
```
Medusa and cassandra container securityContext temporarily allows root writes to accommodate current image users
```yaml
securityContext:
    readOnlyRootFilesystem: false
```
Unit and integration tests updated to reflect these changes

jeffbanks:
@jsanda the current scope includes the k8ssand tickets: 695, 732, 696 as reference.

charts/k8ssandra/templates/_helpers.tpl
```diff
@@ -369,3 +369,82 @@ Cassandra image (Management API image and version tag).
...
+{{- define "k8ssandra.pod.pod-security-ctx-defaults" -}}
+{{ "securityContext: {}" -}}
+{{- end }}
+
+{{/**
+Container securityContext override checks.
+If specified securityContext, uses what is provided in values yaml.
+Otherwise, applies k8ssandra defaults as defined in security-ctx-defaults.
+**/}}
+{{- define "k8ssandra.container.security-ctx-base-config-init" -}}
```
jsanda, line 397: You can reduce all of these to a single template. You calling them with root scope. You just need to change the scope to the container in question. For example instead of,

```
{{ include "k8ssandra.container.security-ctx-base-config-init" . }}
```

you would instead have

```
{{ include "k8ssandra.container.security-ctx-base-config-init" .Values.baseConfig }}
```
This sets the variable scope of the template to an object that has a `securityContext` property.

jsanda (changes requested):
@jeffbanks did you run into any problems with any of the containers by making the root file system read-only?

charts/k8ssandra/templates/_helpers.tpl
jeffbanks: @jsanda
Are you thinking that for example, in the podTemplateSpec of cassdc, we would instead use this type of include?

```yaml
securityContext: {{ include "k8ssandra.container.security-ctx-base-config-init" .Values.cassandra.baseConfig.securityContext }}
```

jsanda: Yup. I had a mistake in my previous comment. Instead of `.Values.baseConfig` it should have been `.Values.cassandra.baseConfig`.

I would renamed the template to something more generic like `k8ssandra.container.security-ctx` since it would be used across multiple containers.

jeffbanks: @jsanda 

I did some digging into sprig and the helm capabilities.  I was, as you suggested, able to roll some of this up into less verbose logic for the helpers.  Thanks for suggesting this as I took a direct approach due to many errors that were cropping up.  I've added that update in the latest, and it looks easier to reason about IMO.

Right now, the security context for root read-only is relaxed for the `cassandra` and `mgmt-api` images to allow a root user to perform writes and reads.  Just those two containers' securityContext settings are impacted at this time.

For example, the `c* container` was failing to start due to errors such as:

```text
Starting Management API
cp: cannot create regular file '/opt/cassandra/conf/cassandra-env.sh': Read-only file system
cp: cannot create regular file '/opt/cassandra/conf/cassandra-rackdc.properties': Read-only file system
cp: cannot create regular file '/opt/cassandra/conf/cassandra.yaml': Read-only file system
cp: cannot create regular file '/opt/cassandra/conf/jmxremote.password': Read-only file system
cp: cannot create regular file '/opt/cassandra/conf/jvm.options': Read-only file system
cp: cannot create regular file '/opt/cassandra/conf/logback.xml': Read-only file system
```

As you know, we have some other tickets that require effort to get the images utiliized by k8ssandra to run as a non-root user.  For now, this is a starting point for getting k8ssandra securityContext configurations at the container and pod levels.  They are implemented along with the recent `volume` and `volume` mount changes as part of the overall security [epic](https://k8ssandra.atlassian.net/browse/K8SSAND-705).

charts/k8ssandra/templates/cassandra/cassdc.yaml
```diff
@@ -161,10 +161,12 @@ spec:
 {{- end }}
   podTemplateSpec:
     spec:
+      securityContext: {{- ternary ($defaultSecurityCtx) (.Values.cassandra.podSecurityContext | toYaml) (empty .Values.cassandra.podSecurityContext) | nindent 10 }}
```
jsanda, line 164: The template here still pretty redundant and could be extracted into a single helper template that would eliminate a good bit of duplication.

charts/k8ssandra/values.yaml
```diff
@@ -35,8 +35,26 @@ cassandra:
...
+
+  # -- Security context override for pod where Cassandra container resides
+  podSecurityContext:
+    readOnlyRootFilesystem: false
+
+  # -- Cassandra base init container
+  baseConfig:
+    # -- Security context override for base init container
+    # K8ssandra defaults readOnlyRootFilesystem: true
+    securityContext: {}
```
jsanda, line 52: Are there security settings that won't work with this container? If so we should either document it or not allow it. I'm fine with just documenting it and doing that in a follow PR if that works for you.

charts/k8ssandra/values.yaml
```diff
@@ -35,8 +35,26 @@ cassandra:
...
   # -- The server-config-init init container
   configBuilder:
+    # -- Security context override for server-config-init container
+    # K8ssandra defaults readOnlyRootFilesystem: true
+    securityContext: {}
```
jsanda, line 57: Are there security settings that won't work with this container? If so we should either document it or not allow it. I'm fine with just documenting it and doing that in a follow PR if that works for you.

charts/k8ssandra/values.yaml
```diff
@@ -48,6 +66,9 @@ cassandra:
       pullPolicy: IfNotPresent
   # -- The jmx-credentials init container that configures JMX credentials.
   jmxCredentialsConfig:
+    # -- Security context override for jmx init container
+    # K8ssandra defaults readOnlyRootFilesystem: true
+    securityContext: {}
```
jsanda, line 71: Are there security settings that won't work with this container? If so we should either document it or not allow it. I'm fine with just documenting it and doing that in a follow PR if that works for you.

charts/k8ssandra/values.yaml
```diff
@@ -457,6 +479,17 @@ stargate:
   # https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/ for background.
   tolerations: []
 reaper:
+  # -- Security context override for reaper container
```
jsanda, line 482: Are there security settings that won't work with these containers? If so we should either document it or not allow it. I'm fine with just documenting it and doing that in a follow PR if that works for you.

charts/k8ssandra/values.yaml
```diff
@@ -533,6 +566,20 @@ medusa:
   # -- Enable Medusa resources as part of this release. If enabled,
   # `bucketName` and `storageSecret` **must** be defined.
   enabled: false
+
+  # -- Security context override for Medusa container.
+  # K8ssandra defaults readOnlyRootFilesystem: true
+  # TODO - overriding to disable default for now, due to medusa write perm needed
+  securityContext:
```
jsanda, line 573: Are there security settings that won't work with this container? If so we should either document it or not allow it. I'm fine with just documenting it and doing that in a follow PR if that works for you.

jeffbanks:
Created a [ticket](https://k8ssandra.atlassian.net/browse/K8SSAND-827) to track documenting and validation of the entire scope of securityContext for pod and container levels.  There are numerous permutations to fully document and approve those properties for all containers and the images used in K8ssandra. The original scope of this PR was to provide a default securityContext to restrict root user writes.

jsanda:
>The original scope of this PR was to provide a default securityContext to restrict root user writes.

Thanks for the clarification.

charts/k8ssandra/templates/reaper/reaper.yaml
```diff
@@ -32,4 +32,22 @@ spec:
   {{- if .Values.reaper.tolerations }}
   tolerations: {{ toYaml .Values.reaper.tolerations | nindent 4 }}
   {{- end }}
+  schemaInitContainerConfig:
+    {{- if (and .Values.reaper.schemaInitContainerConfig .Values.reaper.schemaInitContainerConfig.securityContext) }}
```
jsanda, line 36: Could you have used the `ternary` function here like you did elsewhere?

charts/k8ssandra/templates/reaper/reaper.yaml
```diff
@@ -32,4 +32,22 @@ spec:
...
+    securityContext: {{ toYaml .Values.reaper.schemaInitContainerConfig.securityContext | nindent 6 -}}
+    {{- else }}
+    securityContext: {{ "readOnlyRootFilesystem: true" | nindent 6 }}
+    {{- end }}
+
+  {{- if .Values.reaper.podSecurityContext }}
```
jsanda, line 42: Could you have used the `ternary` function here like you did elsewhere?

charts/k8ssandra/templates/reaper/reaper.yaml
```diff
@@ -32,4 +32,22 @@ spec:
...
+  podSecurityContext: {{ toYaml .Values.reaper.podSecurityContext | nindent 4 -}}
+  {{- else }}
+  podSecurityContext: {}
+  {{- end }}
+
+  {{- if .Values.reaper.securityContext }}
```
jsanda, line 48: Could you have used the `ternary` function here like you did elsewhere?

jsanda (commented):
We will need a docs update for this (cc @johnsmartco)

@jeffbanks after this is merged it might be nice to either to a short blog post or forum post about this!

charts/k8ssandra/templates/reaper/reaper.yaml
jeffbanks: Yes, we can,  it just included an extra bit of checking for the existence of the config entry itself and looked a bit messy to stuff all of that into a one-liner.

charts/k8ssandra/templates/reaper/reaper.yaml
jeffbanks: I will take a look for the pod security context, as this may be better as a ternary operation.

charts/k8ssandra/templates/reaper/reaper.yaml
jeffbanks: I will take a look for the pod security context, as this may be better as a ternary operation.

jeffbanks: Added ternary in latest push.

charts/k8ssandra/templates/reaper/reaper.yaml
jeffbanks: Added ternary in latest push.

scripts/clean-ns.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
jeffbanks, line 1: My helper to cleanup some stale state during local testing.

jeffbanks:
@jsanda just the two known integration tests are failing as expected.  Updated the template usage in `reaper.yaml` to reduce some of that template cruft.

jeffbanks:
> @jeffbanks after this is merged it might be nice to either to a short blog post or forum post about this!

@jsanda
I'll start pulling some details.  I've created a [ticket](https://github.com/k8ssandra/k8ssandra/issues/1049) that I will use to track that effort as a separate PR.

jsanda (approved):
@jeffbanks thanks for knocking this out!

PR #1009, opened 2021-07-28
Upgrade Stargate dependency 1.0.29 -> 1.0.31

adejanovski:


burmanm:
@adejanovski This keeps spamming the same 1.0.29 -> 1.0.31 update all the time.

adejanovski:
Anything preventing us from merging the PR?
If you close the PR it should stop proposing the upgrade. The perk of having the PR pushed again each day is that it will replay the CI with the up to date code.

Issue #1020, opened 2021-08-09
Integration test timeouts waiting for dc


jeffbanks:
**Issue**
CassandraDatacenters never reach a valid state in two integration tests.

**Describe the bug**
During execution of integration tests for k8ssandra, two of the tests exhibit the following timeout behavior.

**integration-tests (3.11.10, TestRestoreAfterUpgrade)**
```
2021/08/09 18:28:35 Checking cassandradatacenter dc1 state in namespace k8ssandra2021080918085907...
    integration_steps.go:276: 
        Timed out after 900.000s.
        Expected
            <bool>: false
        to be true
2021/08/09 18:29:05 Installing and starting k8ssandra took 15m16.053640166s
2021/08/09 18:29:05 Not cleaning up cluster as requested
```

**integration-tests (3.11.10, TestUpgradeScenario)**
```
2021/08/09 18:35:46 Checking cassandradatacenter dc1 state in namespace k8ssandra2021080918175707...
    integration_steps.go:276: 
        Timed out after 900.000s.
        Expected
            <bool>: false
        to be true
2021/08/09 18:36:16 Installing and starting k8ssandra took 15m15.774429268s
2021/08/09 18:36:16 Not cleaning up cluster as requested
--- FAIL: TestUpgradeScenario (1686.72s)
    --- PASS: TestUpgradeScenario/Upgrade_from_v1.0.0 (508.25s)
    --- FAIL: TestUpgradeScenario/Upgrade_from_latest (1098.33s)
FAIL
FAIL	github.com/k8ssandra/k8ssandra/tests/integration	1687.209s
FAIL
Error: make: *** [Makefile:43: integ-test] Error 1
Error: Process completed with exit code 2.
```

**To Reproduce**
Push branch PR up to commit: dc193c0799f473f57b26f47745c739881c9d2c59

Example:
[Integration Test Run 1](https://github.com/k8ssandra/k8ssandra/runs/3283363266?check_suite_focus=true)
[Integration Test Run 2](https://github.com/k8ssandra/k8ssandra/pull/1013)

**Expected behavior**
Tests pass.

**Environment (please complete the following information):**
Local integration testing runs and GitHub environment.

**Helm charts version info**
```
k8ssandra       default         1               2021-08-09 14:45:33.6203166 -0500 CDT   deployed        k8ssandra-1.4.0-SNAPSHOT
[...]

PR #1031, opened 2021-08-17
do not change num_tokens when upgrading

### LINKED ISSUE #1029: Upgrading from Cassandra 3.11 to 4.0 fails if num_tokens is not set
## Bug Report

**Describe the bug**
The default value for `num_tokens` in 3.11 is 256. The default in 4.0 is 16. We have a helper template that sets the default based on the Cassandra version. In k8ssandra 1.3.0 we made the default Cassandra version 4.0. Cassandra does not allow you to change the value of `num_tokens` in an existing datacenter.

**To Reproduce**

1. Install k8ssandra 1.2.0 using default values.
1. Upgrade to 1.3.0
1. The following exception will be reported in Cassandra's logs:

```
ERROR [main] 2021-08-16 21:56:18,179 CassandraDaemon.java:909 - Fatal configuration error
org.apache.cassandra.exceptions.ConfigurationException: Cannot change the number of tokens from 256 to 16
	at org.apache.cassandra.service.StorageService.joinTokenRing(StorageService.java:1068)
	at org.apache.cassandra.service.StorageService.joinTokenRing(StorageService.java:1017)
	at org.apache.cassandra.service.StorageService.initServer(StorageService.java:799)
	at org.apache.cassandra.service.StorageService.initServer(StorageService.java:729)
	at org.apache.cassandra.service.CassandraDaemon.setup(CassandraDaemon.java:420)
	at org.apache.cassandra.service.CassandraDaemon.activate(CassandraDaemon.java:763)
	at org.apache.cassandra.service.CassandraDaemon.main(CassandraDaemon.java:887)
```

It causes Cassandra to fail startup.

**Expected behavior**
`num_tokens` should not be changed when upgrading, regardless of Cassandra version.

**Screenshots**

* Helm charts version info
1.3.0

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-805) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-805
┆Priority: Medium

jsanda:
**What this PR does**:
Changes the `k8ssandra.default_num_tokens` template to check if it is an upgrade. If it is an upgrade and `num_tokens` is not set then we make sure that `num_tokens` has the correct default based on the current version of Cassandra. So if the upgrade goes from 3.11.10 to 4.0 for example, then `num_tokens` will remain set to 256.

Fixes #1029, #1020

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/charts/k8ssandra/templates/_helpers.tpl
+++ b/charts/k8ssandra/templates/_helpers.tpl
@@ -197,7 +197,7 @@ Set default num_tokens based on the server version
   */}}
   {{- if $numTokens }}
     {{- if and $datacenter.num_tokens (ne (int $datacenter.num_tokens) (int $numTokens)) }}
-      {{- fail (printf "num_tokes cannot be changed once the CassandraDatacenter is created. The actual value is %d, but the specified value is %d" (int $datacenter.num_tokens) $numTokens) }}
+      {{- fail (printf "num_tokens cannot be changed once the CassandraDatacenter is created. The actual value is %d, but the specified value is %d" (int $datacenter.num_tokens) $numTokens) }}
     {{- end }}
     {{- nindent 6 (print "num_tokens: " $numTokens) }}
   {{- else }}
--- a/charts/k8ssandra/templates/_helpers.tpl
+++ b/charts/k8ssandra/templates/_helpers.tpl
@@ -163,21 +163,21 @@ Set default num_tokens based on the server version
 {{- $datacenter := (index .Values.cassandra.datacenters 0) -}}
 {{- if .Release.IsInstall }}
   {{- if $datacenter.num_tokens }}
-    num_tokens: {{ $datacenter.num_tokens }}
+    {{- nindent 6 (print "num_tokens: " $datacenter.num_tokens) }}
   {{- else }}
   {{- if hasPrefix "3.11" .Values.cassandra.version }}
-    num_tokens: 256
+    {{- nindent 6 (print "num_tokens: 256") }}
   {{- else }}
-    num_tokens: 16
+    {{- nindent 6 (print "num_tokens: 16") }}
   {{- end }}
 {{- end }}
 {{- else }}
   {{ $datacenterObj := (lookup "cassandra.datastax.com/v1beta1" "CassandraDatacenter" .Release.Namespace $datacenter.name) }}
   {{- if $datacenterObj }}
     {{- if hasPrefix "3.11" $datacenterObj.spec.serverVersion }}
-      num_tokens: 256
+      {{- nindent 6 (print "num_tokens: 256") }}
     {{- else }}
-      num_tokens: 16
+      {{- nindent 6 (print "num_tokens: 16") }}
     {{- end }}
   {{ end }}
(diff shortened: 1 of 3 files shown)


charts/k8ssandra/templates/_helpers.tpl
```diff
@@ -161,15 +161,54 @@ Set default num_tokens based on the server version
...
-    {{- if hasPrefix "3.11" .Values.cassandra.version }}
-      num_tokens: 256
-    {{- else }}
-      num_tokens: 16
-    {{- end }}
+{{- if .Release.IsInstall }}
+  {{- /*
+  If num_tokens is set we simply use that value.
+  */}}
+  {{- if $datacenter.num_tokens }}
```
burmanm, line 168: Won't this fail if user didn't set it originally? Yes, it's in the created CassandraDatacenter, but it's not in the Helm's internal properties.

charts/k8ssandra/templates/_helpers.tpl
```diff
@@ -161,15 +161,54 @@ Set default num_tokens based on the server version
...
+    {{- nindent 6 (print "num_tokens: 256") }}
+  {{- else }}
+    {{- nindent 6 (print "num_tokens: 16") }}
+  {{- end }}
+{{- end }}
+{{- else }}
+  {{- /*
+  If upgrading and num_tokens is set simply use that value.
+  */}}
+  {{- if $datacenter.num_tokens }}
```
jdonenine, line 184: So, to avoid the problem where the values are trying to change the num_tokens and they shouldn't, should we simply just not ever use the values on upgrade?  Always use what we find active in the currently deployed datacenter?

jdonenine (approved):
Just the one comment about how to best handle the case where the values could be setting something incorrectly, things look good.

I tested a number of scenarios locally and everything resulted as expected.

charts/k8ssandra/templates/_helpers.tpl
jsanda: Yeah, I think that sounds right to use whatever is set in the existing CassandraDatacenter, with a small exception. It is possible that we do not find an existing CassandraDatacenter in which case we should apply the value specified for `num_tokens` or apply the default log if the property is not set. I will update the PR accordingly.

jdonenine:
@jsanda Do you want to go ahead with this PR now that we know the IT failures are no longer related to it?

jsanda:
Let me check the logs from the latest failures since I pushed more changes yesterday. If those look ok, then I am good with merging. I will report back shortly.

jsanda:
With the logs in https://github.com/k8ssandra/k8ssandra/pull/1031/checks?check_run_id=3365825424 I am seeing `num_tokens: 16`. I am not sure why though. I have tested the same scenario locally and everything looks good. That scenario being:

* Install k8ssandra 1.0 with Cassandra 3.11.10 with default for `num_tokens`
* Upgrade to latest k8ssandra in my branch but staying on Cassandra 3.11.10.
* `num_tokens` does not change

Until I figure out what is happening, I think we need to hold off on merging.

PR #1052, opened 2021-08-20
Medusa and reaper operator securityContext

jeffbanks:
**What this PR does**:
Security context for medusa and reaper operators within K8ssandra.

Covers security requirements as described in:
* [k8ssand-697](https://k8ssandra.atlassian.net/browse/K8SSAND-697)
* [k8ssand-696](https://k8ssandra.atlassian.net/browse/K8SSAND-696)

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/charts/medusa-operator/templates/deployment.yaml
+++ b/charts/medusa-operator/templates/deployment.yaml
@@ -1,3 +1,5 @@
+{{- $defaultPodSecurityCtx :=  (include "k8ssandra-common.pod.pod-security-ctx-defaults" .) -}}
+{{- $defaultSecurityCtx := (include "k8ssandra-common.container.security-ctx-defaults" .) -}}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -12,6 +14,7 @@ spec:
     metadata:
       labels: {{- include "k8ssandra-common.labels" . | indent 8 }}
     spec:
+      securityContext: {{- ternary ($defaultPodSecurityCtx) (.Values.podSecurityContext | toYaml) (empty .Values.podSecurityContext) | nindent 8 }}
       serviceAccountName:  {{ include "k8ssandra-common.serviceAccountName" . }}
       containers:
       - args:
@@ -27,4 +30,5 @@ spec:
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         name: medusa-operator
         resources: {{- toYaml .Values.resources | nindent 10 }}
+        securityContext: {{- ternary ($defaultSecurityCtx) (.Values.securityContext | toYaml) (empty .Values.securityContext) | nindent 12 }}
       terminationGracePeriodSeconds: 10
--- a/charts/reaper-operator/templates/deployment.yaml
+++ b/charts/reaper-operator/templates/deployment.yaml
@@ -1,18 +1,21 @@
+{{- $defaultPodSecurityCtx :=  (include "k8ssandra-common.pod.pod-security-ctx-defaults" .) -}}
+{{- $defaultSecurityCtx := (include "k8ssandra-common.container.security-ctx-defaults" .) -}}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: {{ include "k8ssandra-common.fullname" . }}
-  labels: {{ include "k8ssandra-common.labels" . | indent 4 }}
+  labels: {{ include "k8ssandra-common.labels" . | nindent 4 }}
 spec:
   replicas: 1
   selector:
-    matchLabels:
-      {{- include "k8ssandra-common.selectorLabels" . | nindent 6 }}
+    matchLabels: {{- include "k8ssandra-common.selectorLabels" . | nindent 6 }}
   template:
     metadata:
       labels: {{- include "k8ssandra-common.labels" . | indent 8 }}
     spec:
-      serviceAccountName:  {{ include "k8ssandra-common.serviceAccountName" . }}
+      serviceAccountName: {{ include "k8ssandra-common.serviceAccountName" . }}
+      securityContext:
+        {{- ternary ($defaultPodSecurityCtx) (.Values.podSecurityContext | toYaml) (empty .Values.podSecurityContext) | nindent 8 }}
       containers:
       - args:
         - --enable-leader-election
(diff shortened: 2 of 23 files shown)


charts/medusa-operator/templates/deployment.yaml
```diff
@@ -1,3 +1,5 @@
+{{- $defaultPodSecurityCtx := "{}" }}
```
jsanda, line 1: `podSecurityContext` is already defined in medusa-operator/values.yaml. Why redefine it here?

charts/medusa-operator/templates/deployment.yaml
```diff
@@ -12,6 +14,7 @@ spec:
     metadata:
       labels: {{- include "k8ssandra-common.labels" . | indent 8 }}
     spec:
+      securityContext: {{- ternary ($defaultPodSecurityCtx) (.Values.podSecurityContext | toYaml) (empty .Values.podSecurityContext) | nindent 8 }}
```
jsanda, line 17: Could this be simplified to something like this:

```
      securityContext:
        {{- toYaml .Values.podSecurityContext | nindent 8 }}
```

charts/medusa-operator/templates/deployment.yaml
```diff
@@ -1,3 +1,5 @@
+{{- $defaultPodSecurityCtx := "{}" }}
+{{- $defaultSecurityCtx := "readOnlyRootFilesystem: true" }}
```
jsanda, line 2: Isn't this already the default in values.yaml? If so why redeclare it here?

charts/medusa-operator/templates/deployment.yaml
```diff
@@ -27,4 +30,5 @@ spec:
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         name: medusa-operator
         resources: {{- toYaml .Values.resources | nindent 10 }}
+        securityContext: {{- ternary ($defaultSecurityCtx) (.Values.securityContext | toYaml) (empty .Values.securityContext) | nindent 12 }}
```
jsanda, line 33: Could this be simplified to something like this:

```
          securityContext:
            {{- toYaml .Values.securityContext | nindent 12 }}
```

charts/reaper-operator/templates/deployment.yaml
```diff
@@ -1,18 +1,21 @@
+{{- $defaultPodSecurityCtx := "{}" }}
+{{- $defaultSecurityCtx := "readOnlyRootFilesystem: true" }}
```
jsanda, line 2: Similar to my comments for the medusa-operator chart, why redefine these here when they are already defined in values.yaml?

charts/reaper-operator/templates/deployment.yaml
```diff
@@ -1,18 +1,21 @@
...
-      {{- include "k8ssandra-common.selectorLabels" . | nindent 6 }}
+    matchLabels: {{- include "k8ssandra-common.selectorLabels" . | nindent 6 }}
   template:
     metadata:
       labels: {{- include "k8ssandra-common.labels" . | indent 8 }}
     spec:
-      serviceAccountName:  {{ include "k8ssandra-common.serviceAccountName" . }}
+      serviceAccountName: {{ include "k8ssandra-common.serviceAccountName" . }}
+      securityContext:
+        {{- ternary ($defaultPodSecurityCtx) (.Values.podSecurityContext | toYaml) (empty .Values.podSecurityContext) | nindent 8 }}
```
jsanda, line 18: Could this be simplified to something like this:

```
      securityContext:
        {{- toYaml .Values.podSecurityContext | nindent 8 }}
```

charts/reaper-operator/templates/deployment.yaml
```diff
@@ -27,4 +30,6 @@ spec:
         name: reaper-operator
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         resources: {{- toYaml .Values.resources | nindent 10 }}
+        securityContext:
+          {{- ternary ($defaultSecurityCtx) (.Values.securityContext | toYaml) (empty .Values.securityContext) | nindent 10 }}
```
jsanda, line 34: Could this be simplified to something like this:

```
          securityContext:
            {{- toYaml .Values.securityContext | nindent 12 }}
```

jeffbanks: I can maybe see a simplified approach for the `pod securityContext` as it has fewer impact points throughout the `deployment` and `cassdc` for value defaults.  But there was a reason I was making the pod and container templates the same.

The reason for the template ternary empty checks on `securityContext` is that the default would need to reside in all locations of the values.yaml.   If that were to contain more properties (which we are still figuring out), this would become messy in the values file IMO.  

That complexity would apply to both pod and container securityContext.  Today, we think the default is {} for the pod securityContext, but if this were to expand to additional properties, this becomes a nice thing to handle in the helpers in one place.  Also, it would maintain consistency with the container securityContext defaults (having as {}  in the values yaml and the helper translates that to a default w/ 0..N properties).

_Examples:_

// Today's container default
```yaml
securityContext: 
  readOnlyRootFilesystem: true
  ...
```

// As an example of pod defaults other than {}
```yaml
securityContext: 
  runAsUser: 1000
  runAsGroup: 3000
  fsGroup: 2000
  ...
```

One observation that you mentioned is that we can eliminate the value defaults in the operator's `deployment` and I would need to define this in a `_helpers.tpl` in the `k8ssandra-commons` as `deployment` doesn't have access to `_helpers.tpl` in k8ssandra templates.  

That would lead me to suggest we also migrate the existing k8ssandra (used in cassdc.yaml) defaults for securityContext to the commons template helper.  We would then have one place to adjust securityContext defaults (which may grow in length) for containers and pods.  wdyt?

jeffbanks:
@jsanda I added the templates to commons to see if this makes more sense as described in my previous comment.

charts/reaper-operator/templates/deployment.yaml
jsanda: 👍  to adding the templates in `_helpers.tpl`.

One other thing to keep in mind. The properties available in a SecurityContext and PodSecurityContext are slightly different, so you will probably want to use separate templates them.

jeffbanks: 💯 That is correct, we will have a default common template for pod and container security contexts.

I've pushed the common template updates, but looks to be breaking some integration tests, so once that is resolved, I'll have you verify this once more.  Thanks!

jsanda (approved):
@jeffbanks nice work!

jeffbanks:
Initial set of security context additions to medusa and reaper operators in support of overall security hardening and configuration.

PR #1062, opened 2021-08-24
Reaper config volume & mountPath

jeffbanks:
**What this PR does**:
Provides a reaper config volume & empty mount path that can be targeted for `securityContext` permissions.

Support of:
* [k8ssand-700](https://k8ssandra.atlassian.net/browse/K8SSAND-700)
* [k8ssand-701](https://k8ssandra.atlassian.net/browse/K8SSAND-701) 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/charts/k8ssandra/templates/cassandra/cassdc.yaml
+++ b/charts/k8ssandra/templates/cassandra/cassdc.yaml
@@ -179,6 +179,10 @@ spec:
             mountPath: /opt/metrics-collector/config/
           - name: cassandra-tmp
             mountPath: /tmp/
+          {{- if .Values.reaper.enabled }}
+          - name: reaper-config
+            mountPath: /etc/reaper/
+          {{- end }}
       - name: server-config-init
         securityContext: {{- ternary ($defaultSecurityCtx) (.Values.cassandra.configBuilder.securityContext | toYaml) (empty .Values.cassandra.configBuilder.securityContext) | nindent 10 }}
      {{- if (or .Values.cassandra.auth.enabled .Values.reaper.enabled) }}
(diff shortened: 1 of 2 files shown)


charts/k8ssandra/templates/cassandra/cassdc.yaml
```diff
@@ -179,6 +179,10 @@ spec:
             mountPath: /opt/metrics-collector/config/
           - name: cassandra-tmp
             mountPath: /tmp/
+          {{- if .Values.reaper.enabled }}
```
Miles-Garnsey, line 182: I know that this is going to be a dumb question, but just out of curiosity, why are we putting a reaper-config in the CassandraDC resource?

I get the bit about wanting a read-only FS in the container for securityContext purposes, that makes sense.

jeffbanks: @Miles-Garnsey Not a dumb question at all.  I was working with the medusa vol configs locally, and it made sense to me at the time that we create the reaper vols and assign mountPath in this location of cassdc as part of the init container.  In this case, the /etc/cassandra-reaper.yml will be replaced with /etc/reaper/cassandra-reaper.yml by the reaper image setup and we are wanting /etc/{} to be in-place.

Let's phone a friend (@jsanda) to get another opinion.  I can see where decoupling this setup may make more sense.  

Thanks for raising the question!

jsanda: Medusa runs in the Cassandra pod which is why we need the Medusa-related volumes declared in the CassandraDatacenter. Reaper however runs in its own pod; therefore, I would assume we only need the volume in the Reaper pod. 

To support this we will need a change in reaper-operator to add the volume and mount. This gets into the debate of whether we only expose the config through the Reaper CRD, allow the user to provide his own Reaper config, or both. I don't know its worth the time and effort to try and bite all of this off right now. I would just go the minimum changes necessary.

PR #1095, opened 2021-09-09
Temp relax read-only for containers

### LINKED ISSUE #1094: Cassandra won't start when deployed on GKE with 1.4.0-SNAPSHOT
## Bug Report

**Describe the bug**
Cassandra fails to start with the current `main` as the process cannot create directories for data nor commit logs when deploying on GKE:

```
ERROR [main] 2021-09-09 16:21:28,902 CassandraDaemon.java:909 - Exception encountered during startup
java.lang.ExceptionInInitializerError: null
    at org.apache.cassandra.service.CassandraDaemon.setup(CassandraDaemon.java:260)
    at org.apache.cassandra.service.CassandraDaemon.activate(CassandraDaemon.java:763)
    at org.apache.cassandra.service.CassandraDaemon.main(CassandraDaemon.java:887)
Caused by: java.lang.IllegalStateException: Failed to mkdirs /opt/cassandra/data/data; unable to start server
    at org.apache.cassandra.config.DatabaseDescriptor.createAllDirectories(DatabaseDescriptor.java:1422)
    at org.apache.cassandra.db.commitlog.CommitLog.<init>(CommitLog.java:94)
    at org.apache.cassandra.db.commitlog.CommitLog.construct(CommitLog.java:77)
    at org.apache.cassandra.db.commitlog.CommitLog.<clinit>(CommitLog.java:64)
    ... 3 common frames omitted
```

Looking at the directories and symlinks, it looks like `/var/lib/cassandra` isn't mounted so that the `cassandra`  user can write to it:

Capture d’écran 2021-09-09 à 18 42 48
Capture d’écran 2021-09-09 à 18 43 44

**To Reproduce**
Deploy K8ssandra 1.4.0-SNAPSHOT on GKE using the following Helm values file:

```
cassandra:
  version: "4.0.0"
  heap:
   size: 4g
  gc:
    g1:
      enabled: true
      setUpdatingPauseTimePercent: 5
      maxGcPauseMillis: 300
  resources:
    requests:
      memory: "8Gi"
      cpu: "2000m"

  datacenters:
  - name: dc1
    size: 3
  ingress:
    enabled: false

  cassandraLibDirVolume:
    storageClass: standard-rwo
    size: 300Gi
```

**Expected behavior**
Cassandra should start and data volumes should mount so that required directories can be created by the `cassandra` user.
In K8ssandra 1.3.1, the rights for the `/var/lib/cassandra` are set differently, allowing Cassandra to start properly:
Capture d’écran 2021-09-09 à 19 03 32

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-875) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-875
┆Priority: Medium

jeffbanks:
**What this PR does**:
Temporarily relax the default `securityContext` for containers to resolve issue reported with c* image failing `mkdir` as root in GKE.

Fixes #1094 
[k8ssand-875](https://k8ssandra.atlassian.net/browse/K8SSAND-875)

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/charts/cass-operator/values.yaml
+++ b/charts/cass-operator/values.yaml
@@ -58,15 +58,12 @@ podSecurityContext: {}
 securityContext:
   # -- Mark root filesystem as read only
   readOnlyRootFilesystem: false
-
   # -- Run cass-operator container as non-root user
-  # runAsNonRoot: true
-
+  runAsNonRoot: true
   # -- Group for the user running the cass-operator container / process
-  # runAsGroup: 65534
-
+  runAsGroup: 65534
   # -- User for running the cass-operator container / process
-  # runAsUser: 65534
+  runAsUser: 65534
 
 # -- Resources requests and limits for the cass-operator pod. We usually
 # recommend not to specify default resources and to leave this as a conscious
--- a/charts/k8ssandra-common/templates/_helpers.tpl
+++ b/charts/k8ssandra-common/templates/_helpers.tpl
@@ -29,7 +29,7 @@ SecurityContext helpers for pod and container scope.
 k8ssandra container securityContext defaults
 **/}}
 {{- define "k8ssandra-common.container.security-ctx-defaults" -}}
-readOnlyRootFilesystem: true
+readOnlyRootFilesystem: false
 {{- end }}
 
 {{/**
--- a/charts/k8ssandra/templates/cassandra/cassdc.yaml
+++ b/charts/k8ssandra/templates/cassandra/cassdc.yaml
@@ -166,7 +166,6 @@ spec:
 {{- end }}
   podTemplateSpec:
     spec:
-      securityContext: {{- ternary ($defaultPodSecurityCtx) (.Values.cassandra.podSecurityContext | toYaml) (empty .Values.cassandra.podSecurityContext) | nindent 10 }}
       initContainers:
       - name: base-config-init
         image: {{ include "k8ssandra.cassandraImage" . }}
--- a/tests/unit/template_cassdc_test.go
+++ b/tests/unit/template_cassdc_test.go
@@ -159,7 +159,9 @@ var _ = Describe("Verify CassandraDatacenter template", func() {
 			AssertContainerSecurityContextExistsAndMatches(cassdc, CassandraContainer, expectedCtx)
 
 			// Default pod security context for cassdc
-			Expect(cassdc.Spec.PodTemplateSpec.Spec.SecurityContext).ToNot(BeNil())
+			// TODO - revisit as this was potentially causing issues by defaulting to {}
+			// with respect to running in GKE - see k8ssandra #1094
+			Expect(cassdc.Spec.PodTemplateSpec.Spec.SecurityContext).To(BeNil())
 		})
 
 		It("is not rendered if disabled", func() {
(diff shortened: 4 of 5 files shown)


jeffbanks:
@jsanda 
Perhaps I was a bit too optimistic to restrict ALL containers to have `readOnlyRootFilesystem: true`. 😄   The unit and integration tests seemed to be o.k. originally, but when Alex ran with GKE, he may have located a case we don't have covered properly with `mkdir` commands being issued.  C* user trying to perform writes to where root had ownership.

@jsanda you mentioned yesterday an issue you were investigating with the cass-operator, "_cass-operator code that configures the PodSecurityContext for the StatefulSet_".   Would that be related in anyway to this perhaps?  Throwing it out incase this patch doesn't resolve Alex's issue he found. 

I found some logic [here](https://github.com/k8ssandra/cass-operator/blob/6c64f9c4f5d1d86d7bb979dbf5b14194d1bf5032/pkg/reconciliation/construct_statefulset_test.go) and [here](https://github.com/k8ssandra/cass-operator/blob/6c64f9c4f5d1d86d7bb979dbf5b14194d1bf5032/pkg/reconciliation/construct_statefulset.go) that could be related.

For now, in `k8ssandra/k8ssandra`, I've relaxed the default `securityContext` setting at the container level to see if this allows c* to proceed with its activities.  We have tickets for c*, mgmt-api, reaper, and medusa image modifications to ensure a non-root user is running, so if this is the root cause (no pun intended) we will need to address `root` ownership of those dirs in those tickets.   I'll need to come back here to re-enable read-only for root filesystem. A bit of a chicken and egg problem for now IMO. 🐔 

@adejanovski  since you have your `GKE` test environment setup can you take another run at it with the latest to be sure this is the culprit?  If not the root cause, we will need to dig into the other additions that were applied after 1.3.x release to k8ssandra/k8ssandra and/or the cass-operator to see there is something restricting the mkdir operation.

adejanovski:
@jeffbanks, I'm still seeing the same issue with this change. The rights seem unaffected by the `readOnlyRootFilesystem` setting change:
Capture d’écran 2021-09-10 à 10 06 34

I can confirm the change was applied by checking the pods definition:
Capture d’écran 2021-09-10 à 10 09 20

jeffbanks:
@adejanovski can you send the full pod definition from k8ssandra?  It might be good to see what the `cass-operator` has defined as well or any logs, etc.  We didn't hear back from @jsanda on the question I had about his investigation with the cass-operator this week.  I referenced two areas of code in this ticket that could be impacting what you are seeing. 

You reported 1.3.x is functioning as expected, are there any changes in any of the other related components from 1.3.x to 1.4-snapshot other than just k8ssandra core?

jeffbanks:
Summary of changes:

1) The cass-operator settings for its securityContext were re-enabled as those were found to be commented out from a previous PR.  I had done this while testing and it slipped throught.

2) @jsanda determined that the PodSecurityContext: { } in the cassdc, when removed resolved the issue @adejanovski was seeing in the GKE environment.  As such, the pod level securityContext has been removed for now as this was only a placeholder at this point.  We will need to determine why this was causing unexpected behaviour.

adejanovski (approved):
This fixes the deployment issues. Thanks!

PR #1104, opened 2021-09-16
Update Management API v0.1.28 -> v0.1.29

### LINKED ISSUE #1083: Add support for deployment of C* 4.0.1
N/A

The 4.0.1 release of C* is currently up for a vote, once available we'd like to see support for deploying it via K8ssandra made available.

N/A

**Additional context**
A list of changes found in 4.0.1 can be found [here](https://gitbox.apache.org/repos/asf?p=cassandra.git;a=blob_plain;f=CHANGES.txt;hb=refs/tags/4.0.1-tentative).

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-865) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-865
┆Priority: Medium

emerkle826:
**What this PR does**:
Updates Management API image to v0.1.29

Fixes #1083 

**Checklist**
- [x] Changes manually tested


adejanovski (approved):

charts/k8ssandra/values.yaml
```diff
@@ -8,19 +8,21 @@ cassandra:
   #    - 3.11.10
   #    - 3.11.11
   #    - 4.0.0
+  #    - 4.0.1
   version: "4.0.0"
```
jdonenine, line 12: Should we bump the default to 4.0.1 as well?

I'd say this would be an otherwise contentious discussion, but it's just a patch release, so perhaps makes sense?  What does everyone @k8ssandra/maintainers think?

burmanm: If it's a bugfix release, why wouldn't we bump?

adejanovski: Agreed, we should bump. Good practice being that peeps set it explicitly in their deployments, which allows them to control when an upgrade should happen.

jdonenine: I agree, we should 👍

jdonenine (approved):

jdonenine (changes requested):
Let's make the change to go to 4.0.1 as the default as discussed, thanks!

charts/k8ssandra/values.yaml
emerkle826: so, they only catch with updating it here is potentially the integration tests. Same issue when we added 3.11.11. The upgrade tests fail because "4.0.1" isn't a valid Cassandra version in older K8ssandra chart versions.

I can update the default here and maybe leave the integration test matrix at 4.0.0 to see how the tests go, but we will have to address the tests at some point.

emerkle826:
@jdonenine Updated to 4.0.1. If changes look good, I can squash and merge.

PR #1118, opened 2021-10-02
Upgrade Stargate dependency 1.0.29 -> 1.0.35

adejanovski:


jeffbanks (approved):
LGTM

PR #1126, opened 2021-10-11
Expanding security ctx controls

jeffbanks:
**What this PR does**:
Adds unique defaults, and security context controls for the various k8ssandra components.

Provides the ability to specify unique container securityContext defaults for reaper vs. medusa vs. cassandra components.

Complements effort: [k8ssand-705 epic](https://k8ssandra.atlassian.net/browse/K8SSAND-705) stories.

**Checklist**
- [x] Changes manually tested


charts/k8ssandra/templates/cassandra/cassdc.yaml
```diff
@@ -1,7 +1,13 @@
 {{- $datacenter := (index .Values.cassandra.datacenters 0) -}}
 {{- $medusaImage := (printf "%s:%s" .Values.medusa.image.repository .Values.medusa.image.tag) -}}
-{{- $defaultSecurityCtx := (include "k8ssandra-common.container.security-ctx-defaults" .) -}}
-{{- $defaultPodSecurityCtx := (include "k8ssandra-common.pod.pod-security-ctx-defaults" .) -}}
+{{- $cassServerCfgSecurityCtx := (include "k8ssandra-common.container.cass-svrcfg-security-ctx-defaults" .) -}}
```
jsanda, line 3: It looks like there is a separate template used for each container. What is the value in using the templates vs just making sure the security context is configurable for each container and providing a default value? A template would make more sense to me if it was shared and reused, but that doesn't appear to be the case.

charts/k8ssandra/templates/cassandra/cassdc.yaml
```diff
@@ -166,11 +172,12 @@ spec:
 {{- end }}
   podTemplateSpec:
     spec:
+      securityContext: {{- ternary ($cassPodSecurityCtx) (.Values.cassandra.podSecurityContext | toYaml) (empty .Values.cassandra.podSecurityContext) | nindent 8 }}
```
jsanda, line 175: Following from my comment above about just setting defaults vs templates, I think this could then be simplified to 

```yaml
securityContext: {{- .Values.cassandra.podSecurityContext | toYaml | nindent 8 }}
```
And when the changes from https://github.com/k8ssandra/cass-operator/pull/218 are available we won't even need to set the `securityContext`.

jsanda (commented):
Maybe I am missing something, but it looks like this could be simplified considerably by setting default values for the `securityContext` and `podSecurityContext` properties in the various values.yaml files.

charts/k8ssandra/templates/cassandra/cassdc.yaml
jeffbanks: Default security settings between containers are currently setup to be used as a 1:1 mapping for the k8ssandra common templates.  The `SecurityContext` default settings applied by K8ssandra _may_ have cases in the future where the templates are shared among container settings.  In this PR they are independent of one another to allow for flexibility of change as we evolve the overall set of default security properties. 

Having these scoped to owned containers (and pods) made it easier to adjust a security policy without impacting other components.  Reusing across unrelated components became time consuming to test and track how a new setting impacted the others.  Being a security related feature and its importance, redundancy & clarity seemed more appropriate for these.

One convenient side-effect of the template usage is when a user has `SecurityContext` properties existing in the `values.yaml`, it becomes an easy way to identify that the K8ssandra default security policy is being overridden for a particular container and/or pod.  

Also, not having defaults existing directly in the `values.yaml` is less verbose prior to template rendering and reduces the chance for misconfigured applied security settings.  Our intent I believe is that it will be rare when needing to override the K8ssandra security defaults as they adhere to k8s best-practices.

charts/k8ssandra/templates/cassandra/cassdc.yaml
jeffbanks: See comment above for more explanation on the use of templates.  I'm not clear on why securityContext would no longer need to be set.  Are you saying for the pod in this specific context, or for all of the medusa, reaper, init containers, etc. in the cassdc  spec overall?   The entire scope of these changes are securityContext k8s related.

jeffbanks: Moving forward with changes to remove from templates and apply to directly to the `values.yaml` for the configuration by users.

charts/k8ssandra/templates/cassandra/cassdc.yaml
jeffbanks: Moving forward with changes to remove from templates and apply to directly to the `values.yaml` for the configuration by users.

jeffbanks:
@jsanda Recent changes are scoped to operator `securityContext` along with placeholders for expansion as we evolve phase II changes.

* SecurityContext default placeholders existing for all containers and identified pods.
* Updated operator securityContext settings w/ restrictions for `non-root` and `readOnlyRootFilesystem`.
```
runAsNonRoot: true
runAsUser: 65534
runAsGroup: 65534
readOnlyRootFilesystem: true
```

* Removed securityContext templates and we now directly apply what is in the `values.yaml` for any defaults.
* Tested with Medusa, Reaper and Reaper-operator instances using this PR branch of K8ssandra.

Related `reaper` and `reaper-operator` PRs as reference:
* [Reaper PR](https://github.com/thelastpickle/cassandra-reaper/pull/1139) - closed
* [Reaper-operator PR](https://github.com/k8ssandra/reaper-operator/pull/84) - open, ready for review

jsanda (commented):
When I run `helm install` I get this error:

`Error: Reaper.reaper.cassandra-reaper.io "test-reaper" is invalid: spec.schemaInitContainerConfig: Invalid value: "null": spec.schemaInitContainerConfig in body must be of type object: "null"`

Here is the relevant part when I ran `helm template`

```yaml
spec:
  image: docker.io/thelastpickle/cassandra-reaper:3.0.0
  ServiceAccountName: default
  serverConfig:
    autoScheduling:
      enabled: false
    storageType: cassandra
    jmxUserSecretName: test-reaper-jmx
    cassandraBackend:
      cassandraUserSecretName: test-reaper
      cassandraDatacenter:
        name: dc1
      replication:
        networkTopologyStrategy:
          dc1: 1
  schemaInitContainerConfig:
  configInitContainerConfig:
  podSecurityContext:
    {}
  securityContext:
    {}
```

jsanda:
@jeffbanks I tried again with a new kind cluster and it installed without error. I think your changes are good to go.

jsanda (approved):

PR #1153, opened 2021-10-29
Set preserveUnknownFields to false when upgrading to cass-operator 1.8.0

### LINKED ISSUE #1152: K8SSAND-1427 ⁃ CRD upgrader needs changes for cass-operator 1.8.0 (CRD v1beta1 -> v1)
## Bug Report

**Describe the bug**
When updating the cass-operator from 1.7.1 to 1.8.0, the user needs to run the following command:

```console
kubectl patch crd cassandradatacenters.cassandra.datastax.com -p '{"spec":{"preserveUnknownFields":false}}'
```

Before the update of CRD goes through. This is required to support the Kube 1.22 environment. This isn't done by the CRD upgrader for cass-operator CRDs however, so we need to add this feature. This is required if the user wants to upgrade to newer Kubernetes version (this bug will not occur in kube 1.21 for example).

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1427) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1427
┆priority: Medium

burmanm:
**What this PR does**:
Automates the CRD upgrader's step of setting preserveUnknownFields to false when updating to newest k8ssandra version (cass-operator 1.8.0). This is required, otherwise the CRD upgrade will fail in Kubernetes 1.22

Fixes #1152 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/crds/upgrade.go
+++ b/pkg/crds/upgrade.go
@@ -22,6 +22,7 @@ import (
 
 	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiextv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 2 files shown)


jsanda:
@burmanm do we still want these changes?

burmanm:
Without this, the update from cass-operator 1.8.0 will fail if newer Kubernetes is installed.

jsanda:
For testing would it be sufficient to install on Kubernetes 1.23 with cass-operator 1.8.0 and then do an upgrade?

Interestingly, I don't see a k8ssandra release that uses 1.8.0. k8ssandra 1.3.3 is on cass-operator 1.7.1. k8ssandra 1.4.0 is on cass-operator 1.9.0. Given this would it be better to test upgrading from cass-operator 1.7.1? I ask because I am sure plenty of users install k8ssandra and stick with the defaults for the different components.

burmanm:
Meant to write pre 1.8.0, as in the update from 1.7.1 to 1.9.0 would trigger it also. You can't install 1.7.1 to Kube 1.23, which makes the test painful, you need older Kube version and then update the Kube also.

jsanda:
Thanks for the clarification. I can do a manual upgrade test with an install on k8s 1.21 and upgrade to k8s 1.22.

jsanda:
I talked with @burmanm and test steps will be as follows:

* Install k8s 1.21
* Install k8ssandra 1.3.3
* Upgrade k8ssandra
    * Verify preserveUnknownFields is patched in CRD
* Upgrade k8s to 1.22
* Verify Cassandra is still running and that cass-operator is operational

PR #1175, opened 2021-11-10
Medusa operator 1.22

### LINKED ISSUE #1170: K8SSAND-1039 ⁃ v1beta1 API versions of Ingress are removed in 1.22
In k8s 1.22, extensions/v1beta1 and networking.k8s.io/v1beta1 API versions of Ingress have been removed and replaced with v1 which have some breaking changes in the manifests:

* Migrate manifests and API clients to use the networking.k8s.io/v1 API version, available since v1.19.
* All existing persisted objects are accessible via the new API
* Notable changes:
spec.backend is renamed to spec.defaultBackend
The backend serviceName field is renamed to service.name
Numeric backend servicePort fields are renamed to service.port.number
String backend servicePort fields are renamed to service.port.name
pathType is now required for each specified path. Options are Prefix, Exact, and ImplementationSpecific. To match the undefined v1beta1 behavior, use ImplementationSpecific. 

v1 versions of these APIs were introduced in 1.19, so supporting Ingress in 1.22 would mean we lose support for k8s versions before 1.19, unless we generate different Ingress manifests depending on the target k8s version.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1039) by [Unito](https://www.unito.io)
┆Reviewer: 5e66ca0cae59460cfb01a13c
┆epic: Deployments on k8s v1.22 fail
┆fixVersions: k8ssandra-1.4.0
┆friendlyId: K8SSAND-1039
┆priority: Medium

adejanovski:
**What this PR does**:
Adds k8s 1.22 support to K8ssandra.

This requires to upgrade reaper-operator, medusa-operator and kube-prometheus-stack to the latest versions as well as make the Stargate Ingress template use conditionals based on API capabilities to output either a v1beta1 or a v1 Ingress manifest.

Fixes #1170

**Checklist**
- [x] Changes manually tested


adejanovski:
There are quite a few things happening in this PR:

- medusa-operator was upgraded to 0.4.0 to get 1.22 compatibility (contains upgrades to operator sdk 1.6.1)
- The Stargate Ingress template is now checking the capabilities to see if networking v1 is available or not, and renders the Ingress manifest depending on that
- kube-prometheus-stack was upgraded to the latest to support v1 ingress as well, but Grafana was pinned to v7 in order to avoid using an AGPL licensed version. This required some changes in the integration tests as labels were changed since our version
- The kind cluster created by the integration test script is using 1.22.2 by default
- The default version for integration tests is still 1.21, but 1.22 was added to the matrix
- Upgrade scenarios are skipped with 1.22 since we have no prior version of K8ssandra that would support 1.22
- The way the kubernetes version is determined in CI was reworked to be more consistent and correctly usable in the matrix (we can get it from a workflow dispatch trigger when ran manually and through the k8s matrix workflow, or use a default version if the workflow is triggered by a push/PR).

.github/workflows/integration-tests-release.yaml
```diff
@@ -25,13 +25,30 @@ on:
       kubernetes_version:
         description: 'Kubernetes version'     
         required: true
-        default: 'v1.20.2'
+        default: 'v1.21.1'
```
jsanda, line 28: The version in `tests/integration/scripts/create_kind_cluster.sh` is set to 1.22.2. I think we should either make the default in the workflow 1.22.2 or change the version in the script to 1.21.1.

adejanovski: I'll change the version in the script then, because changing the default in the workflow would mean not executing the upgrade tests on PRs.

jsanda (approved):

PR #1180, opened 2021-11-12
Custom cassandra.yaml for Cassandra and Stargate [K8SSAND-793][K8SSAND-1036][K8SSAND-1038][K8SSAND-1044]

jsanda:
**What this PR does**:
Adds a `cassandraYamlConfigMap` property for Cassandra and Stargate. The property specifies a ConfigMap that contains a custom cassandra.yaml. The custom cassandra.yaml is merged with the base config.

As noted in the comments in values.yaml, there are a number of properties that should not be specified in the custom cassandra.yaml. Some properties, like those that require setting up volumes and mounts need to be handled separately. Others are generated elsehwere in the the cassdc.yaml template.

**Which issue(s) this PR fixes**:
Fixes #1023, #1166, #1169, #1177

**Checklist**
- [ ] Changes manually tested
- [ ] Automated Tests added/updated
- [ ] Documentation added/updated
- [ ] CHANGELOG.md updated (not required for documentation PRs)
- [ ] CLA Signed:  [DataStax CLA](https://cla.datastax.com/)

PR #1198, opened 2021-11-19
don't declare an empty volumeMounts

jsanda:
**What this PR does**:
This is a follow up to #1180. An empty volumeMounts fails on k8s versions earlier than 1.21 with this error:

`Error: CassandraDatacenter.cassandra.datastax.com "dc1" is invalid: spec.podTemplateSpec.spec.containers.volumeMounts: Invalid value: "null": spec.podTemplateSpec.spec.containers.volumeMounts in body must be of type array: null`

This commit makes sure we avoid an empty volumeMounts property.

Fixes #<issue number>

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/charts/k8ssandra/templates/cassandra/cassdc.yaml
+++ b/charts/k8ssandra/templates/cassandra/cassdc.yaml
@@ -323,6 +323,7 @@ spec:
           {{- end }}
         {{- end }}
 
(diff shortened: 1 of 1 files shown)


jeffbanks (approved):
LGTM

PR #1199, opened 2021-11-19
don't create empty (i.e., null) containers in Reaper object

jsanda:
**What this PR does**:
On k8s versions < 1.21 creation of the Reaper object can fail with this:

`Error: Reaper.reaper.cassandra-reaper.io "k8ssandra-reaper" is invalid: spec.schemaInitContainerConfig: Invalid value: "null": spec.schemaInitContainerConfig in body must be of type object: null`

This commit fixes it by not creating empty containers.

Fixes #<issue number>

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/charts/k8ssandra/templates/reaper/reaper.yaml
+++ b/charts/k8ssandra/templates/reaper/reaper.yaml
@@ -32,14 +32,14 @@ spec:
   {{- if .Values.reaper.tolerations }}
   tolerations: {{ toYaml .Values.reaper.tolerations | nindent 4 }}
   {{- end }}
+  {{- if (and .Values.reaper.schemaInitContainerConfig .Values.reaper.schemaInitContainerConfig.securityContext) }}
   schemaInitContainerConfig:
-    {{- if (and .Values.reaper.schemaInitContainerConfig .Values.reaper.schemaInitContainerConfig.securityContext) }}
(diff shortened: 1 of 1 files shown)


charts/k8ssandra/templates/reaper/reaper.yaml
```diff
@@ -32,14 +32,14 @@ spec:
   {{- if .Values.reaper.tolerations }}
   tolerations: {{ toYaml .Values.reaper.tolerations | nindent 4 }}
   {{- end }}
+  {{- if (and .Values.reaper.schemaInitContainerConfig .Values.reaper.schemaInitContainerConfig.securityContext) }}
```
jeffbanks, line 35: In this change, if the context in yaml is specified as { } , which is our default, then we will apply whatever else is in the schema and config blocks. 

What if the user has a value setting for config containers without a securityContext entry? It would be null and we would disregard any ContainerConfig contents.  Wondering if it would be better to default the configs irrespective of the securityContext or maybe I'm missing something here.

jsanda: The security context is the only thing that we are exposing for those init containers so the checks should be fine as is.

jeffbanks (approved):

PR #1236, opened 2022-01-05
several k8ssandra-operator chart updates [K8SSAND-1128][K8SSAND-1129]

### LINKED ISSUE #1235: K8SSAND-1128 ⁃ k8ssandra-operator chart should support cluster-scope deployments
https://github.com/k8ssandra/k8ssandra-operator/issues/95 added support for cluster-scope deployments of k8ssandra-operator. The chart needs to be updated for this. This includes:

* Create the ClusterRole and ClusterRoleBinding
* Set the `WATCH_NAMESPACE` env var in the operator deployment to empty string
* Configure cass-operator to be cluster-scoped

There may be additional changes required, but this is the stuff I can think of off the top of my head.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1128) by [Unito](https://www.unito.io)
┆Reviewer: 5fa45931c8247300705e1fed
┆fixVersions: k8ssandra-operator-v1.0.0
┆friendlyId: K8SSAND-1128
┆priority: Medium

jsanda:
**What this PR does**:
* Add/update CRDs
* Add support for cluster-scoped deployment
* Fix naming
* Update RBAC
* `clusterScoped` chart property is now global in both k8ssandra-operator and in cass-operator charts

By making `clusterScoped` global it only needs to be set once vs setting it separately for both charts, e.g.,

`helm install test ./charts/k8ssandra-operator --set "global.clusterScoped=true"`

whereas without the global variable you would instead have to do,

`helm install ./charts/k8ssandra-operator --set "clusterScoped=true,cass-operator.clusterScoped=true"`

Making `clusterScoped` global makes it easier for users without introducing a lot additional complexity in the chart templates.

Fixes #1235

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/charts/k8ssandra-operator/templates/role.yaml
+++ b/charts/k8ssandra-operator/templates/role.yaml
@@ -1,4 +1,4 @@
-{{- if not .Values.clusterScoped }}
+{{- if not .Values.global.clusterScoped }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
--- a/charts/k8ssandra-operator/values.yaml
+++ b/charts/k8ssandra-operator/values.yaml
@@ -16,7 +16,7 @@ replicaCount: 1
 # K8ssandraCluster in the same namespace in which the operator is deployed
 # or if watches and manages K8ssandraClusters across all namespaces.
 # Not supported yet in k8ssandra-operator
-#clusterScoped: false
+clusterScoped: false
 
 # -- Determines if the k8ssandra-operator should be installed as the control plane
 # or if it's simply in a secondary cluster waiting to be promoted
@@ -33,7 +33,7 @@ image:
   pullPolicy: IfNotPresent
 
   # -- Tag of the k8ssandra-operator image to pull from image.repository
-  tag: 99c58003
+  tag: 5b95fc98
 
   # -- Docker registry containing all cass-operator related images. Setting this
   # allows for usage of an internal registry without specifying serverImage,
--- a/charts/k8ssandra-operator/values.yaml
+++ b/charts/k8ssandra-operator/values.yaml
@@ -1,3 +1,9 @@
+global:
+  # -- Determines whether k8ssandra-operator only watch and manages
+  # K8ssandraCluster in the same namespace in which the operator is deployed
+  # or if watches and manages K8ssandraClusters across all namespaces.
+  clusterScoped: false
+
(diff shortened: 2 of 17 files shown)


charts/k8ssandra-operator/values.yaml
```diff
@@ -33,7 +33,7 @@ image:
   pullPolicy: IfNotPresent
 
   # -- Tag of the k8ssandra-operator image to pull from image.repository
-  tag: 99c58003
+  tag: 5b95fc98
```
jdonenine, line 36: Should we go ahead and bump to the tag that was published this morning after the medusa integration got merged? `b05f3ec8`

I'll try to test with that image today.

jdonenine: Ah, never mind, there's a new CRD for the backup type, so we can't use it without more changes.  We can just follow this up with that.

charts/k8ssandra-operator/templates/role.yaml
```diff
@@ -1,7 +1,8 @@
+{{- if not .Values.global.clusterScoped }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
-  name: {{ include "k8ssandra-common.fullname" . }}-k8ssandra-operator
```
jdonenine, line 4: There's a few places where the suffix was removed like this... it makes things a lot easier to type and look at so I'm good with it, but just wanted to be sure it was intentional -- is this the "naming change" you mention in the PR description?

It does mean that copy and paste of examples in the current install guide no longer work -- we can clean that up as we migrate and write full docs though I think.

jdonenine (approved):
I was able to test a number of different scenarios using the new charts:
* Single cluster, single DC, shared namespace
* Single cluster, single DC, separate namespaces
* 4 clusters, 1 control plane, 3 data planes, 3 DCs, separate namespaces

Everything worked well.

I left one question about the naming change, but I think that's all fully as planned, so I'll go ahead and give my approval now.

charts/k8ssandra-operator/templates/role.yaml
jsanda: The naming changes were intentional so as to be consistent with how we do naming elsewhere in the charts. 

I forgot about the install guide when I made the changes. I'll open a PR to update it.

PR #1316, opened 2022-02-25
Fix versions dropdown on mobile resolutions

jdonenine:
**What this PR does**:
Makes the new `Versions` dropdown available on the mobile resolution version of the menu.

This will need to be cherry picked to `docs` and `docs-v1` as well.

N/A

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/docs/layouts/partials/navbar-version-selector.html
+++ b/docs/layouts/partials/navbar-version-selector.html
@@ -1,0 +1,8 @@
+<a class="nav-link dropdown-toggle" href="#" id="navbarDropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+	<span>{{ .Site.Params.version_menu }}</span>
+</a>
+<div class="dropdown-menu" aria-labelledby="navbarDropdownMenuLink">
+	{{ range .Site.Params.versions }}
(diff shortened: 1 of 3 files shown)


docs/layouts/partials/navbar.html
```diff
@@ -22,7 +22,7 @@
 			</li>
 			{{ end }}
 			{{ if  .Site.Params.versions }}
-			<li class="nav-item dropdown d-none d-lg-block">
+			<li class="nav-item dropdown mr-2 mr-xl-3 mb-2 mb-lg-0">
```
jdonenine, line 25: This is the change that actually makes the dropdown available on all resolutions (just like the other menu items)

docs/layouts/partials/navbar-version-selector.html
```diff
@@ -0,0 +1,8 @@
+<a class="nav-link dropdown-toggle" href="#" id="navbarDropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
```
jdonenine, line 1: I overrode the default version of this partial so that it's structured similarly to the other non-drop down menus with the li > a > span structure, this way it can inherit styles just like the rest of the menu items and not be a totally one-off case.

johnsmartco (approved):
Lgtm

PR #1329, opened 2022-03-01
Add K8ssandra Operator component topic on v2 branch

johnsmartco:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested


docs/content/en/components/k8ssandra-operator/_index.md
```diff
@@ -0,0 +1,139 @@
...
+* A supported cloud provider's Kubernetes environment
+
+## Architecture
+
+K8ssandra Operator consists of two primary components:
+
+* A control plane
+* A data plane
+
+The control plane creates and manages objects that exist only in the API server. The control plane does not deploy or manage pods. Note that the control plane can be installed in only one cluster; that is, in the control plane cluster.
```
jdonenine, line 22: This isn't strictly true in that a control-plane cluster *can* be used to deploy a datacenter, we'd just recommend for folks not to.  So I'm not totally sure how you want to word this, but we'll want to remove the bit about `The control plane does not deploy or manage pods`

jdonenine (changes requested):
One comment on some of the intro content, otherwise what is here I think is good - it's basically content from our preview blog, right? Which I think should work well as introductory content.

@jsanda recently merged some deeper dive architecture content here that I think we should consider how we want to use:  https://github.com/k8ssandra/k8ssandra-operator/blob/main/docs/arch-overview.md

I don't know that it should replace the content here, but it's definitely more detailed and can show the full picture -- what you get when you deploy a full cluster for example, with medusa and reaper and what not.

docs/content/en/components/k8ssandra-operator/_index.md
johnsmartco: Ok thanks will fix that and will grab John Sanda's deep dive architecture info.

johnsmartco:
Also - I just realized that I already queued up the components/k8ssandra-operator/_index.md and graphics files on PR 1325 (which is mostly about the install options, but anyway...).  So I've copied your comments from here. Will close this redundant PR without merging, and will add to PR 1325.

PR #1346, opened 2022-03-07
Upgrade medusa-operator dependency 0.1.0 -> 0.4.0

adejanovski:


jsanda:
This was already done in #1175

PR #1359, opened 2022-03-18
Set securityContext::runAsNonRoot to true for all containers

burmanm:
**What this PR does**:
@jsanda For discussion. "runAsNonRoot" prevents the container from running if it tries to run as the root. This PR shows the current state when that validation is enabled.

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/charts/k8ssandra/values.yaml
+++ b/charts/k8ssandra/values.yaml
@@ -435,7 +435,7 @@ stargate:
   enabled: true
   # -- version of Stargate to deploy. This is used in conjunction with cassandra.version to select the Stargate container image.
   # If stargate.image is set, this value has no effect.
-  version: "1.0.44"
+  version: "1.0.52"
   # -- Sets the Stargate container image. This value must be compatible
   # with the value provided for stargate.clusterVersion. If left blank (recommended),
   # k8ssandra will derive an appropriate image based on cassandra.clusterVersion.
@@ -569,6 +569,10 @@ stargate:
   # -- Tolerations to apply to the Stargate pods. See
   # https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/ for background.
   tolerations: []
+  # -- SecurityContext for all Stargate containers and init containers. Do not change from value 999, it will fail.
+  securityContext:
+    runAsNonRoot: true
+    runAsUser: 999
 reaper:
   # -- Security context override for reaper container
   securityContext:
--- a/charts/k8ssandra/values.yaml
+++ b/charts/k8ssandra/values.yaml
@@ -577,12 +577,12 @@ reaper:
   # -- Security context override for reaper container
   securityContext:
     runAsNonRoot: true
-    runAsUser: 65534
+    runAsUser: 1001
   # -- Security context override for reaper schema init container
   schemaInitContainerConfig:
     securityContext:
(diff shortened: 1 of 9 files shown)


charts/k8ssandra/values.yaml
```diff
@@ -565,13 +571,16 @@ stargate:
   tolerations: []
 reaper:
   # -- Security context override for reaper container
-  securityContext: {}
+  securityContext:
+    runAsNonRoot: true
```
jsanda, line 575: We also need to specify `runAsUser` here; otherwise, we will get this error:

```yaml
        message: container has runAsNonRoot and image has non-numeric user (reaper),
          cannot verify user is non-root
        reason: CreateContainerConfigError
```

charts/k8ssandra/values.yaml
```diff
@@ -565,13 +571,16 @@ stargate:
...
   # -- Security context override for reaper schema init container
   schemaInitContainerConfig:
-    securityContext: {}
+    securityContext:
+      runAsNonRoot: true
```
jsanda, line 579: We also need to specify `runAsUser` here; otherwise, we will get this error:

```yaml
        message: container has runAsNonRoot and image has non-numeric user (reaper),
          cannot verify user is non-root
        reason: CreateContainerConfigError
```

charts/k8ssandra/values.yaml
```diff
@@ -565,13 +571,16 @@ stargate:
...
   # -- Security context override for reaper config init container
   configInitContainerConfig:
```
jsanda, line 581: Interestingly there is a bug with `configInitContainerConfig`. The reaper.yaml template fails to render when it is set.   The template has:

```
{{- if (and .Values.reaper.configInitContainerConfig .Values.reaper.configInitContainerConfig.securityContext) }}
  configInitContainerConfig:
    securityContext: {{ toYaml .Values.reaper.configInitContainerConfig.securityContext | nindent 6 }}
  {{- end }}
```
but the Reaper spec does not have a `configInitContainerConfig` property. Can you fix the template and remove this property from values.yaml?

jsanda (dismissed):
Looks like you missed the `waitForCassandra` init container for Stargate.

charts/k8ssandra/values.yaml
jeffbanks: @jsanda I believe the `configInitContainerConfig`  was used in the `reaper-operator` with reference in the types [here](https://github.com/k8ssandra/reaper-operator/blob/2a746e60e8dfb7529e6acc7ce5adf714fd063fba/api/v1alpha1/reaper_types.go).

jsanda: @jeffbanks interesting. I don't see it in [0.3.5](https://github.com/k8ssandra/reaper-operator/blob/v0.3.5/api/v1alpha1/reaper_types.go). And indeed if you set the property in values.yaml you will get a template rendering error in reaper.yaml.

jeffbanks: Confirmed this is a post 0.3.5 change/commit for `reaper-operator` as part of [this PR](https://github.com/k8ssandra/reaper-operator/pull/84).

PR #1373, opened 2022-04-01
Fix gc properties 4.0

### LINKED ISSUE #1269: K8SSAND-1192 ⁃ maxGcPauseMillis helm chart setting does not execute against JVM command
## Bug Report
**Describe the bug**
When setting maxGcPauseMillis, and then checking the options that are running on the JVM, it does not seem to have gone through.

**To Reproduce**

1. In your helm chart, change the maxGcPausMillis value:

```yaml
 datacenters:
      -
        audit_logging: {}
        client_backpressure: {}
        fql: {}
        gc:
          cms: {}
          g1: 
            enabled: true
            maxGcPauseMillis: **200**
            parallelGcThreads: 16
            concurrentGcThreads: 16
```

1. Restart the DataCenter
1. Go into a CLI and ps -ef on the java job to see the setting as the value of 500

**Expected behavior**
We would have expected the value to change to 200

* Helm charts version info
4.14.1
 maxGcPauseMillis: **200**
4.14.1
`1.20.9`
* Kubernetes cluster kind:

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1192) by [Unito](https://www.unito.io)
┆fixVersions: k8ssandra-v1.5.0
┆friendlyId: K8SSAND-1192
┆priority: Medium

adejanovski:
**What this PR does**:
Fixes GC settings for Cassandra 4.0 which were not applied correctly as they are supposed to go in jvm11-server.options and jvm-server.options, but were only going in the latter.
I also introduced the possibility to use ZGC with Cassandra 4.0 as it's available with JDK11 and cass-config-definitions support it.

Fixes #1269

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/CHANGELOG-1.5.md
+++ b/CHANGELOG-1.5.md
@@ -18,7 +18,9 @@ and date `## vX.Y.Z - YYYY-MM-DD` and create a new placeholder section for  `unr
 
 ## unreleased
 
+* [FEATURE] Enable the use of ZGC (Z Garbage Collector) for Cassandra 4.0
 * [CHANGE] Upgrade Stargate to v1.0.52
 * [CHANGE] Upgrade Medusa to v0.12.1
 * [CHANGE] Upgrade Reaper to v3.1.1
 * [CHANGE] Upgrade Management API to v0.1.37
+* [BUGFIX] Fix GC subsettings mappings for Cassandra 4.0
--- a/charts/k8ssandra/templates/_helpers.tpl
+++ b/charts/k8ssandra/templates/_helpers.tpl
@@ -328,8 +328,10 @@ Add garbage collection settings based on the following rules in the order listed
 {{- $datacenter := (index .Values.cassandra.datacenters 0) }}
 {{- $clusterCmsEnabled := .Values.cassandra.gc.cms.enabled }}
 {{- $clusterG1Enabled := .Values.cassandra.gc.g1.enabled }}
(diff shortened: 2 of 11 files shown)


jsanda (commented):
I did a manual test with these chart properties:

```yaml
cassandra:
  version: 4.0.3
  gc:
    g1:
      enabled: true
      maxGcPauseMillis: 250
  heap:
    size: 512M
  datacenters:
  - name: dc1
    size: 1
```
I see `-XX:MaxGCPauseMillis=250` in jvm11-server.options, but I also see `-XX:MaxGCPauseMillis=500` in jvm.options.

I also tested with ZGC and still see `-XX:+UseG1GC` in jvm.options. I do see `-XX:+UseZGC` in jvm11-server.options.

adutra:
I think we have a bug in config builder. I noticed the same behavior. The file jvm.options shouldn't even be generated for C 4.0. But we have good news: that file is not sourced by C 4.0 so everything that is in there is simply ignored.

adejanovski:
yes, we've noticed that `jvm.options` was always generated with default values, but as Alex said, it's ignored.

jsanda (approved):
Chatted offline with @adejanovski and confirmed that the presence of properties in jvm.options shouldn't be a problem since Cassandra will use jvm11-server.options. We would need to change cass-config-builder and/or cass-config-definitions so that it doesn't generate the settings for jvm.options. It's unnecessary and out of scope.

PR #1390, opened 2022-04-25
Upgrade cass-operator dependency 1.10.3 ->

adejanovski:


burmanm:
So now the script is even deleting newest versions? ;)

adejanovski:
ha :D 

I don't understand what's going on and why this is only happening for cass-operator 🤔 
There seem to be some stale/cached returns from the docker hub apis in GHA which drive the workflow crazy.
Maybe there's something obvious I'm missing in the code that creates this...

PR #1480, opened 2022-08-03
Fix CRD upgrader breakage for the k8ssandra chart

### LINKED ISSUE #1478: K8SSAND-1702 ⁃ CRD upgrader is broken for K8ssandra v1.x
Since we made changes to allow the CRD upgrader to support K8ssandra-operator, it seems like we broke it for K8ssandra v1.x.

Here's the traces I've had today when performing an upgrade on a v1.x cluster:

```
2022/07/28 11:45:10 Upgrading CRDs to version 1.5.0
I0728 11:45:11.096540       1 request.go:665] Waited for 1.023299486s due to client-side throttling, not priority and fairness, request: GET:https://10.60.16.1:443/apis/scheduling.k8s.io/v1?timeout=32s
2022/07/28 11:45:12 Downloading release 1.5.0 from Helm repository
2022/07/28 11:45:13 Failed to update CRDs: no chart name found
```

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1702) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1702
┆priority: Medium

adejanovski:
**What this PR does**:
Uses `k8ssandra` as default chart name if it isn't provided.
This should fix the K8ssandra v1.x upgrade issues as a new `chartName` argument was introduced to support K8ssandra-operator, but no default value was defined to handle the former invocations made by K8ssandra v1.

Fixes #1478 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/pkg/crds/upgrade.go
+++ b/pkg/crds/upgrade.go
@@ -58,6 +58,10 @@ func New(namespace string) (*Upgrader, error) {
 
 // Upgrade installs the missing CRDs or updates them if they exists already
 func (u *Upgrader) Upgrade(chartName string, targetVersion string) ([]unstructured.Unstructured, error) {
+	chartNameToUpgrade := chartName
+	if chartName == "" {
+		chartNameToUpgrade = helmutil.DefaultChartName
+	}
 	extractDir, err := helmutil.GetChartTargetDir(targetVersion)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


jsanda (commented):
@adejanovski did the hit the upgrade error before #1478 was merged? Is that PR a sufficient fix since it adds the `chartName` arg?

adejanovski:
> @adejanovski did the hit the upgrade error before #1478 was merged? Is that PR a sufficient fix since it adds the `chartName` arg?

I assume you meant #1438, an we didn't get that error before. #1438 introduced the `--chart-name` arg which isn't passed by the CRD upgrader invocations from K8ssandra v1.x, which leads to an empty `chartName` in the `Upgrade()` func (previously `k8ssandra` was the only chart to upgrade).
With this change, we deal with this case by using `k8ssandra` as default if the provided `chartName` is empty.

jsanda (approved):
#1438 did add the `chart-name` arg to the k8ssandra chart. Is the problem due to our always using `latest` for k8ssandra-tools? If so let's get a new issue created to make sure we are not using latest.

We should probably add the upgrade job to the cass-operator chart as well (new issue of course).

PR #1505, opened 2022-09-14
Add documentation on bootstrap and decommission order

adutra:
**What this PR does**:
Adds documentation on bootstrap and decommission order

None

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/docs/content/en/tasks/scale/_index.md
+++ b/docs/content/en/tasks/scale/_index.md
@@ -113,7 +113,12 @@ Alternatively, we can also patch the existing object with a `kubectl patch` comm
 kubectl patch k8c my-k8ssandra --type='json' -p='[{"op": "replace", "path": "/spec/cassandra/datacenters/0/size", "value": 4}]'
 ```
 
-### Underlying considerations when increasing size values
+### Underlying considerations when scaling up
+
+After the new nodes are up and running, `nodetool cleanup` should run on all of the nodes except the
+new ones to remove keys and data that no longer belong to those nodes. There is no need to do this
+manually. The cass-operator deployment, which again is installed with K8ssandra, automatically runs
+`nodetool cleanup` for you.
 
 By default, cass-operator configures the Cassandra pods so that Kubernetes will not schedule
 multiple Cassandra pods on the same worker node. If you try to increase the cluster size beyond the
@@ -205,32 +210,32 @@ When applied, this configuration updates the size property of the `CassandraData
 cass-operator will in turn update the underlying `StatefulSet`, allowing more than one Cassandra
 node to sit on the same worker node.
 
-After the new nodes are up and running, `nodetool cleanup` should run on all of the nodes except the
-new ones to remove keys and data that no longer belong to those nodes. There is no need to do this
-manually. The cass-operator deployment, which again is installed with K8ssandra, automatically runs
-`nodetool cleanup` for you.
-
-### Datacenter status and conditions when scaling up
+### Conditions and events when scaling up
 
 If you check the status of the `CassandraDatacenter` object, there should be a `ScalingUp` condition
-with its status set to `true`. It should look like this:
+with its status set to `true`. You should also see one or more `ScalingUpRack` events, one per rack:
 
 ```bash
- kubectl get cassandradatacenter dc1 -o yaml
+ kubectl describe cassandradatacenter dc1
 ```
 
 **Output:**
 
 ```text
 ...
-status:
-  cassandraOperatorProgress: Updating
-  conditions:
-  - lastTransitionTime: "2021-03-30T22:01:48Z"
-    message: ""
-    reason: ""
-    status: "True"
-    type: ScalingUp
+Status:
+  Cassandra Operator Progress:  Updating
+  Conditions:
+    Last Transition Time:    2022-09-16T11:39:35Z
+    Message:                 
+    Reason:                  
+    Status:                  True
+    Type:                    ScalingUp
+
+Events:
+  Type    Reason               Age        From           Message
+  ----    ------               ----       ----           -------
+  Normal  ScalingUpRack        4m         cass-operator  Scaling up rack default
 ...
 ```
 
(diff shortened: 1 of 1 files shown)


docs/content/en/tasks/scale/_index.md
```diff
@@ -188,7 +188,49 @@ The StatefulSet controller manages the deletion of Cassandra pods. It deletes on
 This means for example that `my-k8ssandra-dc1-default-sts-3` will be deleted before `my-k8ssandra-dc1-default-sts-2`.
 {{% /alert %}}
```
jsanda, line 190: This file is in the v1 part of the docs. Don't we want to update https://docs-v2.k8ssandra.io/tasks/scale?

docs/content/en/tasks/scale/_index.md
```diff
@@ -188,7 +188,49 @@ The StatefulSet controller manages the deletion of Cassandra pods. It deletes on
...
+number of nodes, the operator will start with the rack that comes first in the datacenter
+definition.
+
+Let's take a look at an example. Suppose we have a 3-node datacenter with the following racks:
+
+* `rack1` with 1 node
+* `rack2` with 1 node
+* `rack3` with 1 node
+
+Now let's scale up to, say, 8 nodes. The operator will bootstrap 5 new nodes, exactly in the
```
jsanda, line 207: Neither the v1 nor the v2 docs mention anything about the cleanup operation that is run. We should add some info about that.

jsanda, line 207: cass-operator generates some k8s events when a node is added. It would be helpful to point that out. It would also be helpful to point out the status conditions that get set.

docs/content/en/tasks/scale/_index.md
```diff
@@ -188,7 +188,49 @@ The StatefulSet controller manages the deletion of Cassandra pods. It deletes on
...
+
+The resulting topology will be as follows:
+
+* `rack1` with 3 nodes
+* `rack2` with 3 nodes
+* `rack3` with 2 nodes
+
+This is indeed the most balanced topology we could have achieved with 8 nodes and 3 racks.
+
+Now let's scale down back to 3 nodes. The operator will decommission 5 nodes, exactly in the
```
jsanda, line 224: cass-operator generates some k8s events when a node is remove. It would be helpful to point that out. It would also be helpful to point out the status conditions that get set.

jsanda (changes requested):

docs/content/en/tasks/scale/_index.md
adutra: Yes I was planning to revamp the entire page, I just pushed the changes.

docs/content/en/tasks/scale/_index.md
adutra: There is a mention already:

```
After the new nodes are up and running, `nodetool cleanup` should run on all of the nodes except the
new ones to remove keys and data that no longer belong to those nodes. There is no need to do this
manually. The cass-operator deployment, which again is installed with K8ssandra, automatically runs
`nodetool cleanup` for you.
```

adutra: The conditions ScalingUp and ScalingDown are already mentioned. I can mention the events too, but please keep in mind that in my experience, events are not reliable (some events don't show up in `kubectl get events`).

adutra: I added a note about events, but again, as I was testing my own docs, I observed that some events get lost. In particular, I'm failing to see the ScalingDownRack events, even if I see them in the operator logs.

adutra:
This is ready for another review @jsanda or @adejanovski .

docs/content/en/tasks/scale/_index.md
```diff
@@ -4,95 +4,156 @@ linkTitle: "Scale"
...
+kubectl apply -f k8ssandra.yaml
 ```
 
-{{% alert title="Tip" color="success" %}}
-Another way to upgrade your K8ssandra cluster is by passing in a `--set` parameter. Also include a `--reuse-values` parameter so that Helm will reuse previous values (other than the one you're overriding with each `--set` parameter). Without `--reuse-values` it's easy to make a mistake if you have other, additional properties that you previously set.  
+Alternatively, we can also patch the existing object with a `kubectl patch` command:
 
 ```bash
-helm upgrade my-k8ssandra k8ssandra/k8ssandra --reuse-values --set cassandra.datacenters\[0\].size=4,cassandra.datacenters\[0\].name=dc1
+kubectl patch k8c my-k8ssandra --type='json' -p='[{"op": "replace", "path": "/spec/cassandra/datacenters/0/size", "value": 4}]'
```
jsanda, line 113: 👍

docs/content/en/tasks/scale/_index.md
```diff
@@ -4,95 +4,156 @@ linkTitle: "Scale"
...
 ```
 
-{{% /alert %}}
+### Underlying considerations when scaling up
 
-## Underlying considerations when increasing size values
+After the new nodes are up and running, `nodetool cleanup` should run on all of the nodes except the
```
jsanda, line 118: 👍

docs/content/en/tasks/scale/_index.md
```diff
@@ -102,63 +163,103 @@ Events:
...
+  Normal   StartedCassandra             2m11s     cass-operator  Started Cassandra for pod my-k8ssandra-dc1-default-sts-1
 ...
 ```
 
-After the new nodes are up and running, `nodetool cleanup` should run on all of the nodes except the new ones to remove keys and data that no longer belong to those nodes. There is no need to do this manually. The cass-operator deployment, which again is installed with K8ssandra, automatically runs `nodetool cleanup` for you.
+You can wait for the scale up operation to be finished by checking the value of the `ScalingUp`
+condition. This can also be done with the following `kubectl wait` command:
+
+```bash
+kubectl wait --for=condition=ScalingUp=false cassandradatacenter dc1 --timeout=30m
```
jsanda, line 250: 👍

jsanda (approved):
Thanks for the additional changes, particularly for the output for status conditions and events. It looks great :) The `kubectl patch` and `kubectl wait` commands are really nice. They are nice tips for how to build one's own automation!

PR #1660, opened 2024-08-06
Fix: Create Clusterrolebinding in cass-operator chart in any case

mlallaouret:
**What this PR does**:

Fix clusterrolebinding creation.

After PR https://github.com/k8ssandra/k8ssandra/pull/1657, the ClusterRole is now created unconditionally but its related ClusterRoleBinding is still created only if .Values.global.clusterScoped is true

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/charts/cass-operator/templates/rolebinding.yaml
+++ b/charts/cass-operator/templates/rolebinding.yaml
@@ -1,4 +1,3 @@
-{{- if .Values.global.clusterScoped }}
 kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
@@ -11,7 +10,8 @@ roleRef:
   kind: ClusterRole
   name: {{ template "k8ssandra-common.fullname" . }}
   apiGroup: rbac.authorization.k8s.io
-  {{- else }}
+{{- if not .Values.global.clusterScoped }}
+---
 kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
@@ -24,4 +24,4 @@ roleRef:
   kind: Role
   name: {{ template "k8ssandra-common.fullname" . }}
   apiGroup: rbac.authorization.k8s.io
-  {{- end }}
+{{- end }}
--- a/charts/cass-operator/Chart.yaml
+++ b/charts/cass-operator/Chart.yaml
@@ -3,7 +3,7 @@ name: cass-operator
 description: |
   Kubernetes operator which handles the provisioning and management of Apache Cassandra clusters.
 type: application
-version: 0.53.1
+version: 0.53.2
 appVersion: 1.22.0
 dependencies:
   - name: k8ssandra-common


charts/cass-operator/templates/rolebinding.yaml
```diff
@@ -11,7 +10,8 @@ roleRef:
   kind: ClusterRole
   name: {{ template "k8ssandra-common.fullname" . }}
   apiGroup: rbac.authorization.k8s.io
-  {{- else }}
+{{- if not .Values.global.clusterScoped }}
```
Miles-Garnsey, line 13: Thanks for this contribution @mlallaouret, from what I can understand here, it looks like we need a ClusterRole and ClusterRoleBinding irrespective of whether `.Values.global.clusterScoped` is true. 

It appears that this bit of logic might also create a RoleBinding to the same ClusterRole, but only when  `.Values.global.clusterScoped` is false. From my understanding, the RoleBinding should probably never be created, as I think the ClusterRoleBinding will be sufficient to give bind the ServiceAccount to the ClusterRole, and it is getting created in either case.

Is my understanding correct here? Can we remove the RoleBinding entirely?

Miles-Garnsey (changes requested):

charts/cass-operator/templates/rolebinding.yaml
KeepItSimpleStupid: Hi Miles !
Jumping in because I reviewed Marc's PR before we submitted it to you ;) 

For me, the `RoleBinding` below targets the (conditionally created) `Role` and not the (unconditionally created) `ClusterRole` in its `roleRef`

To sum up :
- if `.Values.global.clusterScoped` is `true`, we have those RBAC resources
  - a `ClusterRole` with full permissions for the operator (cluster-scoped & namespaced-scopes resources )
  - a `ClusterRoleBinding` targeting the `ClusterRole` above

- if `.Values.global.clusterScoped` is `false` (the default for the chart), we have those RBAC resources
  - a `ClusterRole` with minimalistic permissions to access cluster-scoped resources (storageclasses & persistentvolumes)
  - a `ClusterRoleBinding` targeting the `ClusterRole` above
  - a `Role` with full permissions for the operator except those for cluster-scoped resources (storageclasses & persistentvolumes)
  - a `RoleBinding` targeting the `Role` above

If I understand well, the reason behind this is that you can't give access to cluster-scoped resources with a `Role`, only with a `ClusterRole`

Miles-Garnsey:
Hi folks, apologies we'll have to come back to you next week on this, I'm out of cell range and so is a colleague I need to consult on it.

charts/cass-operator/templates/rolebinding.yaml
burmanm: What @KeepItSimpleStupid said is right. However, these changes in this repo would be overwritten as these are generated files.

So the updates should happen in https://github.com/k8ssandra/cass-operator/blob/master/scripts/release-helm-chart.sh and then rerun here.

I can get back to this subject next week.

KeepItSimpleStupid: Thanks Michael !

Am I wrong if I say that the script you pointed changes only :
- `charts/cass-operator/crds` folder
- `charts/cass-operator/templates/role.yaml`
- `charts/cass-operator/Chart.yaml` (to bump appVersion)
- `charts/cass-operator/values.yaml`

So touching `charts/cass-operator/templates/rolebinding.yaml` in this PR should be safe ?

burmanm: Right, so it seems the rolebinding.yaml wasn't regenerated. Not that it needs to be. So the PR should be safe, yep.

burmanm (approved):

KeepItSimpleStupid:
Thanks @burmanm & @Miles-Garnsey !
I created this PR on the k8ssandra-operator helm chart as a follow up : https://github.com/k8ssandra/k8ssandra-operator/pull/1381

KeepItSimpleStupid:
> Thanks @burmanm & @Miles-Garnsey ! I created this PR on the k8ssandra-operator helm chart as a follow up : [k8ssandra/k8ssandra-operator#1381](https://github.com/k8ssandra/k8ssandra-operator/pull/1381)

@burmanm @Miles-Garnsey Would it be possible to bump the version of the cass-operator chart used in the k8ssandra-operator chart (and release it) ? That's our only way to access the fix contained in this Pull Request. Thanks !

burmanm:
Yeah, we'll try to cut one today or Monday.

PR #1715, opened 2026-04-29
Update k8ssandra-client to v0.8.11

burmanm:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/charts/cass-operator/Chart.yaml
+++ b/charts/cass-operator/Chart.yaml
@@ -3,7 +3,7 @@ name: cass-operator
 description: |
   Kubernetes operator which handles the provisioning and management of Apache Cassandra clusters.
 type: application
-version: 0.64.0
+version: 0.64.1
 appVersion: 1.30.0
 dependencies:
   - name: k8ssandra-common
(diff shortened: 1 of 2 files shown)


charts/cass-operator/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: cass-operator
 description: |
   Kubernetes operator which handles the provisioning and management of Apache Cassandra clusters.
 type: application
-version: 0.64.0
+version: 0.64.1
```
artsiom-bukhautsou, line 6: why do we need to update this as well?

artsiom-bukhautsou (approved):

charts/cass-operator/Chart.yaml
burmanm: Otherwise we would overwrite the existing Helm chart and people on 0.64.0 wouldn't get an update. This is the version that Helm uses for versioning, not the appVersion.

artsiom-bukhautsou: thanks

PR #1720, opened 2026-06-30
Support restricted-v2 in Openshift

burmanm:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/charts/cass-operator/templates/deployment.yaml
+++ b/charts/cass-operator/templates/deployment.yaml
@@ -10,6 +10,21 @@
 {{- $metricsPort := regexFind "(?:)(\\d*)$" .Values.metrics.address -}}
 {{- $exposeMetrics := and $metricsPort (not (eq "127.0.0.1" $metricsHostname )) -}}
 {{- $metricsTLSEnabled := .Values.metrics.tls.enabled | default false -}}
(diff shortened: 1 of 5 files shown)


charts/cass-operator/templates/deployment.yaml
```diff
@@ -10,6 +10,21 @@
 {{- $metricsPort := regexFind "(?:)(\\d*)$" .Values.metrics.address -}}
 {{- $exposeMetrics := and $metricsPort (not (eq "127.0.0.1" $metricsHostname )) -}}
 {{- $metricsTLSEnabled := .Values.metrics.tls.enabled | default false -}}
+{{- $openshiftMode := .Values.openshift.mode | default "auto" -}}
```
adejanovski, line 13: **Question:** are there cases where you think the auto detection would fail?
Any reason why we'd want to disable this while running on openshift?

burmanm: A user might want to, for example if they have a policy where instead of auto-generated uid they have a manual process to enforce uidRange. Or they have their own SCC instead of the restricted-v2.

I don't know how common those cases are, but I figured we shouldn't enforce these without option to avoid it.

adejanovski (approved):

PR #1723, opened 2026-07-23
Naming fixes, TLS configuration fixes, metrics fixes..

### LINKED ISSUE #955: Fix metrics service issues in Helm charts
### What happened?

There are some mistakes in the metrics service object when deploying using Helm. We target a wrong port and there's some naming issues in templates that do not match correctly. As such, the connection will fail using this service.

There are some smaller issues as well related to the TLS objects as well as some deviation from the Kustomize templates, which we could fix at the same time.

### What did you expect to happen?

_No response_

### How can we reproduce it (as minimally and precisely as possible)?

Deploy current Helm chart.

### cass-operator version

1.31.0

### Kubernetes version

1.36

### Method of installation

Helm

### Anything else we need to know?

_No response_

burmanm:
**What this PR does**:
* Updates TLS configuration
* Fixes metrics ports and webhook service stuff
* Updates documentation with "helm-docs" ran after a long while
* Updates values.yaml with fixes for those "helm-docs" requirements
* Align certain parameters from our kustomize rendering

This should not change default settings rendering too much (other than fixing some ports and names). 

Fixes https://github.com/k8ssandra/cass-operator/issues/955

**Checklist**
- [ ] Changes manually tested


charts/cass-operator/templates/deployment.yaml
```diff
@@ -5,11 +5,13 @@
...
 {{- end -}}
-
 {{- $metricsHostname := regexFind "[^:]*" .Values.metrics.address -}}
 {{- $metricsPort := regexFind "(?:)(\\d*)$" .Values.metrics.address -}}
-{{- $exposeMetrics := and $metricsPort (not (eq "127.0.0.1" $metricsHostname )) -}}
-{{- $metricsTLSEnabled := .Values.metrics.tls.enabled | default false -}}
+{{- $metricsEnabled := not (eq "0" $metricsPort) -}}
+{{- $exposeMetrics := and $metricsEnabled $metricsPort (not (eq "127.0.0.1" $metricsHostname)) -}}
+{{- $metricsTLSEnabled := and $metricsEnabled (.Values.metrics.tls.enabled | default false) -}}
+{{- $certificatesRequired := or $webhooks $metricsTLSEnabled -}}
```
rzvoncek: question: I cannot find a place where this variable is used. can we maybe remove it?

charts/cass-operator/values.yaml
```diff
@@ -152,12 +150,14 @@ tolerations: []
 # localhost (such as when using kube-rbac-proxy to secure access to them), set the value
 # to 127.0.0.1:8080
 metrics:
+  # -- Address where the metrics endpoint binds. Use "0" to disable metrics.
```
rzvoncek, line 153: nit: I'm not 100% sure a bare `0` (so not a `:0`) would work here. 

There's that `regexFind "(?:)(\\d*)$"` , does that return both groups?
If it does, then the `not (eq "0" $metricsPort)` probably doesn't do what it should, becaues `metricsPort` is `:0`

charts/cass-operator/UPGRADE.md
```diff
@@ -1,5 +1,16 @@
 # cass-operator
 
+## Upgrading to the next chart release (0.66)
+
+This release removes the unused `image.repositoryOverride`, `image.registryOverride`, and `image.namespaceOverride` values. Configure Cassandra image defaults and overrides through `global.imageConfig` instead since the older ImageConfig v1 is removed from the current version of cass-operator.
+
+If you wish to disable the cert-manager managed self signed certificate, the process has changed a little. Set `admissionWebhooks.certificateSecret` to use an externally managed webhook TLS certificates. Setting that secret will override our default method of creating a self-signed certificate for the webhooks. Note that the Secret must have certain structure at this point:
+
+* Must contain `tls.crt`, `tls.key`, and `ca.crt`
+* Have the `cert-manager.io/allow-direct-injection: "true"` annotation.
```
rzvoncek, line 10: comment: I wonder if it's reasonable to add a check for this and make helm fail if it's not present.

charts/cass-operator/UPGRADE.md
```diff
@@ -1,5 +1,16 @@
...
+
+This does not remove the requirement to have cert-manager installed in the cluster.
+
 ### Steps to renegerate using kustomize in cass-operator
```
rzvoncek:
```suggestion
### Steps to regenerate using kustomize in cass-operator
```

charts/cass-operator/values.yaml
burmanm: So, I didn't want to change the regexp, although I know it's broken in the Helm chart. Golang's regexp does not support non-catching groups, the (?:) one, so it ignores the output.

Thus, both ":0" and "0" will both return "0" as $metricsPort.

charts/cass-operator/templates/deployment.yaml
burmanm: Yeah, I'll remove it, it was part of my development but I decided to not use it.

charts/cass-operator/UPGRADE.md
burmanm: I guess it's theoretically possible, we can check that the Secret exist before allowing it to be used. I'm not sure about the keys though, annotations are maybe..

{{ range $index, $service := (lookup "v1" "Secret" .Release.Namespace .Values.admissionWebhooks.certificateSecret).metadata.annotations }}
    {{/* check that we have that? */}}
{{ end }}

Or fail if not..

burmanm: I wonder though, if I add this check, how can one helm template it anymore? There's no way to add "--api-versions" or similar checks for these cases I think. 

In which case one couldn't use it with ArgoCD for example, unless we add more parameters "disableThisCheck" .. and that's not pretty.

rzvoncek: okay, I see, let's not overcomplicate this then.

rzvoncek (approved):

charts/cass-operator/UPGRADE.md
burmanm: This text wasn't part of this PR.

PR #1727, opened 2026-09-16
Add optional automountServiceAccountToken support to common ServiceAccount

jmederer:
**What this PR does**:
Extend the k8ssandra-common.serviceAccount template so that
automountServiceAccountToken is only rendered when explicitly set by the
user via values.yaml. The field is intentionally left commented out in
the cass-operator, medusa-operator and reaper-operator values files to
avoid changing the default behaviour of existing deployments.

Only set the field when it is explicitly configured, preserving
backwards compatibility while allowing stricter token mounting on
request.

n/a

**Checklist**
- [x] Changes manually tested


charts/cass-operator/values.yaml
```diff
@@ -105,6 +105,10 @@ serviceAccount:
   name: ""
   # -- Annotations to add to the service account.
   annotations: {}
+  # -- Specifies whether a service account token is automatically mounted into
+  # containers running on pods that use this service account, and into the
+  # operator pod itself.
```
burmanm, line 110: Hey, this comment could mention "set to true by default" or something. Otherwise it might trick users into uncommenting that line and without knowledge of what's happening that would render their service unusable.

burmanm:
Thanks for the PR. Do not worry about medusa-operator/reaper etc, they're not maintained. Only cass-operator and the common are used here. The common would need a bump to the next minor version if you introduce changes to it, otherwise they will not take effect in cass-operator (cass-operator would need a bump of k8ssandra-common in Chart.yaml) or k8ssandra-operator. 

As for your use-case, did you consider two options, ServiceAccount creation time automount and at the Pod level? Note that the latter would override the first.

jmederer:
Hi @burmanm 

Use-case / both levels – Yes, we deliberately considered both options and implemented both, and we're aware the Pod-level setting overrides the ServiceAccount-level one. Our background is a Kyverno policy in our cluster that requires every ServiceAccount to be created with automountServiceAccountToken: false. To satisfy that policy we need to render the field on the ServiceAccount resource, while the Pod-level override lets us re-enable mounting per workload where it's actually needed.

Backwards compatibility – To make sure this doesn't disturb anyone else, both options are only rendered when the value is explicitly set in values.yaml. The field is left commented out by default, so existing deployments render exactly as before and nothing changes for users who don't opt in. We also documented the override precedence next to the option. And i changed it to true as you mentioned

Should i remove the changes form medusa and reaper?

An the MR common (template) https://github.com/k8ssandra/k8ssandra-operator/pull/1794/changes

burmanm:
The change in k8ssandra-operator uses Pod level setting and here it would use ServiceAccount level setting. So there's a bit of inconsistency.

So, the remaining stuff (the version handling isn't automated on this repository):

Update https://github.com/k8ssandra/k8ssandra/blob/main/charts/k8ssandra-common/Chart.yaml#L6

from `version: 0.29.2` to `version: 0.29.3`

Update https://github.com/k8ssandra/k8ssandra/blob/main/charts/cass-operator/Chart.yaml#L10 to use 0.29.3 as the dependency version. And the chartVersion to 0.67.2 if you want it updated instantly to use the new dependency.

Then in the k8ssandra-operator, update the same (in Chart.yaml), only the dependency to 0.29.3. That should be enough, then you can render all the ServiceAccounts with that setting. There's no need to touch any deployment template in the k8ssandra-operator, only the helper here. The Helm chart only gets updated during the k8ssandra-operator release however, unlike cass-operator charts.

Otherwise, we would have the Pod level override only for k8ssandra-operator, not for cass-operator. To have it on the cass-operator also, we need to copy the template changes to cass-operator chart also. If that's required, if the SA level is enough, then the k8ssandra-operator PR becomes just a bump of that dependency.

burmanm:
Hey, there's still issue with the `.Values.serviceAccount.automountServiceAccountToken` value used in two places. If set, it will be set in the SA and in the Deployment for the Pod. 

Those two values can't diverge.

jmederer:
> Hey, there's still issue with the `.Values.serviceAccount.automountServiceAccountToken` value used in two places. If set, it will be set in the SA and in the Deployment for the Pod.
> 
> Those two values can't diverge.

You're absolutely right — I made that mistake. The same value was indeed being used in two places. I've fixed it now: the Deployment pod spec uses a separate top-level automountServiceAccountToken value, while serviceAccount.automountServiceAccountToken is only used for the ServiceAccount resource itself. Thank you again for catching that!

burmanm (approved):

